PR #10, opened 2019-04-15
Added contributing document.

gyliu513:

Fixes #

**Release note**:

```release-note
NONE
```

/cc @bradtopol @jichenjc

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,49 @@
+# Contributing guidelines
+
+## Sign the CLA
+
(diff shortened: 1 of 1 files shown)


bradtopol:
overall this looks like a good start and we can iterate on it. One question.  You say folks should use minikube. Any reason you picked minikube over kind? I heard kind was faster at least according to the other SIG Cluster lifecycle folks

gyliu513:
Thanks @bradtopol , I updated the document to using Kind as you proposed. ;-)

Issue #21, opened 2019-04-15
kustomize generation for yamls 


jichenjc:
/kind feature

need provide kustomize for create something like cluster-api image defintion, cloud secrets, include from cluster-api definition etc

go to cluster-api-provider-openstack/cmd/clusterctl/examples/openstack for further info

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

gyliu513:
I'd like the sub-task as follows:

- [ ] Create hard-code `cluster.yaml`, `machines.yaml` and `provider-components.yaml` to make sure the whole process works
- [ ] Create kind of `generate_yaml.sh` to leverage `kustomize` to generate the above three files.

jichenjc:
yes, it can be divided into 2 parts (or even more), thanks for the comments

xunpan:
Do you mind to close this ticket if have no other comments. 
Current generation script can generate .yaml with kustomize correctly.

PR #25, opened 2019-04-17
Add template for clouds, cluster, machine

jichenjc:
part of #21

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/cmd/clusterctl/definition/ibmcloud/machines.yaml.template
+++ b/cmd/clusterctl/definition/ibmcloud/machines.yaml.template
@@ -1,0 +1,44 @@
+items:
+- apiVersion: "cluster.k8s.io/v1alpha1"
+  kind: Machine
+  metadata:
+    generateName: ibmcloud-master-
+    labels:
+      set: master
+  spec:
+    providerSpec:
+      value:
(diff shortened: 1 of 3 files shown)


cmd/clusterctl/definition/ibmcloud/machines.yaml.template
```diff
@@ -0,0 +1,44 @@
...
+    generateName: ibmcloud-master-
+    labels:
+      set: master
+  spec:
+    providerSpec:
+      value:
+        apiVersion: "ibmcloudproviderconfig/v1alpha1"
+        kind: "IbmcloudMachineProviderSpec"
+        hostname: <host name>
+        domain: <domain name>
```
gyliu513: @jichenjc do we need floating IP here for external access?

jichenjc: looks like (from my hands on) the public IP is assigned by default and you can't request one machine with given IP address ,so no need to put into template but we can get ip later when machine created

gyliu513: good to know, thanks ;)

Issue #26, opened 2019-04-17
update IbmcloudClusterProviderSpecStatus and IbmcloudClusterProviderSpecSpec


jichenjc:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
something like following 
IbmcloudClusterProviderSpecSpec
IbmcloudClusterProviderSpecStatus 

existing now, need adjust them 

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api version:
- Minikube/KIND version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

jichenjc:
code merged

PR #27, opened 2019-04-17
Adjust provider spec (such as statusstatus, specspec etc)

jichenjc:

Part of #26 

**Release note**:

```release-note

```


pkg/apis/ibmcloud/v1alpha1/ibmcloudclusterproviderspec_types.go
```diff
@@ -20,37 +20,17 @@ import (
...
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-
-// IbmcloudClusterProviderSpec is the Schema for the ibmcloudclusterproviderspecs API
+// IbmcloudClusterProviderSpec defines the desired state of IbmcloudClusterProvider
 // +k8s:openapi-gen=true
 type IbmcloudClusterProviderSpec struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-
-	Spec   IbmcloudClusterProviderSpecSpec   `json:"spec,omitempty"`
```
gyliu513: I think we still need this as https://github.com/multicloudlab/cluster-api-provider-ibmcloud/blob/master/cmd/clusterctl/definition/ibmcloud/cluster.yaml.template#L12-L15 , otherwise, how to define the cloud provider？

```
    providerSpec:
      value:
        apiVersion: "ibmcloudclusterproviderconfig/v1alpha1"
        kind: "IbmcloudClusterProviderSpec"
```

jichenjc: this is defintion of openstack... we don't have 'cluster wide' definition yet, we need define this but not in this PR..

type OpenstackClusterProviderSpec struct {
        metav1.TypeMeta   `json:",inline"`
        metav1.ObjectMeta `json:"metadata,omitempty"`

        // NodeCIDR is the OpenStack Subnet to be created. Cluster actuator will create a
        // network, a subnet with NodeCIDR, and a router connected to this subnet.
        // If you leave this empty, no network will be created.
        NodeCIDR string `json:"nodeCidr,omitempty"`
        // DNSNameservers is the list of nameservers for OpenStack Subnet being created.
        DNSNameservers []string `json:"dnsNameservers,omitempty"`
        // ExternalNetworkID is the ID of an external OpenStack Network. This is necessary
        // to get public internet to the VMs.
        ExternalNetworkID string `json:"externalNetworkId,omitempty"`

        // ManagedSecurityGroups defines that kubernetes manages the OpenStack security groups
        // for now, that means that we'll create two security groups, one allowing SSH
        // and API access from everywhere, and another one that allows all traffic to/from
        // machines belonging to that group. In the future, we could make this more flexible.
        ManagedSecurityGroups bool `json:"managedSecurityGroups"`

        // Tags for all resources in cluster
        Tags []string `json:"tags,omitempty"`

        // Default: True. In case of server tag errors, set to False
        DisableServerTags bool `json:"disableServerTags,omitempty"`
}

gyliu513: Got it, thanks @jichenjc

Issue #43, opened 2019-04-23
clusterctl


xunpan:
/kind feature

**Describe the solution you'd like**
`clusterctl` does not work. This is logged to trace issues for `clusterctl`

xunpan:
it works now after #72 merged

PR #45, opened 2019-04-23
clusterctl : register ibmcloud and fix machine template issue

xunpan:
This is part of #43 

`provider-component.yaml` is better to have before following up `clusterctl` fix.

### DIFF
--- a/pkg/cloud/ibmcloud/deployer.go
+++ b/pkg/cloud/ibmcloud/deployer.go
@@ -1,0 +1,87 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


xunpan:
controller does not need to register "ibmcloud". So, I remove it.

pkg/cloud/ibmcloud/deployer.go
```diff
@@ -0,0 +1,87 @@
...
+	clustercommon.RegisterClusterProvisioner(ProviderName, NewDeploymentClient())
+}
+
+type DeploymentClient struct{}
+
+func NewDeploymentClient() *DeploymentClient {
+	return &DeploymentClient{}
+}
+
+func (d *DeploymentClient) GetIP(cluster *clusterv1.Cluster, machine *clusterv1.Machine) (string, error) {
```
gyliu513: @xunpan can you please add some comments here, like 

```
// TODO: Need to consolidate this with https://github.com/multicloudlab/cluster-api-provider-ibmcloud/blob/master/pkg/cloud/ibmcloud/actuators/machine/actuator.go#L161-L208
```

@jichenjc comments?

jichenjc: sorry, i had a talk with @xunpan just now, and saw this after merge action
let's keep discussing and add additional PR if needed

PR #51, opened 2019-04-24
add dep ensure into makefile

jichenjc:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,15 +2,26 @@
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -65,5 +66,8 @@ docker-build: test
 docker-push:
 	docker push ${IMG}
 
+depend:
+	dep ensure
```
gyliu513: I think you need some logic as https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/Makefile#L45-L47

Issue #19, opened 2019-04-15
userdata creation (for k8s cluster startup)


jichenjc:
/kind feature
need provide a set of configuration scripts for the new deployed guest to run this 
in order to boot cluster and join cluster for master/node 

cloud-init will be primary focus for this PR

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

gyliu513:
Sub tasks:
- [ ] Investigate how user-data works in SL
- [ ] Hard-code user-date for master and work
- [ ] Update user-data to be more generic

jichenjc:
The user-data actually need to be delivered into at least ubuntu and rhel based ...

jichenjc:
@xunpan do you want to take this one? this might take some time ... or you may select your prefer by assign your self to other issues...

xunpan:
I will start from Ubuntu with a Kubernetes deployment scripts

xunpan:
Finished scripts. 
I will try to understand IBM Cloud bootstrap script execution and try to `Create` machine from code.

xunpan:
update script based on runtime information
https://github.com/multicloudlab/cluster-api-provider-ibmcloud/pull/39

To make script run in softlay, we need a script file hosting place. Currently, I'd like to implement a github one.

xunpan:
Try to understand how to use `userData` to execute a script in softlayer. 
A ticket is logged to trace the related information. https://github.com/softlayer/softlayer-go/issues/133

xunpan:
The related part should be done.
After we have related yaml files generated, we can test it. Just close it.

PR #56, opened 2019-04-25
deploy K8s when VM is started up

xunpan:
It seems we will deploy a cluster soon (with bug fix of course because I did not test related part before)


xunpan:
Fixes #19

PR #60, opened 2019-04-26
Generate

xunpan:
1. example to  clouds.yaml 
2. generate userdata


cmd/clusterctl/configuration/ibmcloud/provider-component/user-data/ubuntu/templates/master-user-data.sh
```diff
@@ -0,0 +1,163 @@
...
+curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+touch /etc/apt/sources.list.d/kubernetes.list
+sh -c 'echo "deb http://apt.kubernetes.io/ kubernetes-xenial main" > /etc/apt/sources.list.d/kubernetes.list'
+apt-get update -y
+apt-get install -y \
+    prips
+
+# Getting master public ip 
+# TODO: is there any general way to get IP address in IBM Cloud?
+# e.g. Openstack: curl --fail -s http://169.254.169.254/2009-04-04/meta-data/local-ipv4
```
jichenjc: a set of openstack stuffs in those files, will this be changed in this commit or you prefer to use next PR to handle?

cmd/clusterctl/configuration/ibmcloud/clouds.yaml.template
```diff
@@ -1,3 +1,2 @@
-clouds:
-  username: <username>
-  toke: <toke>
+userName: "your-name"
```
gyliu513: I think we should keep the format as follows to enable future extension.

```
clouds:
  ibmcloud:
    auth:
      username: "test@test.com"
      password: "here'syourpassword"
```

cmd/clusterctl/configuration/ibmcloud/provider-component/user-data/ubuntu/templates/master-user-data.sh
xunpan: This is just an example to show that we may need the common way as `Openstack` to get host API. I will handle it if `TODO` is done or removed.

cmd/clusterctl/configuration/ibmcloud/clouds.yaml.template
xunpan: I will implement it when the whole creation path succeed.

PR #63, opened 2019-04-26
make creation path work

xunpan:
This will have several commits to make Machine `Create` path work for ibmcloud


xunpan:
Test passed. VM instances of Kubernetes master is created with `Machine` resource. 

Next step, I will install Kubernetes cluster on the node.

PR #69, opened 2019-04-29
specified ssh user and fix script run issue

xunpan:
1. `sshUserName`
2. move `userData` script into `/` to avoid it is executed twice. I need this commit to continue the development.


cmd/clusterctl/configuration/ibmcloud/cluster.yaml.template
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
     clusterNetwork:
         services:
-            cidrBlocks: ["10.96.0.0/12"]
+            cidrBlocks: ["10.224.0.0/12"]
```
gyliu513: Why update here? This is only for service IP, I think we can keep here unchanged?

PR #81, opened 2019-04-30
klog messages

### LINKED ISSUE #64: use klog instead of log
/kind feature

cluster-api uses `klog`. So we should also use it to avoid mixed log functions

xunpan:
Fixes #64

### DIFF
--- a/pkg/cloud/ibmcloud/actuators/cluster/actuator.go
+++ b/pkg/cloud/ibmcloud/actuators/cluster/actuator.go
@@ -17,9 +17,7 @@ limitations under the License.
 package cluster
 
 import (
-	"fmt"
-	"log"
-
+	"k8s.io/klog"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 	client "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1"
 )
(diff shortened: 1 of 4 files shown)


pkg/cloud/ibmcloud/actuators/cluster/actuator.go
```diff
@@ -43,13 +41,17 @@ func NewActuator(params ActuatorParams) (*Actuator, error) {
 
 // Reconcile reconciles a cluster and is invoked by the Cluster Controller
 func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) error {
-	log.Printf("Reconciling cluster %v.", cluster.Name)
-	// "TODO: Not yet implemented"
+	klog.Infof("Reconciling cluster %v.", cluster.Name)
+
+	klog.Infof("Reconciling cluster %v: Not implemented yet.", cluster.Name)
```
jichenjc: I assume you have those 2 on purpose ,right?

pkg/cloud/ibmcloud/actuators/machine/actuator.go
```diff
@@ -160,31 +159,32 @@ func (ic *IbmCloudClient) Delete(ctx context.Context, cluster *clusterv1.Cluster
...
+	klog.Infof("Updating machine %v for cluster %v.", machine.Name, cluster.Name)
+
 	klog.Infof("TODO: Not yet implemented")
 	return nil
 }
 
 // Exists test for the existance of a machine and is invoked by the Machine Controller
 func (ic *IbmCloudClient) Exists(ctx context.Context, cluster *clusterv1.Cluster, machine *clusterv1.Machine) (bool, error) {
-	guest, err := ic.guestExists(machine)
+	_, err := ic.guestExists(machine)
```
jichenjc: I assume the function will return NOTFOUND error in err, right?

pkg/cloud/ibmcloud/clients/machines.go
```diff
@@ -73,28 +76,34 @@ func NewInstanceServiceFromMachine(kubeClient kubernetes.Interface, machine *clu
...
-		log.Print(".")
 		// TODO(gyliu513) make it configurable or use the notification mechanism to optimize
 		// the process instead of hardcoded waiting.
 		time.Sleep(5 * time.Second)
 		transactions, _ = s.GetActiveTransactions()
 	}
-	log.Println("wait done")
+	s.Session.Debug = false
+
+	klog.Info("Waiting for trasactions <%D> done.")
```
jichenjc: where's %D?

xunpan: Type. I fixed it

pkg/cloud/ibmcloud/actuators/machine/actuator.go
xunpan: This function only returns bool. It does not need `guest`. And `guestExists` always return `error` if there is no host found

pkg/cloud/ibmcloud/actuators/cluster/actuator.go
xunpan: I will change it to `TODO` for second message

PR #83, opened 2019-04-30
Example instead of configuration

xunpan:
as other cluster providers did


pkg/cloud/ibmcloud/clients/machines.go
```diff
@@ -73,28 +76,34 @@ func NewInstanceServiceFromMachine(kubeClient kubernetes.Interface, machine *clu
...
-		log.Print(".")
 		// TODO(gyliu513) make it configurable or use the notification mechanism to optimize
 		// the process instead of hardcoded waiting.
 		time.Sleep(5 * time.Second)
 		transactions, _ = s.GetActiveTransactions()
 	}
-	log.Println("wait done")
+	s.Session.Debug = false
+
+	klog.Info("Waiting for trasactions <%D> done.")
```
jichenjc: can you help to fix this ?I didn't pay attention so I closed 2 commits same time

also a typo

jichenjc: transaction

PR #85, opened 2019-04-30
function name should be verbe+none to tell the action

xunpan:
rename function 
and
fix log issues

### DIFF
--- a/pkg/cloud/ibmcloud/clients/machines.go
+++ b/pkg/cloud/ibmcloud/clients/machines.go
@@ -74,7 +74,7 @@ func NewInstanceServiceFromMachine(kubeClient kubernetes.Interface, machine *clu
 	return NewGuestService(sess), nil
 }
 
-func (gs *GuestService) guestWaitReady(Id int) {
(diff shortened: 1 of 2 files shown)


pkg/cloud/ibmcloud/clients/machines.go
```diff
@@ -163,7 +163,7 @@ func (gs *GuestService) GuestDelete(Id int) error {
 	return err
 }
 
-func (gs *GuestService) GuestList() ([]datatypes.Virtual_Guest, error) {
+func (gs *GuestService) listGuest() ([]datatypes.Virtual_Guest, error) {
```
jichenjc: make this ListGuest, it might be used somewhere later elsewhere...

xunpan: Is this a perdefined interface? 
I'd like to keep it as is if not. We can keep a function scope limited unless it is really used by others.

gyliu513: We can update it to `ListGuest` when this is requested ;-)

PR #93, opened 2019-04-30
Deleted hostname from machines.yaml template.

gyliu513:

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/clusterctl/examples/ibmcloud/machines.yaml.template
+++ b/cmd/clusterctl/examples/ibmcloud/machines.yaml.template
@@ -13,7 +13,6 @@ items:
       value:
         apiVersion: "ibmcloudproviderconfig/v1alpha1"
         kind: "IbmcloudMachineProviderSpec"
-        hostname: <host name>
         domain: <domain name>
         maxMemory: 4096
         startCpus: 1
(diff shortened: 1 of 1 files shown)


xunpan:
@gyliu513 Please rebase

cmd/clusterctl/examples/ibmcloud/machines.yaml.template
```diff
@@ -13,7 +13,6 @@ items:
       value:
         apiVersion: "ibmcloudproviderconfig/v1alpha1"
         kind: "IbmcloudMachineProviderSpec"
-        hostname: <host name>
```
jichenjc: @xunpan I asked this question and at that time seems we want to use hostname+domain as name, so why we rename this ?

xunpan: we use resource name as the host name. So, this host name is useless. Currently, this is fine for use.

And, I did not delete the internal data structure to make it a feature on demand later.

jichenjc: ok, that works for me , just want to make sure I catch up with your changes 
thanks for all the work :)

Issue #116, opened 2019-05-05
Push latest image to quay.io via CI process


gyliu513:
/kind bug
/sig ibmcloud

We should always push controller and clusterctl to quay.io via CI process.

## Controller
```
quay.io/cluster-api-provider-ibmcloud/controller
```

## Clusterctl
```
quay.io/cluster-api-provider-ibmcloud/clusterctl
```

@morvencao

xunpan:
@gyliu513 as we already in account `cluster-api-provider-ibmcloud`, shall we use name `controller` and `clusterctl` for simple. Especially for `clusterctl`, I think it is better descriptive in `docker run quay.io/cluster-api-provider-ibmcloud/clusterctl ...`

gyliu513:
@xunpan +1 to use simple name, you can create PR to fix this.

jichenjc:
+1 to the simple naming given the namespace we added

gyliu513:
@xunpan can you help check how federation v2 push the latest image to `quay.io`? We can follow same way for ibmcloud.

/assign @xunpan

xunpan:
federation push image to `quay.io` when CI succeed. It pushes images for `kubernetes-multicluster/federation-v2` with `canary`, `latest` and tag version, e.g. `v0.0.9`.

gyliu513:
@xunpan Yes, shall we follow same logic for ibmcloud images?

xunpan:
@gyliu513  I'm trying to understand how to set credential to access `quay.io` in travis. After that, I will fix this issue.

gyliu513:
>  how to set credential to access quay.io

Yes, this is exactly what we should do.

morvencao:
@xunpan I think we need to put the credential in k8s secret that prow jobs can access.

jichenjc:
only want to make sure my understanding, we will push the image when 
CI passed and merged into master or just CI passed and push to registry for test purpose?
both works for me but I want to make sure I understand the correct way, thank you @morvencao @gyliu513 @xunpan

xunpan:
The logic is pushing image after CI passed for `canary` image. The code at the moment is not merged yet. CI is prerequisite for merge but not do code merge automatically.

For release versioned and latest (latest release one), they are only push by a tag (for release).

xunpan:
@morvencao I guess it is not for this case to pass credential manually by k8s secret. It is better to set it in security place and follow up existing practise.

[...]

PR #118, opened 2019-05-05
use simple image name

xunpan:
as discussed in #116, we use simple name instead of full name because it is already under ibm account.


gyliu513:
/cc @jichenjc 

LGTM

PR #119, opened 2019-05-06
make independent binary image build

xunpan:
1. the target name should better be noun 
  - `build-images` -> `images`. `make images` seems better than `make build-images`. BTW, `all` target has dependecy to `images` already.
  -  keep `push-images` (and push*) as is because it is more descriptive
2. add single binary image build and push
  For development work, a single controller build and push is most common operation. I need it to refresh image quickly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,21 +150,24 @@ deploy: generate_yaml_test
 # images section
 ############################################################
 # Build the docker image
-build-images: test
-	docker build . -f cmd/manager/Dockerfile -t $(REGISTRY)/$(CONTROLLER_IMG):$(VERSION)
+clusterctl-image:
 	docker build . -f cmd/clusterctl/Dockerfile -t $(REGISTRY)/$(CLUSTERCTL_IMG):$(VERSION)
-
-# Push the docker image
-push-images:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -150,21 +150,24 @@ deploy: generate_yaml_test
...
 	docker push $(REGISTRY)/$(CLUSTERCTL_IMG):$(VERSION)
-
-build-push-images:
-	docker build . -f cmd/manager/Dockerfile -t $(REGISTRY)/$(CONTROLLER_IMG):$(VERSION)
+push-controller-image:
 	docker push $(REGISTRY)/$(CONTROLLER_IMG):$(VERSION)
-	docker build . -f cmd/clusterctl/Dockerfile -t $(REGISTRY)/$(CLUSTERCTL_IMG):$(VERSION)
-	docker push $(REGISTRY)/$(CLUSTERCTL_IMG):$(VERSION)
+
+images: test cmd-image controller-image
```
gyliu513: s/cmd-image/clusterctl-image

/cc @morvencao

morvencao: I need to update the prow jobs in https://github.com/kubernetes/test-infra/pull/12479

Issue #103, opened 2019-05-03
display correct message for kubectl get machine


xunpan:
/kind feature

**Describe the solution you'd like**
```
$ kubectl --kubeconfig kubeconfig get machine     
NAME                    PROVIDERID   PHASE
ibmcloud-master-5rcv6                
ibmcloud-node-cvww2                  
```
Empty display is not good.
we also need `AGE` as cluster

jichenjc:
if I remember correctly, the phase and providerid is optional and should not be displayed by `get machine`  command, it comes from 
https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/apis/cluster/v1alpha1/machine_types.go#L41

and for providerid , @xunpan you can refer to 
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/274/commits

I think `phase` is a little bit hard to set value..

jichenjc:
@xunpan @gyliu513 do we have a way to identical identify a guest on ibmcloud? 
such as we can use openstack:////<uuid> as openstack identifier, what do we want to ibmcloud?

gyliu513:
not sure about this, @xunpan , can you help check?

xunpan:
I'm not sure if this is an convention or not. I'd like to follow the way you mentioned for `openstack`.
 - `aws` provider uses `aws:////instanceid`
 - `azure` provider uses `azure:////VMID`

 For IBM Cloud, I plan to use `ibmcloud:////guestid`

jichenjc:
ok, let's use ibmcloud for now, thanks for bring this out

PR #121, opened 2019-05-06
add provider id

xunpan:
This is part of issue #103

### DIFF
--- a/pkg/cloud/ibmcloud/actuators/machine/actuator.go
+++ b/pkg/cloud/ibmcloud/actuators/machine/actuator.go
@@ -133,7 +133,7 @@ func (ic *IbmCloudClient) Create(ctx context.Context, cluster *clusterv1.Cluster
 		return fmt.Errorf("Guest %s does not exist after created in cluster %s", machine.Name, cluster.Name)
 	}
 
-	return ic.updateAnnotation(machine, strconv.Itoa(*guest.Id))
+	return ic.updateMachine(machine, strconv.Itoa(*guest.Id))
 }
 
 // Delete deletes a machine and is invoked by the Machine Controller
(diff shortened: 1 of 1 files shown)


pkg/cloud/ibmcloud/actuators/machine/actuator.go
```diff
@@ -176,6 +176,13 @@ func (ic *IbmCloudClient) Exists(ctx context.Context, cluster *clusterv1.Cluster
 		return false, err
 	}
 
+	if (guest != nil) && (machine.Spec.ProviderID == nil || *machine.Spec.ProviderID == "") {
+		// TODO: this does not work in ibm cloud
```
gyliu513: s/TODO/TODO(xunpan) for better trace.

pkg/cloud/ibmcloud/actuators/machine/actuator.go
```diff
@@ -176,6 +176,13 @@ func (ic *IbmCloudClient) Exists(ctx context.Context, cluster *clusterv1.Cluster
 		return false, err
 	}
 
+	if (guest != nil) && (machine.Spec.ProviderID == nil || *machine.Spec.ProviderID == "") {
+		// TODO(xunpan): this does not work in ibm cloud
```
jichenjc: `this does not work in ibm cloud`
you mean this function not work or not work in Exists function ?maybe open a futher issue to track this ?

xunpan: This code is found in other providers. But it does not work in ibmcloud: with this code, the `ProviderID` is not updated as expected.

My code in `Create` works. It updates `ProviderID`. I keep code here to try to understand if there has any misunderstanding in updating `ProviderID` in `Exists`. This is not a issue but need further understanding of the code. The full code logic should works for any case because I check the `ProviderID` if it is specified already.

jichenjc: thank you for the detailed info, really appreciate, that works for me

Issue #16, opened 2019-04-15
machine actuator implementation


jichenjc:
/kind feature
we need add Create/Update/Delete/Exists functions to machine so that we can make it able to be consumed by cluster-api Reconcile function 

we can create several patches to add those functions

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

jichenjc:
This can be enhanced to be 
Create
Delete
Update
Exists
GetIP
GetKubeConfig

let's add one by one

jichenjc:
before that, we need something like 
ClusterStatusFromProviderStatus
ClusterSpecFromProviderSpec

so we can consume the status and spec from cluster-api.....

gyliu513:
@xunpan will work for `update`.

FYI @jichenjc

xunpan:
Discussion about @jichenjc 's comments
https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/128/files/8d9a595c053465a9bb00cffbfbd47c6dc5edf32d#r282150120
> why we have pending status? if you see line 90 this means for already created instance you create it again
> you will mark as pending? I think just creating is enough ? if you agree I can create a PR to update it @xunpan

### Common Case
Currently, the behavior is:
If a new machine is created, there should not be a VM instance already. So that API can create a new VM instance with bootstrap configuration. 

### Existing Case
I guess there has logic issue here, if there has an exist VM instance. We should:
  - report error: we will not update a existing machine. Then we should delete this 'Pending' status.
  - currently log info, and `Pending` in status. We can use `update` operation to delete existing VM instance (better to check if it meets our spec before deletion) and create a new one. 

I think option 2 should be better. Because we create name of vm instance which should be not duplicated in cloud. Even if it is duplicated (e.g. Machine transfer case, or the VM is created manually, or by other cluster-api instances), we should identify this machine is managed by cluster-api (better to know current cluster instance ID I guess) or not. 
However, there is still many things need to be investigation. And it is my thought for the behavior of `Existing Case`. Any commnets for this @jichenjc @gyliu513 

If update is needed, refer below thoughts
### Update
I'm considering how to implement update. This is logic in my mind:
1. support compute node
    - check if anything changed for VM, delete and recreate if exists
    - check if K8s status is ok, if not, try to run start up scripts in the machine after a TTL.
2. support master node
    - we need related operations as `compute node`
[...]

PR #150, opened 2019-05-17
update: when a machine specification changes, recreate the machine

xunpan:
update. basic functions

Related to  #16

/cc @gyliu513 @jichenjc

It is tested. However, I found an issue in `clusterctl` steps. I'd like to make sure if it is related to this issue.

/sig ibmcloud

### DIFF
--- a/pkg/cloud/ibmcloud/actuators/machine/actuator.go
+++ b/pkg/cloud/ibmcloud/actuators/machine/actuator.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/softlayer/softlayer-go/datatypes"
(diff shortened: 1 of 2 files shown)


pkg/cloud/ibmcloud/actuators/machine/actuator.go
```diff
@@ -179,8 +180,59 @@ func (ic *IbmCloudClient) Delete(ctx context.Context, cluster *clusterv1.Cluster
...
+		//   - ssh key name
+		//   - host name
+		return nil
+	}
+
+	klog.Infof("Recreating VM instances for machine %v of cluster %v.", machine.Name, cluster.Name)
+	err = ic.Delete(ctx, cluster, machine)
+	if err != nil {
+		return err
+	}
```
jichenjc: just curious, do IBM cloud wait for delete complete then send back notification? 
openstack used to have a bug that after delete, it's still there because of async then create again with similar info lead to issue...

xunpan: Good point. I encountered a Deletion issue during my test. Not sure of it yet. It may cause the `cluterctl` issue I mentioned above.

xunpan:
After this fix, the update operation works. You can change `Machine` and controller will delete old VM instance and create a new VM instance.
However, controller reports error messages during the period. E.g. the deleted VM instance need time to GC and we still need `Update` return error and retry later. I will check this part after this PR merged.

@gyliu513 @jichenjc

jichenjc:
/lgtm

let's merge this and enhance later, thanks for the update

PR #151, opened 2019-05-17
get vm by name instead of search in a list

xunpan:
make code logic simple and clear. 

/cc @gyliu513 @jichenjc 

Still need to do below things in this PR before review
 - [x] cover domain filter
 - [x] test function

### DIFF
--- a/pkg/cloud/ibmcloud/clients/machines.go
+++ b/pkg/cloud/ibmcloud/clients/machines.go
@@ -168,16 +168,17 @@ func (gs *GuestService) DeleteGuest(Id int) error {
 	return err
 }
 
-func (gs *GuestService) GetGuest(name string) (*datatypes.Virtual_Guest, error) {
+func (gs *GuestService) GetGuest(name, domain string) (*datatypes.Virtual_Guest, error) {
 	s := services.GetAccountService(gs.sess)
 
 	hostFilter := filter.Build(
 		filter.Path("virtualGuests.hostname").Eq(name),
+		filter.Path("virtualGuests.domain").Eq(domain),
 	)
 
 	guests, err := s.Filter(hostFilter).GetVirtualGuests()
 	if err != nil {
-		klog.Errorf("Error getting virtual guests: %v", err)
+		klog.Errorf("Error getting virtual guests by filter (hostname=%s, domain=%s): %v", name, domain, err)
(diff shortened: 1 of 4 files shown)


pkg/cloud/ibmcloud/clients/machines.go
```diff
@@ -176,32 +177,30 @@ func (gs *GuestService) DeleteGuest(Id int) error {
...
-	for _, guest := range guests {
-		// FIXME: how to unique identify one guest
-		if *guest.Hostname == name {
-			klog.Infof("Found guest with ID %d for %q", *guest.Id, name)
-			return &guest, nil
-		}
+	if len(guests) > 1 {
+		// I noticed that IBM Cloud can use same name for 2 machines.
+		// It is bad for our case. Print a message to make it to be noticed.
+		klog.Errorf("Getting more than one virtual guest. The first is used.")
```
gyliu513: Print out the guest name here for better debug.

pkg/cloud/ibmcloud/clients/machines.go
```diff
@@ -176,32 +177,32 @@ func (gs *GuestService) DeleteGuest(Id int) error {
...
-	for _, guest := range guests {
-		// FIXME: how to unique identify one guest
-		if *guest.Hostname == name {
-			klog.Infof("Found guest with ID %d for %q", *guest.Id, name)
-			return &guest, nil
-		}
+	if len(guests) > 1 {
+		// I noticed that IBM Cloud can use same name for 2 machines.
+		// It is bad for our case. Print a message to make it to be noticed.
+		klog.Errorf("Getting more than one virtual guests by filter (hostname=%s, domain=%s): %v. The first is used.",
```
gyliu513: How about `The first host %s is used`?

pkg/cloud/ibmcloud/clients/machines.go
xunpan: I use guest ID because names are the same when issue happens.

xunpan:
addressed comments and rebase done

@gyliu513

pkg/cloud/ibmcloud/clients/machines.go
xunpan: I use guest ID because names are the same when issue happens.

PR #172, opened 2019-05-24
Remove ssh user for node

jichenjc:
we should not and never SSH into nodes in our solution...
temply allow ssh into master is acceptable.

THis remove the sshkey/sshuser in node template

Fixes #

/sig ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/cmd/clusterctl/examples/ibmcloud/machines.yaml.template
+++ b/cmd/clusterctl/examples/ibmcloud/machines.yaml.template
@@ -47,8 +47,6 @@ items:
         osReferenceCode: <the operating system selection>
         localDiskFlag: true
         hourlyBillingFlag: true
-        sshKeyName: cluster-api-provider-ibmcloud
-        sshUserName: <ssh user name>
         userDataSecret:
           name: worker-user-data
           namespace: ibmcloud-provider-system


cmd/clusterctl/examples/ibmcloud/machines.yaml.template
```diff
@@ -47,8 +47,6 @@ items:
         osReferenceCode: <the operating system selection>
         localDiskFlag: true
         hourlyBillingFlag: true
-        sshKeyName: cluster-api-provider-ibmcloud
```
xunpan: This should be kept. It is used to create VM instances.

jichenjc: why?? do we really need this for worker node ?(not master node)

gyliu513: Sometimes the end user may also want to logon to the worker node to do some debug, comments?

jichenjc: that's debug phase, in our template, we should not encourage people to do so and
if they want , they are free to do..

xunpan: If this is not kept, the created VM cannot be accessed by anyone after it is created.  I prefer to use it by default because it helps if I need to login to the host for any cases. It is harmless. Any comments?

jichenjc: ok, I think it's only for debug purpose...anyway, let's ignore this and see whether we really have some request (in our IaaS dev, we recommend no ssh for worker node...)

PR #175, opened 2019-05-24
make .gitignore easy to maintain

xunpan:
don't mess up .gitignore
Let's keep it simple and easy to maintain


.gitignore
```diff
@@ -1,145 +1,23 @@
...
 
-# OSX leaves these everywhere on SMB shares
-._*
-
-# OSX trash
-.DS_Store
-
-# Eclipse files
-.classpath
-.project
```
gyliu513: Need keep .project as it will be generated by eclipse.

.gitignore
```diff
@@ -1,145 +1,23 @@
-*.json
-*.sublime-project
-*.sublime-workspace
+# temporay files
 *.swp
-.idea
-.DS_Store
```
gyliu513: Keep .DS_Store for mac os.

.gitignore
```diff
@@ -1,145 +1,23 @@
...
-.classpath
-.project
-.settings/**
-
-# Files generated by JetBrains IDEs, e.g. IntelliJ IDEA
-.idea/
-*.iml
-
-# Vscode files
-.vscode
```
gyliu513: Keep .vscode for VSCode.

.gitignore
```diff
@@ -1,145 +1,23 @@
...
-# OSX trash
-.DS_Store
-
-# Eclipse files
-.classpath
-.project
-.settings/**
-
-# Files generated by JetBrains IDEs, e.g. IntelliJ IDEA
-.idea/
```
gyliu513: Keep this for IDEA.

xunpan:
Actually, I'm thinking what this file is used for in best practise.

 - For project related files: generated, temprary files, etc. We need to ignore it for sure.
 - For editor/OS related files, do we need to keep it in `.gigignore` for all?
   At first, I guess so because it helps developers. https://github.com/kubernetes-sigs/kubefed/pull/871.
However, after thinking it over, I think developer specific files should better not in this file. It does not help all development process. Developer maintains their own environment seems better because they have their own editor preferences: https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore.

Any comments?

gyliu513:
Thanks @xunpan make sense.

/lgtm

PR #179, opened 2019-05-24
remove uselsess id annotation, we have it in PROVIDERID

xunpan:
clean duplicated information


jichenjc:
make sense, we should use provider ID ,however, will it be helpful to some other functions to use the annotation such as cloud provider? do know have the knowledge defined in the cluster-api CRD?

xunpan:
`providerID` is maintained in `cluster-api` `Machine` definition.

jichenjc:
I knew it's maintained in cluster-api, will the CRD of machine defined in cluster-api or other cloud relaetd project know / has capability to access the provider ID info, right?

xunpan:
No matter `annotation` or `providerID`, they are part of `Machine` data. They share same RBAC. If you can query it, you can get informaiton from both. It has no different I think.

jichenjc:
ok, thanks for the info , really appreciate , I might have wrong thought here..

/lgtm

PR #188, opened 2019-05-27
the generate.sh exited when the cluster.yaml exists

hchenxa:

below code will exited when the cluster.yaml exists.
```
# Set up the output dir if it does not yet exist
mkdir -p $PWD/$OUTPUT
cp -n $PWD/cluster.yaml.template $PWD/$OUTPUT/cluster.yaml  >>>>>>>> The return code was 1
cp -n $PWD/machines.yaml.template $PWD/$OUTPUT/machines.yaml 
```

so in order to avoid the file overwrite and keep the updated file of cluster.yaml and machines.yaml, create this PR.

Fixes #

/cc @gyliu513 @clyang82 

/sig ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/cmd/clusterctl/examples/ibmcloud/generate-yaml.sh
+++ b/cmd/clusterctl/examples/ibmcloud/generate-yaml.sh
@@ -112,8 +112,8 @@ cat "$WORKER_USER_DATA" > $USERDATA/$PROVIDER_OS/worker-user-data.sh
 
 # Set up the output dir if it does not yet exist
 mkdir -p $PWD/$OUTPUT
-cp -n $PWD/cluster.yaml.template $PWD/$OUTPUT/cluster.yaml
-cp -n $PWD/machines.yaml.template $PWD/$OUTPUT/machines.yaml
+cp -n $PWD/cluster.yaml.template $PWD/$OUTPUT/cluster.yaml || true
+cp -n $PWD/machines.yaml.template $PWD/$OUTPUT/machines.yaml || true
 
 # Build provider-components.yaml with kustomize
 kubectl kustomize $PWD/../../../../config > $PWD/$OUTPUT/provider-components.yaml


gyliu513:
/lgtm
/approve

You can post a PR to openstack provider as well to fix https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/cmd/clusterctl/examples/openstack/generate-yaml.sh#L143-L144

cmd/clusterctl/examples/ibmcloud/generate-yaml.sh
```diff
@@ -112,8 +112,8 @@ cat "$WORKER_USER_DATA" > $USERDATA/$PROVIDER_OS/worker-user-data.sh
 
 # Set up the output dir if it does not yet exist
 mkdir -p $PWD/$OUTPUT
-cp -n $PWD/cluster.yaml.template $PWD/$OUTPUT/cluster.yaml
-cp -n $PWD/machines.yaml.template $PWD/$OUTPUT/machines.yaml
+cp -n $PWD/cluster.yaml.template $PWD/$OUTPUT/cluster.yaml || true
+cp -n $PWD/machines.yaml.template $PWD/$OUTPUT/machines.yaml || true
```
jichenjc: should we always overwrite the file?

gyliu513: I think we should not overwrite, as the overwrite will use template to overwrite which will clear some useful data.

hchenxa: the `cluster.yaml` and `machines.yaml` only contain some cluster info and machine info which should matainered by cluster deployer self, so I think we can only overwrite the `provider-components.yaml` when use overwrite flag.

cc @gyliu513 comments?

gyliu513: `provider-components.yaml` will always overwrite, this is the current logic already. I think the current logic is ok. @hchenxa @jichenjc

xunpan:
FYI: this should be only MacOS issue. Linux works fine based on test.

jichenjc:
ok, good to know :)

Issue #154, opened 2019-05-20
Check if we can learn from vsphere provider for image CI


gyliu513:
/kind bug
/sig ibmcloud

Please refer to https://github.com/kubernetes/cloud-provider-vsphere/tree/master/hack/images/ci to see if this helps?

/assign @xunpan

xunpan:
It shows a way to push image on go `gcr.io`. It uses secret `cloud-provider-vsphere-e2e-config` to store the credential for login before push. Again, not know how they create and keep credential private.

jichenjc:
can we post some question in their slack channel or submit a issue to ask for additional info? I think it's important for us to have e2e test in long term

gyliu513:
You can also open an issue in vsphere cloud provider repo to ask

jichenjc:
I opened https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/268
but not sure I am asking right thing, @gyliu513 @xunpan please add more info if I have some misunderstanding

jichenjc:
@xunpan can you talk to sig-testing as kubernetes-sigs/cluster-api-provider-vsphere#268 has some comments and I am not very familiar with this .. thanks

gyliu513:
Working at https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/203#issuecomment-496895135

Issue #202, opened 2019-05-29
Enable daily build on master


clyang82:
/kind bug
/sig ibmcloud

Right now, we have some code changes on master branch so that the v0.0.1 image cannot be working. we have to setup the daily build on master.

FYI @gyliu513 @jichenjc

xunpan:
Is this duplidated of https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/154?

clyang82:
Close it since the PR was merged.

PR #203, opened 2019-05-29
Add travis for building

clyang82:
This PR is used to enable travis to generate daily image for `manager` and `clusterctl` from master. refer to #202 

Signed-off-by: Chun Lin Yang <clyang@cn.ibm.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
 GIT_HOST = sigs.k8s.io
 PWD := $(shell pwd)
 BASE_DIR := $(shell basename $(PWD))
+# customize kubebuilder path
+KUBEBUILDER_PATH ?= /usr/local
+export KUBEBUILDER_ASSETS=$(KUBEBUILDER_PATH)/kubebuilder/bin
+# customize kubectl path
+KUBECTL_PATH ?= /usr/local/bin
 # Keep an existing GOPATH, make a private one if it is undefined
 GOPATH_DEFAULT := $(PWD)/.go
 export GOPATH ?= $(GOPATH_DEFAULT)
@@ -32,8 +37,7 @@ LDFLAGS   := "-w -s -X 'main.version=${VERSION}'"
 # Image URL to use all building/pushing image targets
 CONTROLLER_IMG ?= controller
 CLUSTERCTL_IMG ?= clusterctl
-VERSION ?= $(shell git describe --exact-match 2> /dev/null || \
-                 git describe --match=$(git rev-parse --short=8 HEAD) --always --dirty --abbrev=8)
+
 REGISTRY ?= quay.io/cluster-api-provider-ibmcloud
 
 ifneq ("$(realpath $(DEST))", "$(realpath $(PWD))")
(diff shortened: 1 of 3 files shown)


xunpan:
Could you help to share the reason to add travis-ci back? I remember we replace it before with test-infra.

clyang82:
> Could you help to share the reason to add travis-ci back? I remember we replace it before with test-infra.

I want to use travis to generate daily build from master. Can test-infra do it?

gyliu513:
@xunpan Seems we can use both travis and test-infra together, https://github.com/kubeflow/katib is an example which use both travis and test-infra.

xunpan:
I just want to make sure we do not make duplication work. 

If we use this for image push, I'm ok with it. Because currently, test-infra cannot push image. Based on my latest information from community. The security information privacy is still under discussion.  And no further feedback from community. 

If this is fine, I think @gyliu513 we can close https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/154. Because it makes less sense anymore

clyang82:
I think it is ready to merge. Encrypted environment variables have been removed for security reasons in forked repo - https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions.
I have changed the travis to run as Cron job daily on master. any pull requests won't trigger it.

xunpan:
@clyang82 
In federation v2, the QUAY name and password is set as environment variables in travis CI. It is used by CI task `in after_success`
```
after_success:
- test $DOC_ONLY && echo "Skipped!" || make push
```
Did you mean this has security issue?

config/manager/manager.yaml
```diff
@@ -55,7 +55,7 @@ spec:
         operator: Exists
       containers:
       - name: ibmcloud-machine-controller
-        image: quay.io/cluster-api-provider-ibmcloud/controller:v0.0.1
+        image: quay.io/cluster-api-provider-ibmcloud/controller:master
```
xunpan: as a version, `latest` seems better than `master`

gyliu513: +1 to use latest.

clyang82:
> @clyang82
> In federation v2, the QUAY name and password is set as environment variables in travis CI. It is used by CI task `in after_success`
> 
> ```
> after_success:
> - test $DOC_ONLY && echo "Skipped!" || make push
> ```
> 
> Did you mean this has security issue?
> 
> more info:
> `make push` above will login before push image.

It seems the DOC_ONLY is always set to `true` for PR. And in the below pic from federation-v2, encrypted environment variables have been removed.
image

config/manager/manager.yaml
clyang82: Done

xunpan:
FYI: had an offline sync-up with @clyang82 before. We're sure that `DOC_ONLY` from the references are set on demand. And @clyang82 will follow his own new method to see if it works.

Makefile
```diff
@@ -1,6 +1,11 @@
 GIT_HOST = sigs.k8s.io
 PWD := $(shell pwd)
 BASE_DIR := $(shell basename $(PWD))
+# customize kubebuilder path
+KUBEBUILDER_PATH ?= /usr/local
```
xunpan: we've already have download scripts for all dependency binaries. Is it possible to reuse it instead of managing a new download method here?

xunpan:
After discuss with @clyang82, we will merge this PR first to check if it works. And the review comments will be handled in later PR.

/lgtm
/approve

Issue #205, opened 2019-05-30
simplify the build and use of cluster-api


chenzhiwei:
/kind feature
/sig ibmcloud

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

PR #206, opened 2019-05-30
simplify the steps with make targets

chenzhiwei:

Simplify the steps, easy to generate yaml file and create cluster.

Related to #205 

/sig ibmcloud

**Release note**:

```release-note
NONE
```


jichenjc:
The idea is good, the question is we have a set of documents seems contradict to the new stuffs and new doc, will those be updated later or in this PR?

chenzhiwei:
yes, this is just the first one.

i will create a series of prs.

.gitignore
```diff
@@ -1,6 +1,9 @@
 # temporay files
 *.swp
 
+# env.sh
+/env.sh
```
xunpan: I guess you only want to remove `env.sh` under root directory. What about using same style as existing path: start with no`/`.

Same comments for `_output`

cmd/clusterctl/examples/ibmcloud/machines.yaml.template
```diff
@@ -7,50 +7,51 @@ items:
...
     providerSpec:
       value:
         apiVersion: "ibmcloudproviderconfig/v1alpha1"
         kind: "IBMCloudMachineProviderSpec"
-        domain: <domain name>
-        maxMemory: 4096
-        startCpus: 4
-        dataCenter: <datacenter name>
-        osReferenceCode: <the operating system selection>
+        domain: CONTROL_PLANE_DOMAIN
```
xunpan: what about `s/CONTROL_PLANE/MACHINE/`? control plane makes less sense for machines I guess.

xunpan:
I have a qustion.
What do you think to add `env.sh.template`? The related variable seems not shared, and the full capitalized variable seems not easy to read than origional yaml file configuration. What is the benefit if we change it to this way?

.gitignore
chenzhiwei: actually other patterns should add a '/' prefix.

without `/` prefix will ignore all files named with `env.sh`.

chenzhiwei:
capitalize the environment variable is a shell practice, if you run `env` you will see all environment variables in the OS are upper case letters.

adding dev.sh.template is used to make configuration simple, users only need to modify a single file and less variables to achieve the same result.

chenzhiwei:
will do a new refactor to not disturb the current method.

chenzhiwei:
@gyliu513 @jichenjc @xunpan You can now review this new change.

This change has no impact on existing logic.

jichenjc:
@chenzhiwei this is a big change so maybe need some time 

and If I didn't miss something, at least https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud (the html on README is conflict to your change), I think we need fix it as well

And @xunpan @gyliu513 we said we want to keep consistent to other providers before
for this change, I think we don't need be consistent as this brings value (sorry, I don't have time to try myself but I think @chenzhiwei will share more info on the benefit of the PR), thoughts?

chenzhiwei:
You can check here: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/811bdf5e590aa9bc3a48182d7500f93b3b8f1f01/docs/getting-started.md

This will significantly simplify the steps to generate yaml files and create clusters.

jichenjc:
I fully understand :) the point is you have to clear the old stuffs as those 2 are conflict

chenzhiwei:
The new and old are just duplicated, not conflict.

This is a series of PRs, currently I will maintain two methods, when all the changes are made, then I think we can have a PR to remove the old one.

I will also propose similar changes to other providers later when ibmcloud adopts this new method.

jichenjc:
ok to me 
/lgtm

docs/getting-started.md
```diff
@@ -0,0 +1,103 @@
+# Getting Started
```
gyliu513: Please run `doctoc` to generate a toc.

docs/getting-started.md
```diff
@@ -0,0 +1,103 @@
...
+If you already have an SSH key in IBM Cloud, you need to put the SSH private key to `$HOME/.ssh/id_ibmcloud`.
+
+If you don't have an SSH key, you can generate one with `make ssh-key` locally and create it on IBM Cloud.
+
+
+## Cluster Creation
+
+1. Create an environment file and fill in your IBM Cloud info
+
+    ```
```
gyliu513: highlight `shell` by adding `shell` to the end. Ditto for other shell commands.

xunpan:
Captalized environment variables is fine. It is convention. But, I'm not fully conviced that this method is more convenient by using environment variables like names.  
 - less sharing in environment configuartion. It almost each name for a field in yaml file value.
 - capitalized name configuration in one file VS. [name, descriptive_value] in 2 files. I think later one is more clear.

I do not reject this if all other reviewers agree on. But I'm still keep thinking it.

gyliu513:
/lgtm

Leave to @jichenjc and @xunpan fo approve

jichenjc:
as @xunpan has some concern, let's wait for his comment first

chenzhiwei:
@xunpan Usually there will be a comment in each environment definition, like following:

```
# This is your IBM Cloud username
USERNAME=myname
```

But for this case, the environment variables are easy to understand and most users only need to update several of them.

For long term on cluster-api, not only ibmcloud provider, it is better to have a single configuration file:

```
apiVersion: ibmcloud.cluster.k8s.io/v1alpha1
kind: IBMCloudConfiguration

auth:
  apiUsername: your-name
  authenticationKey: your-key

cluster:
  name: test1
  domain: cluster.local
```

When I get familiar with cluster-api, I will propose this to community.

chenzhiwei:
AZure has similar with my proposal: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/getting-started.md#customizing-the-cluster-deployment

chenzhiwei:
will re-open if needs.

PR #228, opened 2019-06-06
add command line options for ibm cloud operation

xunpan:
This is not the perfect solution but a better way in my thought.

The `options` package with `init()` function to register flags to controller to avoid deeply passing arguments in a framework. This is not what I like but suitable in this case.

Make `options` under `ibmcloud` because most of configurable items should for ibmcloud.

Fixes #

/sig ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -28,6 +28,7 @@ import (
 	clusterapis "sigs.k8s.io/cluster-api/pkg/apis"
 
 	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/apis"
+	_ "sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/cloud/ibmcloud/options"
 	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/controller"
 	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/record"
 )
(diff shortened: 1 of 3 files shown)


xunpan:
Help message with `-h`
```
  -token-ttl duration
        TTL for kubeadm bootstrap tocken of the target Kubernetes cluster (default 1h0m0s)
```

any comments?

wanghh2000:
@xunpan  I saw klog.InitFlags(nil) at the first line of `main`, do we need move it into options?

clyang82:
@xunpan there is a typo error - `tocken` to `token`

jichenjc:
I didn't try this by hand. but looks to me the option is added for CLI to be used and the code createBootstrapToken is used by Create and in turn called by Reconcile function , this reconcile function should be used in the machine controller ? how the option send to the machine controller I am very curious :), hope to get some guide here :)

xunpan:
@wanghh2000 : klog should be fine. It is global to all messages for the controller. So, it is good to put in main function. Do you help to show any example command line or log for your error throwing case?

@clyang82 : thanks for your comments. Updated.

@jichenjc : yes. Command line options will pass to actuator for use.
`flag.DurationVar` just bind the variable, `TokenTTL`,  into global variable `flag.CommandLine`. As it is in `init()`, it is executed before `main` function of the binary. The command line options will be parsead in main function of controller with `flag.Parse()`. After parsing, the binded variable in `options.TokenTTL` will be set if specified.

jichenjc:
great to know this :) thanks a lot for the detailed explanation @xunpan

wanghh2000:
Pls go ahead, I will make my code change base on your fix, and retest. Thanks.

PR #233, opened 2019-06-11
Upgrade gopkg deps

clyang82:
Signed-off-by: clyang82 <clyang@cn.ibm.com>

In this PR, update the deps to the new version and remove the `master` branch. use the specific revision instead `master` to avoid any exceptions when we do next release.

/sig ibmcloud


gyliu513:
@clyang82 I think you also need another commit here as well for the vendor.

Gopkg.toml
```diff
@@ -41,20 +41,20 @@ required = [
...
-  name="sigs.k8s.io/controller-tools"
-  version="v0.1.1"
+  name = "sigs.k8s.io/controller-tools"
+  version = "v0.1.9"
 
 [[constraint]]
-  name="github.com/softlayer/softlayer-go"
-  branch="master"
+  name = "github.com/softlayer/softlayer-go"
+  revision = "7c592eb"
```
gyliu513: +1 to use revision here, this can make sure us to a stable release or continue use this revision when release the ibmcloud provider.

@xunpan @jichenjc for comments.

xunpan: I think it is fine for softlayer-go because it never has an official release. A snapshot version is better to avoid updating vendors frequently without expectation.

But for `cluster-api`, I prefer to use a release version unless there has strong reason to use a specific snapshot.

xunpan:
@clyang82 please make sure IBM cloud provider works after upgrade related package. This changes a lot of code and CI will not test clusterctl create functions.

clyang82:
> @clyang82 please make sure IBM cloud provider works after upgrade related package. This changes a lot of code and CI will not test clusterctl create functions.

I have tried manually. It works. Thanks.

gyliu513:
/approve

leave to @xunpan and @jichenjc for lgtm

Gopkg.toml
jichenjc: agree, we can bump the version if need

for cluster-api, we are using 0.1-release branch now, so we should be fine

jichenjc:
let's hold until we are informed that real test succeed

clyang82:
> let's hold until we are informed that real test succeed

What do you mean about real test?

jichenjc:
oops, I missed the comments you added (the manual test passed), sorry
/lgtm

Gopkg.toml
```diff
@@ -41,20 +41,20 @@ required = [
...
 
 [[constraint]]
   name = "k8s.io/klog"
-  version = "0.3.0"
+  version = "0.3.2"
 
 [[constraint]]
   name = "sigs.k8s.io/cluster-api"
-  branch = "release-0.1"
+  revision = "280e527"
```
xunpan: what is the reason to change it? As my comments above, we'd better use stable release for such kind of project.
@jichenjc @clyang82

clyang82: In order to leverage new feature of `cluster-api`, for example: generate random cluster name instead of hardcode `clusterapi`. that will enable to run the job in parallel.

gyliu513: @xunpan https://github.com/kubernetes-sigs/cluster-api/pull/984

jichenjc: @xunpan that's my bad, I thought this is for softlayer ... I also think we need stick to release branch 
per above comments, if we need some feature, we can backport from master to release branch

@clyang82 can you help to revert this line back ? THanks

gyliu513: @xunpan @jichenjc Consider master is a development branch, we can update the `cluster-api` to a new stable version after the release including kubernetes-sigs/cluster-api#984 was cut. For now, how about keep the current code as it is? @clyang82 can follow up another PR to upgrade the cluster-api later.

xunpan: @clyang82 I'm not sure.
The new code always fix many bugs. No doubt about it. The thing is that we should balance the code maintenance and bug fix for 3rd parties. I think only block issue can break the rule for maintenance. Or else, waiting for new stable release should be better.

jichenjc: +1 to @xunpan , I don't think switch to master develop branch is a very good idea for us at current stage

clyang82: agree with your idea. Here is new PR #239 to revert the change for clusterapi branch. Thanks.

PR #274, opened 2019-08-08
Changes related to sig ibm cloud rename

spzala:
Changes related to sig ibm cloud became provider ibm cloud. 

Fixes # 269

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -7,7 +7,7 @@ approvers:
   - cluster-api-ibmcloud-maintainers
 
 reviewers:
-  - sig-ibmcloud-leads
+  - provider-ibmcloud-leads
   - cluster-api-maintainers
   - cluster-api-ibmcloud-maintainers
   - cluster-api-ibmcloud-reviewers
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -23,13 +23,9 @@ aliases:
     - roberthbailey
     - rsdcastro
     - spew
-  sig-ibmcloud-leads:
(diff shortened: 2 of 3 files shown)


jichenjc:
seems the owner files are invalid...

gyliu513:
@spzala do you want to update this file https://github.com/kubernetes/community/blob/master/sig-cloud-provider/OWNERS ?

spzala:
> @spzala do you want to update this file https://github.com/kubernetes/community/blob/master/sig-cloud-provider/OWNERS ?

@gyliu513 I guess we shouldn't update it. The owners there are sig leads - so 3 current leads. It doesn't include the sub-project leads. Also, I believe that should not have anything to do with the plug-in invalid-owners-file error we are seeing here. The error is because Ajay and Khalid and possibly Richard are not org member. 

The sub-project owners are listed here, https://github.com/kubernetes/community/blob/master/OWNERS_ALIASES#L182
And you can see a commit below related to removing non-org member, 
https://github.com/kubernetes/community/pull/3941/commits/fd0631e420901228f1ff8a5ee70a595f55abe4bd  (Neither are kubernetes org members) 

I guess what we can do is - either remove non-org member owners from the owners file and add them later once they become member or work with them and see if they can apply for membership now and wait for them to become member. I will talk to Khalid and Richard. Does that sounds good? Thanks!

gyliu513:
@spzala Yes, we can remove non kubernetes org member first ;-)

PR #275, opened 2019-08-13
UBUNTU_LATEST does not work

xunpan:
1. The old code hardcoded docker version. But it does not exist now. Just removed and tested fine.
2. Make file change. There is no need to check and generate before build a binary. It wastes time for waiting this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,13 +149,16 @@ generate_yaml_test: kubectl
 ############################################################
 build: manager clusterctl
 
-manager: depend generate check
+manager: depend generate check mgr
+
+clusterctl: depend generate check cmd
+
+mgr:
 	CGO_ENABLED=0 GOOS=$(GOOS) go build \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -149,13 +149,13 @@ generate_yaml_test: kubectl
 ############################################################
 build: manager clusterctl
 
-manager: depend generate check
```
gyliu513: why did you remove `depend generate check`?

xunpan: The related steps lasts for long time in building a binary. I think it is not necessary for testing a simple change in code for the binary.

gyliu513: We cannot delete this, as the ` depend generate check` can make sure that we can always have correct format, and updated vendor files.

xunpan: update code to make old target keep as is.

add new short target for quick build to accelerate working efficiency.

PR #280, opened 2019-11-10
Normalize variable and function naming

sivanzcw:
* Fix golint error and normalize variable or function naming


gyliu513:
Thanks @sivanzcw !

/ok-to-test
/lgtm
/approve

PR #316, opened 2021-04-26
Add provider ID

yuchizh:

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/templates/cluster-template.yaml
+++ b/templates/cluster-template.yaml
@@ -64,12 +64,18 @@ spec:
     initConfiguration:
       nodeRegistration:
         criSocket: /var/run/containerd/containerd.sock
-        kubeletExtraArgs: {eviction-hard: 'nodefs.available<0%,nodefs.inodesFree<0%,imagefs.available<0%'}
+        kubeletExtraArgs:
+          cloud-provider: external
+          provider-id: ibmvpc://${CLUSTER_NAME}/'{{ v1.local_hostname }}'
+          eviction-hard: nodefs.available<0%,nodefs.inodesFree<0%,imagefs.available<0%
     joinConfiguration:
       discovery: {}
       nodeRegistration:
         criSocket: /var/run/containerd/containerd.sock
-        kubeletExtraArgs: {eviction-hard: 'nodefs.available<0%,nodefs.inodesFree<0%,imagefs.available<0%'}
+        kubeletExtraArgs:
+          cloud-provider: external
+          provider-id: ibmvpc://${CLUSTER_NAME}/'{{ v1.local_hostname }}'
+          eviction-hard: nodefs.available<0%,nodefs.inodesFree<0%,imagefs.available<0%
 ---
 kind: IBMVPCMachineTemplate
 apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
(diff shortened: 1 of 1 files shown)


templates/cluster-template.yaml
```diff
@@ -64,12 +64,18 @@ spec:
     initConfiguration:
       nodeRegistration:
         criSocket: /var/run/containerd/containerd.sock
-        kubeletExtraArgs: {eviction-hard: 'nodefs.available<0%,nodefs.inodesFree<0%,imagefs.available<0%'}
+        kubeletExtraArgs:
+          cloud-provider: external
+          provider-id: ibmvpc://${CLUSTER_NAME}/'{{ v1.local_hostname }}'
```
jichenjc, line 69: I assume the hostname is unique? I know other cloud provider must use uuid ..

templates/cluster-template.yaml
```diff
@@ -130,4 +136,6 @@ spec:
       joinConfiguration:
         nodeRegistration:
           kubeletExtraArgs:
+            cloud-provider: external
```
diegocn, line 139: By setting cloud-provider to external I got my pods stuck unschedulable.

```
Events:
  Type     Reason            Age                 From               Message
  ----     ------            ----                ----               -------
  Warning  FailedScheduling  21s (x13 over 14m)  default-scheduler  0/3 nodes are available: 1 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate, 2 node(s) had taint {node.cloudprovider.kubernetes.io/uninitialized: true}, that the pod didn't tolerate.
```

templates/cluster-template.yaml
```diff
@@ -130,4 +136,6 @@ spec:
...
+            provider-id: ibmvpc://${CLUSTER_NAME}/'{{ v1.local_hostname }}'
```
diegocn, line 140: By setting this `provider-id` I saw that my machines and the machinedeployment itself reached the `Running` status.
Prior to that, they got stuck on the `Unavailable` state, even though they were Ready and receiving pods.

templates/cluster-template.yaml
jichenjc: yes, this value should not be set unless we have cloud-provider-ibmcloud ready then use that cloud provider to run controller 
https://github.com/kubernetes/cloud-provider has further info

templates/cluster-template.yaml
jichenjc: I think this id should be obtained from cloud ? e.g a VM/machine is created and a unique ID 
for it, then use that ID for the provider-id?

templates/cluster-template.yaml
diegocn: So, there is no `cloud-provider-ibmcloud` Ready for use?
I'm asking that because I've tried to create a Service of type LoadBalancer but it got stuck due to the lack of a cloud-provider set on the cluster created by the capi provider ibm.

jichenjc: 2 things 
you need LB, then you can go ahead in capi-ibmcloud to create it through API exposed by IBM cloud
but if you create a service of type LB, check how openstack did https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md, they rely on external cloud provider now , but I think legacy way before cloud provider method should have simliar way 

but the question is you set `external` here which tells k8s to rely on external cloud provider while as far as I can tell it's existing but not open sourced yet .. so here's the gap I guess

gyliu513:
Fixed in #330

PR #326, opened 2021-07-26
v1alpha4 support

mkumatag:

v1alpha4 support

Fixes #

/cc @gyliu513 @jichenjc @wentao-zh 

/area provider/ibmcloud

**Release note**:

```release-note
v1alpha4 support
```


config/rbac/role.yaml
```diff
@@ -6,29 +6,6 @@ metadata:
   creationTimestamp: null
   name: manager-role
 rules:
-- apiGroups:
```
gyliu513, line 9: @mkumatag so we do not need those rules anymore for secrets and events?

gyliu513 (commented):
Thanks @mkumatag , looks great and just one minor comment want to get your conform for RBAC.

config/rbac/role.yaml
mkumatag: na! I think this is happening part of autogenerated code, lemme fix the annotations in the controller to keep these rules intact.

mkumatag: @gyliu513 added the kubebuilder rule to the controller with the required permissions. ptal at latest commit

gyliu513:
/lgtm

Thanks @mkumatag , looks good to me.

@wentao-zh can you help get someone to do some test against this new PR?

wentao-zh:
> /lgtm
> 
> Thanks @mkumatag , looks good to me.
> 
> @wentao-zh can you help get someone to do some test against this new PR?

Yes. I'm working on this testing right now.

gyliu513:
@wentao-zh @mkumatag since cluster api has some break changes between different versions as it is still involving fast, can we merge this PR? I think we do not need to support cluster api 0.3.0 as long as the current branch works after the PR got merged, comments?

mkumatag:
> @wentao-zh @mkumatag since cluster api has some break changes between different versions as it is still involving fast, can we merge this PR? I think we do not need to support cluster api 0.3.0 as long as the current branch works after the PR got merged, comments?

agree,!

jichenjc:
+1 , unless it become better, the API will change very fast, if we really need think about some kind of alpah3 to alpha4, maybe can refer to 
https://cluster-api.sigs.k8s.io/developer/providers/v1alpha3-to-v1alpha4.html
but guess not necessary

mkumatag:
> +1 , unless it become better, the API will change very fast, if we really need think about some kind of alpah3 to alpha4, maybe can refer to
> https://cluster-api.sigs.k8s.io/developer/providers/v1alpha3-to-v1alpha4.html
> but guess not necessary

I have used this guide and adopted the recommendation wherever necessary,

PR #507, opened 2022-01-11
Add IBMPowerVSImage API

### LINKED ISSUE #474: New API/Resource to import the PowerVS Image
/kind feature
/area provider/ibmcloud

This task is for creating a new Resource/controller for importing the image into PowerVS Service Instance

Need a detailed design document for this work.

Prajyot-Parab:
Signed-off-by: Prajyot-Parab <prajyot.parab2@ibm.com>

1. Add IBMPowerVSImage API to import image in the powervs service instance
2. Add Webhook for IBMPowerVSImage API

Fixes #474 

/area provider/ibmcloud

**Release note**:

```release-note

```


controllers/ibmpowervsimage_controller.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (r *IBMPowerVSImageReconciler) reconcile(ctx context.Context, imageScope *scope.PowerVSImageScope) (ctrl.Result, error) {
+	controllerutil.AddFinalizer(imageScope.IBMPowerVSImage, v1beta1.IBMPowerVSImageFinalizer)
+
+	lastJob, err := imageScope.GetImportJob()
+	if err != nil {
+		return ctrl.Result{}, err
+	} else if lastJob != nil {
+		if *lastJob.Status.State != "completed" {
+			return ctrl.Result{}, nil
```
mkumatag, line 108: what happens when there are more than one image/job present?

mkumatag: It will be good if we can track the `jobID` from the create call(store that in the status of the image and then compare/track that here will be better approach.

controllers/ibmpowervsimage_controller.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+		imageScope.IBMPowerVSImage.Status.ImageID = *image.ImageID
+		imageScope.IBMPowerVSImage.Status.ImageState = image.State
+		if imageScope.IBMPowerVSImage.Status.ImageState == "active" {
+			imageScope.IBMPowerVSImage.Status.Ready = true
+		}
+		imageScope.Info(*img.ImageID)
+	} else {
+		imageScope.IBMPowerVSImage.Status.ImageState = "pending"
+		imageScope.IBMPowerVSImage.Status.ImageID = "pending"
```
mkumatag, line 134: prefer leaving these untouched if no info found instead of adding with some dummy data

cloud/scope/powervs_image.go
```diff
@@ -0,0 +1,189 @@
...
+	} else if imageReply != nil {
+		//TODO need a reasonable wrapped error
+		return imageReply, nil
+	}
+
+	params := &p_cloud_images.PcloudV1CloudinstancesCosimagesPostParams{
+		Body: &models.CreateCosImageImportJob{
+			ImageName:     &m.Name,
+			BucketName:    s.BucketName,
+			BucketAccess:  core.StringPtr("public"),
```
mkumatag, line 160: `public` can be created as a constant somewhere

controllers/ibmpowervsimage_controller.go
```diff
@@ -0,0 +1,172 @@
...
+	// Fetch the Image.
+	image, err := util.GetOwnerImage(ctx, r.Client, ibmImage.ObjectMeta)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if image == nil {
+		log.Info("Image Controller has not yet set OwnerRef")
+		return ctrl.Result{}, nil
+	}
+
```
mkumatag, line 74: Need to revisit this code, will do that post fixing the first set of comments

mkumatag (changes requested):
Need to review Reconcile code thoroughly, will do it post fixing these comments.

controllers/ibmpowervsimage_controller.go
Prajyot-Parab: @mkumatag I think there is a limit on import job on PowerVS service instance (it is restricted to 1 job at a time), so above logic should be good considering that, if that is not the case then we can switch to the JobID approach.

controllers/ibmpowervsimage_controller.go
Prajyot-Parab: @mkumatag I had done this as whenever the image is deleted from PowerVS instance, without actually deleting the IBMPowerVSImage resource the reconciler logic should be able to update the status of image back to pending (otherwise it will still be shown as active).

controllers/ibmpowervsimage_controller.go
mkumatag: Its always good to capture the `jobID` and compare the result with a current running job, there is a possibility that the current import has already finished and some other import started meanwhile then we will be end up waiting for the wrong job.

controllers/ibmpowervsimage_controller.go
mkumatag: In that case we may we should just flush the state instead of making that pending, these field should be one to one mapping with the state of the actual resource in the powervs infra

mkumatag: and also any data will be useless if the resource is not in `Ready` state.

cloud/scope/powervs_image.go
```diff
@@ -0,0 +1,198 @@
...
+		Body: &models.CreateCosImageImportJob{
+			ImageName:     &m.Name,
+			BucketName:    s.Bucket,
+			BucketAccess:  core.StringPtr(BucketAccess),
+			Region:        s.Region,
+			ImageFilename: s.Object,
+			StorageType:   s.StorageType,
+		},
+	}
+	jobRef, err := i.IBMPowerVSClient.CreateCosImage(params.Body)
```
mkumatag, line 175:
```suggestion
	body := &models.CreateCosImageImportJob{
			ImageName:     &m.Name,
			BucketName:    s.Bucket,
			BucketAccess:  core.StringPtr(BucketAccess),
			Region:        s.Region,
			ImageFilename: s.Object,
			StorageType:   s.StorageType,
		}
	jobRef, err := i.IBMPowerVSClient.CreateCosImage(body)
```

mkumatag (approved):
/lgtm
/approve

PR #511, opened 2022-01-13
Add sync-period flag

### LINKED ISSUE #510: Add an option to override the reconcile SyncPeriod
/kind feature
/area provider/ibmcloud

Currently, syncPeriod is hardcoded and set as 15 seconds which is too little and causes the burden on the cloud, may also hit the rate limit, hence need a mechanism to override the value and set a decent value as default.

mkumatag:

- Added `sync-period` flag and set 10min as a default value
- Overriding the value via `RequeueAfter` in the machine controller wherever required to cut the deployment time

Fixes #510 

/area provider/ibmcloud

**Release note**:

```release-note
NONE
```


wentao-zh (approved):
/lgtm

PR #574, opened 2022-02-22
Use LocalObjectReference for ImageRef

### LINKED ISSUE #573: Use LocalObjectReference for the ImageRef in the machine spec
/kind bug
/area provider/ibmcloud

Rightnow ImageRef is of type ObjectReference holds a complete information of an object of any kind and which is not really required because we use only the name from the reference and also we don't anticipate to any different kind as value here.

**Possible solution:**

Can be refactored to use v1.LocalObjectReference which has only the name as field.

mkumatag:

Fixes #573 

/area provider/ibmcloud

**Release note**:

```release-note

```


Prajyot-Parab (approved):
/lgtm

PR #580, opened 2022-03-01
Added flag to select ProviderID format for Machine

Karthik-K-N:

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note

```
This PR introduce a flag `providerid-format `to select the `ProviderID` format to set to the machine

v1 version will have format, this is default provider id format: `ibmpowervs://<cluster_name>/<vm_hostname>`
v2 version will have format: `ibmpowervs://<region>/<zone>/<service_instance_id>/<powervs_machine_id>`

When we are setting the `providerid-format ` to v2, We need to explicitly run the external cloud controller manager, which will take the responsibility of setting the ProviderID for node


api/v1beta1/ibmpowervsmachine_types.go
```diff
@@ -74,6 +74,14 @@ type IBMPowerVSMachineSpec struct {
...
 	// +optional
 	ProviderID *string `json:"providerID,omitempty"`
+
+	// Region specifies the Power VS Service instance region
+	// +optional
+	Region *string `json:"region,omitempty"`
+
+	// Zone specifies the Power VS Service instance zone
+	// +optional
+	Zone *string `json:"zone,omitempty"`
```
mkumatag, line 84: keeping them optional will be challenge because is user doesn't provide and then someone set v2 as provider version this will throw an NPE, wondering if we can make these params as required fields.

what other think? @Amulyam24 @Prajyot-Parab

main.go
```diff
@@ -210,4 +212,11 @@ func initFlags(fs *pflag.FlagSet) {
 		10*time.Minute,
 		"The minimum interval at which watched resources are reconciled.",
 	)
+
+	fs.StringVar(
+		&util.ProviderIDFormat,
+		"providerid-format",
```
mkumatag, line 218: this can be `powervs-provider-id-fmt`

main.go
```diff
@@ -210,4 +212,11 @@ func initFlags(fs *pflag.FlagSet) {
...
+		"v2",
```
mkumatag, line 219: wondering if we can set `v1` as default

util/util.go
```diff
@@ -25,6 +25,12 @@ import (
 	"sigs.k8s.io/cluster-api-provider-ibmcloud/api/v1beta1"
 )
 
+// ProviderIDFormat is used to identify the Provider ID format for Machine
+var ProviderIDFormat string
+
+// ProviderIDFormatV2 will set provider id to machine as ibmpowervs://<region>/<zone>/<service_instance_id>/<powervs_machine_id>
+const ProviderIDFormatV2 = "v2"
```
mkumatag, line 32: wondering if we can move this to `pkg/options` package

api/v1beta1/ibmpowervsmachine_types.go
mkumatag: @Karthik-K-N if required, lets setup a meeting to discuss this in detail

mkumatag (changes requested):

main.go
Karthik-K-N: sure will rename

main.go
Karthik-K-N: It is supposed to be v1, I think i missed while testing, Will set back to v1

api/v1beta1/ibmpowervsmachine_types.go
Karthik-K-N: We can get the region and zone using the Power VS service instance id and currently its done like that we are not depending on user to provide it.

Amulyam24: I think keeping it optional would be fine as the region and zone are always being set before setting the provider ID despite the provider version used.

main.go
```diff
@@ -210,4 +211,11 @@ func initFlags(fs *pflag.FlagSet) {
 		10*time.Minute,
 		"The minimum interval at which watched resources are reconciled.",
 	)
+
+	fs.StringVar(
+		&options.PowerVSProviderIDFormat,
+		"powervs-provider-id-fmt",
+		"v1",
+		"ProviderID format is used set the Provider ID format for Machine",
+	)
```
mkumatag, line 220: this accepts any random value as of now, can we limit this to accept only v1 or v2

pkg/options/options.go
```diff
@@ -0,0 +1,23 @@
...
+limitations under the License.
+*/
+
+package options
+
+// PowerVSProviderIDFormat is used to identify the Provider ID format for Machine
+var PowerVSProviderIDFormat string
+
+// PowerVSProviderIDFormatV2 will set provider id to machine as ibmpowervs://<region>/<zone>/<service_instance_id>/<powervs_machine_id>
+const PowerVSProviderIDFormatV2 = "v2"
```
mkumatag, line 23: can we add v1 constant as well?

mkumatag (changes requested):

api/v1beta1/ibmpowervsmachine_types.go
Prajyot-Parab: IMO, +1 to move it as part of status, as we have internal logic to deduce their values.

Karthik-K-N:
Review comments are being addressed in the latest commit and PR is ready for new review.

main.go
```diff
@@ -68,6 +70,11 @@ func main() {
 
 	ctrl.SetLogger(klogr.New())
 
+	if err := validateFlags(); err != nil {
+		setupLog.Info("Flag validation failure", "error", err.Error())
```
mkumatag, line 74: how about using Error() if they have one

main.go
```diff
@@ -210,4 +217,24 @@ func initFlags(fs *pflag.FlagSet) {
 		10*time.Minute,
 		"The minimum interval at which watched resources are reconciled.",
 	)
+
+	fs.StringVar(
+		&options.PowerVSProviderIDFormat,
+		"powervs-provider-id-fmt",
+		"v2",
```
mkumatag, line 224: can we use options.PowerVSProviderIDFormatV2 here instead of v2?

cloud/scope/powervs_machine.go
```diff
@@ -20,13 +20,12 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/options"
```
mkumatag, line 23: nit: move this import to down

mkumatag (changes requested):
minor fmt issues, other lgtm

mkumatag (approved):
/lgtm
/approve
/retest
/hold for @Prajyot-Parab 's review

Prajyot-Parab:
@Karthik-K-N do run `make generate-go-conversions-core`  target command, let me know if it generates new changes to be pushed else PR changes lgtm!

Prajyot-Parab (approved):
/lgtm

PR #808, opened 2022-08-16
Fix E2E job failure

Amulyam24:
The API Key used by an e2e test is given restricted access based on the necessary operations. Using resource-group while logging into IBM Cloud account can be ignored as it would require giving additional resource-group permissions which might not be needed.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note

```


mkumatag (approved):
/lgtm

PR #895, opened 2022-09-26
Update logs as per capi standards

Karthik-K-N:

This PR is an attempt to update usage of log and standardize log messages as per capi standards for all Power VS related controllers: https://cluster-api.sigs.k8s.io/developer/logging.html

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/controllers/ibmpowervsimage_controller.go
+++ b/controllers/ibmpowervsimage_controller.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 
 	"github.com/IBM-Cloud/power-go-client/power/models"
@@ -30,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/klog/v2"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -49,7 +49,6 @@ import (
 // IBMPowerVSImageReconciler reconciles a IBMPowerVSImage object.
 type IBMPowerVSImageReconciler struct {
 	client.Client
-	Log             logr.Logger
 	Recorder        record.EventRecorder
 	ServiceEndpoint []endpoints.ServiceEndpoint
 	Scheme          *runtime.Scheme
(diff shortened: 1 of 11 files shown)


Prajyot-Parab (approved):
/lgtm

controllers/ibmpowervsimage_controller.go
```diff
@@ -75,6 +74,7 @@ func (r *IBMPowerVSImageReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+	log = log.WithValues("cluster", cluster.Name)
```
mkumatag, line 77: again this should be image scoped right?

controllers/ibmpowervsmachine_controller.go
```diff
@@ -87,6 +88,7 @@ func (r *IBMPowerVSMachineReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		log.Info("Machine Controller has not yet set OwnerRef")
 		return ctrl.Result{}, nil
 	}
+	log = log.WithValues("machine", machine.Name)
```
mkumatag, line 91: needs to be ibmpowervsmachine vs machine?

controllers/ibmpowervsmachine_controller.go
```diff
@@ -170,17 +172,12 @@ func (r *IBMPowerVSMachineReconciler) reconcileDelete(scope *scope.PowerVSMachin
...
-		scope.Info("InstanceID is not yet set, hence not invoking the powervs API to delete the instance")
+		scope.Info("InstanceID is not yet set, hence not invoking the PowerVS API to delete the instance")
 		return ctrl.Result{}, nil
 	}
 	if err := scope.DeleteMachine(); err != nil {
 		scope.Info("error deleting IBMPowerVSMachine")
-		return ctrl.Result{}, errors.Wrapf(err, "error deleting IBMPowerVSMachine %s/%s", scope.IBMPowerVSMachine.Namespace, scope.IBMPowerVSMachine.Name)
-	}
-	// Remove the cached VM IP
-	err := scope.DHCPIPCacheStore.Delete(powervs.VMip{Name: scope.IBMPowerVSMachine.Name})
```
mkumatag, line 181: why we have removed this block? can this be part of separate PR if needed?

mkumatag (changes requested):

controllers/ibmpowervsmachine_controller.go
Karthik-K-N: will change to ibmPowerVSMachine to avoid confustion

controllers/ibmpowervsmachine_controller.go
Karthik-K-N: will add it back in this PR but we can think of removing it in future because once the cache time is over its automatically removed.

mkumatag (approved):
/lgtm

PR #968, opened 2022-11-21
Deprecate Image in favor of ImageRef variable (VPC)

Prajyot-Parab:
Signed-off-by: Prajyot-Parab <prajyot.parab2@ibm.com>

- Deprecate Image in favor of ImageRef variable (VPC)

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Deprecate Image in favor of ImageRef variable (VPC)
```


config/crd/bases/infrastructure.cluster.x-k8s.io_ibmvpcmachines.yaml
```diff
@@ -245,11 +245,22 @@ spec:
             description: IBMVPCMachineSpec defines the desired state of IBMVPCMachine.
             properties:
               image:
-                description: 'Image is the id of OS image which would be install on
-                  the instance. Example: r134-ed3f775f-ad7e-4e37-ae62-7199b4988b00
-                  TODO: allow user to specify a image name is much reasonable. Example:
-                  ibm-ubuntu-18-04-1-minimal-amd64-2'
+                description: 'Deprecated: Use ImageRef instead Image is the id of
```
Amulyam24, line 248: small nit: this statement is not clear, probably should be: `'Deprecated: Use ImageRef instead. Image is the id of ...`

config/crd/bases/infrastructure.cluster.x-k8s.io_ibmvpcmachinetemplates.yaml
```diff
@@ -209,11 +209,23 @@ spec:
                       of the machine.
                     properties:
                       image:
-                        description: 'Image is the id of OS image which would be install
-                          on the instance. Example: r134-ed3f775f-ad7e-4e37-ae62-7199b4988b00
-                          TODO: allow user to specify a image name is much reasonable.
-                          Example: ibm-ubuntu-18-04-1-minimal-amd64-2'
+                        description: 'Deprecated: Use ImageRef instead Image is the
```
Amulyam24, line 212: same here as well

config/crd/bases/infrastructure.cluster.x-k8s.io_ibmvpcmachines.yaml
Prajyot-Parab: Fixed.

config/crd/bases/infrastructure.cluster.x-k8s.io_ibmvpcmachinetemplates.yaml
Prajyot-Parab: Fixed.

Prajyot-Parab:
@Amulyam24 Thanks for your feedback, necessary changes done. PTAL

Amulyam24 (approved):
/lgtm

Amulyam24:
> @Amulyam24 Thanks for your feedback, necessary changes done. PTAL

Having both `Image` and `ImageRef` was a bit misleading earlier. Gives a better idea now 👍🏼

mkumatag:
> > @Amulyam24 Thanks for your feedback, necessary changes done. PTAL
> 
> Having both `Image` and `ImageRef` was a bit misleading earlier. Gives a better idea now 👍🏼

But this breaks the backward compatibility, removing a field without even deprecating isn't a great idea.

Amulyam24:
> > > @Amulyam24 Thanks for your feedback, necessary changes done. PTAL
> > 
> > 
> > Having both `Image` and `ImageRef` was a bit misleading earlier. Gives a better idea now 👍🏼
> 
> But this breaks the backward compatibility, removing a field without even deprecating isn't a great idea.

@mkumatag, the initial idea was to rename them, `Image` --> `ImageID` and `ImageRef` --> `Image`.
Any thoughts on it?

mkumatag:
> > > > @Amulyam24 Thanks for your feedback, necessary changes done. PTAL
> > > 
> > > 
> > > Having both `Image` and `ImageRef` was a bit misleading earlier. Gives a better idea now 👍🏼
> > 
> > But this breaks the backward compatibility, removing a field without even deprecating isn't a great idea.
> 
> @mkumatag, the initial idea was to rename them, `Image` --> `ImageID` and `ImageRef` --> `Image`.
> Any thoughts on it?

How do you handle a scenario where the user deployed a cluster with some older version and want to update the version of capibm to the latest version with the above change?

Amulyam24:
> > > > > @Amulyam24 Thanks for your feedback, necessary changes done. PTAL
> > > > 
> > > > 
> > > > Having both `Image` and `ImageRef` was a bit misleading earlier. Gives a better idea now 👍🏼
> > > 
> > > 
> > > But this breaks the backward compatibility, removing a field without even deprecating isn't a great idea.
> > 
> > 
> > @mkumatag, the initial idea was to rename them, `Image` --> `ImageID` and `ImageRef` --> `Image`.
> > Any thoughts on it?
> 
> How do you handle a scenario where the user deployed a cluster with some older version and want to update the version of capibm to the latest version with the above change?

Right, that would be tricky. Would be better to keep it as is then.

PR #1081, opened 2023-02-07
Add common flags for capibmadm cli

dharaneeshvrd:

Add common flags needed for cli.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/cmd/capibmadm/cmd/vpc/vpc.go
+++ b/cmd/capibmadm/cmd/vpc/vpc.go
@@ -14,20 +14,26 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package network
+// Package vpc contains the commands to operate on vpc resources.
+package vpc
 
 import (
 	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/cluster-api-provider-ibmcloud/cmd/capibmadm/options"
 )
 
-// Commands function to add PowerVS network commands.
+// Commands initialises and returns VPC command.
 func Commands() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "network",
-		Short: "Perform PowerVS network operations",
+		Use:   "vpc",
+		Short: "Commands for operations on VPC resources",
 	}
 
-	cmd.AddCommand(CreateCommand())
+	cmd.PersistentFlags().StringVar(&options.GlobalOptions.VPCRegion, "region", options.GlobalOptions.VPCRegion, "IBM cloud vpc region. (Required)")
+	cmd.PersistentFlags().StringVar(&options.GlobalOptions.ResourceGroupName, "resource-group-name", options.GlobalOptions.ResourceGroupName, "IBM cloud resource group name")
+
+	_ = cmd.MarkPersistentFlagRequired("region")
 
 	return cmd
 }
--- a/cmd/capibmadm/options/options.go
+++ b/cmd/capibmadm/options/options.go
@@ -17,7 +17,11 @@ limitations under the License.
 // Package options contains the reusable and global variables.
 package options
 
-import "github.com/spf13/cobra"
+import (
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/cluster-api-provider-ibmcloud/cmd/capibmadm/printer"
+)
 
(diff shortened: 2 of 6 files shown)


cmd/capibmadm/options/options.go
```diff
@@ -28,11 +32,33 @@ var GlobalOptions = &options{}
...
+	PowerVSZone       string
+	VPCRegion         string
+	ResourceGroupName string
+	Debug             bool
+	Output            string
+}
+
+// AddCommonFlags will add common flags to the cli.
+func AddCommonFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&GlobalOptions.Output, "output", GlobalOptions.Output, "Specify output format, Only JSON is supported for now")
```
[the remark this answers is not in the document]
mkumatag: and I'm wondering if we can define output of type utils.PrinterType to handle the validation automatically?

mkumatag (changes requested):

cmd/capibmadm/cmd/vpc/vpc.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// Commands initialises and returns VPC command.
+func Commands() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "vpc",
+		Short: "Commands for operations on VPC resources",
+	}
+
+	options.AddVPCCommonFlags(cmd)
```
mkumatag, line 33: we can directly add those region and resource group flag here directly because we need them for every child command..

cmd/capibmadm/printer/printer.go
```diff
@@ -14,7 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
mkumatag, line 14: any specific reason for moving from utils -> printer?

cmd/capibmadm/options/options.go
```diff
@@ -28,11 +32,32 @@ var GlobalOptions = &options{}
...
+	PowerVSZone       string
+	VPCRegion         string
+	ResourceGroupName string
+	Debug             bool
+	Output            printer.PType
+}
+
+// AddCommonFlags will add common flags to the cli.
+func AddCommonFlags(cmd *cobra.Command) {
+	cmd.Flags().Var(&GlobalOptions.Output, "output", "Supported printer types: table, json")
```
mkumatag, line 44: can we set `PrinterTypeTable` as a default value?

cmd/capibmadm/cmd/vpc/vpc.go
mkumatag: the reason for adding for adding a common flag for powervs because not all the sub commands share same flags

mkumatag (changes requested):

cmd/capibmadm/options/options.go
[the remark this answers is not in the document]
dharaneeshvrd: Used PrinterType to handle the validation. table, json and yaml is currently supported, thought we might not need yaml. So skipped that. PTAL!

cmd/capibmadm/printer/printer.go
dharaneeshvrd: Facing circular import issue after using printer type in options pkg. Also we are introducing validation as well in printer, so its better to keep it in separate pkg.

cmd/capibmadm/options/options.go
dharaneeshvrd: set explicitly. `cmd.Flags().Var()` not providing a option set default val

cmd/capibmadm/cmd/vpc/vpc.go
dharaneeshvrd: Moved common flags to powervs and vpc commands, PTAL

cmd/capibmadm/options/options.go
mkumatag: cmd.Flags().VarP() should give you an option to do so..

mkumatag (changes requested):
one minor comment.

cmd/capibmadm/options/options.go
mkumatag: ignore!

mkumatag (approved):
/lgtm

PR #1136, opened 2023-03-02
exclude go modules for dependbot

### LINKED ISSUE #1133: Exclude the go models for dependabot
/kind cleanup
/area provider/ibmcloud

We have seen that dependabot goes and update all the go module in the repository but we don't expect dependabot to update certain repos automatically, e.g: k8s, ginkgo modules etc.. lets exclude these set of modules to avoid generating noise.

**Environment:**

- Cluster-api version:
- Minikube/KIND version:

Prajyot-Parab:
- exclude go modules for dependbot

Fixes #1133 

/area provider/ibmcloud

**Release note**:

```release-note
exclude go modules for dependbot
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,12 +4,35 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    ignore:
+    # Ignore controller-runtime as its upgraded manually.
+    - dependency-name: "sigs.k8s.io/controller-runtime"
+    # Ignore cluster-api as its upgraded manually.
+    - dependency-name: "sigs.k8s.io/cluster-api/*"
+    # Ignore k8s and its transitives modules as they are upgraded manually
+    # together with controller-runtime.
+    - dependency-name: "k8s.io/*"
+    - dependency-name: "github.com/onsi/*"
+    labels:
+      - "ok-to-test"
 
   # Enable version updates for Go tools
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,14 +4,32 @@ updates:
...
   # Enable version updates for Go tools
   - package-ecosystem: "gomod"
     directory: "/hack/tools"
     schedule:
       interval: "weekly"
-
-  - package-ecosystem: "docker"
-    directory: "/"
-    schedule:
-      interval: "weekly"
```
mkumatag, line 17: why is this removed?

mkumatag (changes requested):

.github/dependabot.yml
Prajyot-Parab: This takes care of updating only `golang` in Dockerfile which anyways we update manually, so seems like irrelevant.

mkumatag: yes, that's the intention of this code to update the golang version in the dockerfile but seems like it is not working as expected and the reason why we are doing it manually. Lets keep this line as is for now and explore later why it is not updating the version(of course only the patched version)

Prajyot-Parab: Done.

mkumatag (approved):
/lgtm

PR #1166, opened 2023-03-21
Fix inbound rule to allow only APIServerPort

### LINKED ISSUE #964: Default VPC security group has inbound rule to allow all ipv4
/kind bug
/area provider/ibmcloud

[A clear and concise description of what the bug is.]

Currently I can see default security group of VPC has inbound rule to allow all ipv4.
Its being updated from here https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/cloud/scope/cluster.go#L159

Is there any specific reason for this?

Ideally we should have inbound rule to allow whatever port we required to access the cluster.

**Environment:**

- Cluster-api version:
- Minikube/KIND version:

Prajyot-Parab:
- Fix inbound rule to allow only APIServerPort

Fixes #964 

/area provider/ibmcloud

**Release note**:

```release-note
Fix inbound rule to allow only APIServerPort
```

### DIFF
--- a/cloud/scope/cluster.go
+++ b/cloud/scope/cluster.go
@@ -186,8 +186,10 @@ func (s *ClusterScope) updateDefaultSG(sgID string) error {
 	options.SetSecurityGroupID(sgID)
 	options.SetSecurityGroupRulePrototype(&vpcv1.SecurityGroupRulePrototype{
 		Direction: core.StringPtr("inbound"),
-		Protocol:  core.StringPtr("all"),
+		Protocol:  core.StringPtr("tcp"),
 		IPVersion: core.StringPtr("ipv4"),
+		PortMin:   core.Int64Ptr(int64(s.APIServerPort())),
+		PortMax:   core.Int64Ptr(int64(s.APIServerPort())),
 	})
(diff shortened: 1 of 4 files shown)


cloud/scope/cluster.go
```diff
@@ -186,8 +186,10 @@ func (s *ClusterScope) updateDefaultSG(sgID string) error {
 	options.SetSecurityGroupID(sgID)
 	options.SetSecurityGroupRulePrototype(&vpcv1.SecurityGroupRulePrototype{
 		Direction: core.StringPtr("inbound"),
-		Protocol:  core.StringPtr("all"),
+		Protocol:  core.StringPtr("tcp"),
 		IPVersion: core.StringPtr("ipv4"),
+		PortMin:   core.Int64Ptr(6443),
```
mkumatag, line 191: What if the exposed api server port is other than 6443?

mkumatag (changes requested):

cloud/scope/cluster.go
Prajyot-Parab: Fixed, ptal.

mkumatag (approved):
/lgtm

PR #1307, opened 2023-07-21
fix machine reconciliation bug

Prajyot-Parab:

This PR will fix the error: "server name already exist"

As observed once the Post request to create a instance is sent, the instance obj returned is NIL and takes time to get updated due to delay in PowerVS response. This results in parallel reconciliation being triggered by changes in machine obj to resend the post request again resulting in conflict and error "server name already exist"

The fix would ensure we mark the instance state as unknown when obj returned is NIL post the create instance request is sent and check for the same conditions before sending duplicate post requests.

- fix machine reconciliation bug

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
fix machine reconciliation bug
```

### DIFF
--- a/api/v1beta2/conditions_consts.go
+++ b/api/v1beta2/conditions_consts.go
@@ -36,8 +36,11 @@ const (
 	// InstanceErroredReason instance is in a errored state.
 	InstanceErroredReason = "InstanceErrored"
 
-	// InstanceNotReadyReason used when the instance is in a pending state.
+	// InstanceNotReadyReason used when the instance is in a not ready state.
 	InstanceNotReadyReason = "InstanceNotReady"
+
+	// InstanceStateUnknownReason used when the instance is in a unknown state.
+	InstanceStateUnknownReason = "InstanceStateUnknown"
 )
 
 const (
--- a/cloud/scope/powervs_machine.go
+++ b/cloud/scope/powervs_machine.go
@@ -202,6 +202,15 @@ func (m *PowerVSMachineScope) CreateMachine() (*models.PVMInstanceReference, err
 		// TODO need a reasonable wrapped error.
 		return instanceReply, nil
 	}
+
+	// Check if create request has been already triggered.
+	// If InstanceReadyCondition is Unknown then return and wait for it to get updated.
+	for _, con := range m.IBMPowerVSMachine.Status.Conditions {
+		if con.Type == infrav1beta2.InstanceReadyCondition && con.Status == corev1.ConditionUnknown {
+			return nil, nil
+		}
+	}
+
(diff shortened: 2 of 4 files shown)


api/v1beta2/conditions_consts.go
```diff
@@ -36,8 +36,11 @@ const (
 	// InstanceErroredReason instance is in a errored state.
 	InstanceErroredReason = "InstanceErrored"
 
-	// InstanceNotReadyReason used when the instance is in a pending state.
+	// InstanceNotReadyReason used when the instance is in a not ready state.
 	InstanceNotReadyReason = "InstanceNotReady"
+
+	// InstanceUnknownReason used when the instance is in a unknown state.
+	InstanceUnknownReason = "InstanceUnknown"
```
Karthik-K-N, line 43: May be InstanceStateUnknown

cloud/scope/powervs_machine.go
```diff
@@ -201,6 +201,12 @@ func (m *PowerVSMachineScope) CreateMachine() (*models.PVMInstanceReference, err
 	} else if instanceReply != nil {
 		// TODO need a reasonable wrapped error.
 		return instanceReply, nil
+	} else if instanceReply == nil {
```
Karthik-K-N, line 204: Lets add a comment why we are adding this condition and when this condition will occur.

api/v1beta2/conditions_consts.go
Prajyot-Parab: Done.

mkumatag (changes requested):
can we add some UT test coverage for this change?

cloud/scope/powervs_machine.go
```diff
@@ -202,6 +202,17 @@ func (m *PowerVSMachineScope) CreateMachine() (*models.PVMInstanceReference, err
 		// TODO need a reasonable wrapped error.
 		return instanceReply, nil
 	}
+
+	if instanceReply == nil {
```
mkumatag, line 206: this if is not required, if we have reached this step only when instanceReply is nil

cloud/scope/powervs_machine.go
```diff
@@ -201,6 +201,14 @@ func (m *PowerVSMachineScope) CreateMachine() (*models.PVMInstanceReference, err
 	} else if instanceReply != nil {
 		// TODO need a reasonable wrapped error.
 		return instanceReply, nil
+	} else if instanceReply == nil {
```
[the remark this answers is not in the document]
Prajyot-Parab: Done.

cloud/scope/powervs_machine.go
Prajyot-Parab: Yes fixed.

PR #1311, opened 2023-07-25
Increase test/e2e timeouts of ginkgo and powervs machines

Prajyot-Parab:
- Increase test/e2e timeouts of ginkgo and powervs machines

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Increase test/e2e timeouts of ginkgo and powervs machines
```

### DIFF
--- a/test/e2e/config/ibmcloud-e2e-powervs.yaml
+++ b/test/e2e/config/ibmcloud-e2e-powervs.yaml
@@ -60,8 +60,8 @@ variables:
 intervals:
   default/wait-controllers: ["3m", "10s"]
   default/wait-cluster: ["20m", "10s"]
-  default/wait-control-plane: ["40m", "10s"]
-  default/wait-worker-nodes: ["40m", "10s"]
+  default/wait-control-plane: ["60m", "10s"]
+  default/wait-worker-nodes: ["60m", "10s"]
   default/wait-delete-cluster: ["20m", "10s"]
   default/wait-machine-upgrade: ["50m", "10s"]
   default/wait-machine-remediation: ["30m", "10s"]
--- a/hack/boskos.sh
+++ b/hack/boskos.sh
@@ -56,7 +56,7 @@ checkout_account(){
 heartbeat_account(){
     count=0
     url="http://${BOSKOS_HOST}/update?name=${BOSKOS_RESOURCE_NAME}&state=busy&owner=${USER}"
-    while [ ${count} -lt 120 ]
+    while [ ${count} -lt 180 ]
     do
         status_code=$(curl -s -o /dev/null -w '%{http_code}' -X POST ${url})
         if [[ ${status_code} != 200 ]]; then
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -219,7 +219,7 @@ test: generate fmt vet setup-envtest $(GOTESTSUM) ## Run tests
 GINKGO_FOCUS ?= Workload cluster creation
 GINKGO_NODES ?= 3
 GINKGO_NOCOLOR ?= false
-GINKGO_TIMEOUT ?= 2h
```
mkumatag, line 222: This you can override while calling make itself

test/e2e/config/ibmcloud-e2e-powervs.yaml
```diff
@@ -60,8 +60,8 @@ variables:
 intervals:
   default/wait-controllers: ["3m", "10s"]
   default/wait-cluster: ["20m", "10s"]
-  default/wait-control-plane: ["40m", "10s"]
-  default/wait-worker-nodes: ["40m", "10s"]
+  default/wait-control-plane: ["60m", "10s"]
```
mkumatag, line 63: if it taking beyond 40m then less likely it will become a node, issue might be something else, can you please test it running locally or prow job with forked repo

mkumatag (changes requested):

test/e2e/config/ibmcloud-e2e-powervs.yaml
Prajyot-Parab: The machines are getting stuck in build state, that is what I observed in the logs (no other noticeable errors). Yes they may or may not complete. I will try it on local setup and see if I can reproduce the same behaviour.

Amulyam24:
If we are increasing the timeout, we also need to do it in `boskos.sh` for sending regular heartbeat while the instance is in use- https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/hack/boskos.sh#L59

Prajyot-Parab:
> If we are increasing the timeout, we also need to do it in `boskos.sh` for sending regular heartbeat while the instance is in use- https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/hack/boskos.sh#L59

@Amulyam24 can you suggest the required change here, I will add it to this PR.

Amulyam24:
> > If we are increasing the timeout, we also need to do it in `boskos.sh` for sending regular heartbeat while the instance is in use- https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/hack/boskos.sh#L59
> 
> @Amulyam24 can you suggest the required change here, I will add it to this PR.

@Prajyot-Parab, you can modify the count value in the condition from 120 to 180, keeping in sync with the 3H timeout.

Prajyot-Parab:
Closing this one, as increase in timeout does not gives us reliable job performance and the e2e is still flaky (observed via multiple manual triggered jobs)

PR #1397, opened 2023-09-07
fix panic error when scope obj is not yet initialized

### LINKED ISSUE #1396: IBMPowerVSCluster runtime error while creating a cluster: invalid memory address or nil pointer dereference.
/kind bug
/area provider/ibmcloud

- Create a PowerVS cluster with incorrect `IBMPOWERVS_SERVICE_INSTANCE_ID`. Cluster will be in provisioning state + the CAPIBM controller will repeatedly crash with panic runtime error.

Error:
```
[manager] panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[manager] 	panic: runtime error: invalid memory address or nil pointer dereference
```

- Error handling should be done properly with no panic + runtime error causing codebase crash

**Environment:**

- Cluster-api version:
- Minikube/KIND version:

Prajyot-Parab:
- fix panic error when scope obj is not yet initialized

Fixes #1396 

/area provider/ibmcloud

**Release note**:

```release-note
fix panic error when scope obj is not yet initialized
```

### DIFF
--- a/controllers/ibmpowervscluster_controller.go
+++ b/controllers/ibmpowervscluster_controller.go
@@ -88,10 +88,12 @@ func (r *IBMPowerVSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		ServiceEndpoint:   r.ServiceEndpoint,
 	})
 
-	// Always close the scope when exiting this function so we can persist any GCPMachine changes.
+	// Always close the scope when exiting this function so we can persist any IBMPowerVSCluster changes.
 	defer func() {
-		if err := clusterScope.Close(); err != nil && reterr == nil {
-			reterr = err
+		if clusterScope != nil {
+			if err := clusterScope.Close(); err != nil && reterr == nil {
+				reterr = err
+			}
 		}
 	}()
 
--- a/controllers/ibmpowervsimage_controller.go
+++ b/controllers/ibmpowervsimage_controller.go
@@ -79,10 +79,13 @@ func (r *IBMPowerVSImageReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	if err != nil {
 		return ctrl.Result{}, errors.Errorf("failed to create scope: %+v", err)
 	}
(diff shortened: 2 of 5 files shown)


cloud/scope/powervs_cluster.go
```diff
@@ -150,6 +150,9 @@ func NewPowerVSClusterScope(params PowerVSClusterScopeParams) (scope *PowerVSClu
 
 // PatchObject persists the cluster configuration and status.
 func (s *PowerVSClusterScope) PatchObject() error {
+	if s == nil {
```
mkumatag, line 153: are you sure it is for the `s` ? may be it is for the patchHelper field!

mkumatag (changes requested):

cloud/scope/powervs_cluster.go
Prajyot-Parab: @mkumatag yes `s` obj is nil
```
code:
fmt.Println("obj:--- ", s)

log:
[manager] obj:---  <nil>
```

mkumatag: but from this is getting triggered, we need to block in the calling method.

mkumatag: I think this is being called in the defer method, lets handle it there, e.g: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/983675a6acf3ff5f3d3bf7dbf94d40a86d268939/controllers/ibmpowervsmachine_controller.go#L142

Prajyot-Parab: Yes fixed it.

mkumatag: but we need to handle such situation in every defer calls

Prajyot-Parab: @mkumatag you meant as a preventative measure right (since we have not came across that issue in other defer calls but yes I agree there is always a possibility)

Prajyot-Parab: Updated.

mkumatag (approved):
/lgtm

Issue #1450, opened 2023-10-19
Explore adding PowerVS K8s cluster pre-req infra creation to capibm flow


Prajyot-Parab:
/kind feature
/area provider/ibmcloud

**Describe the solution you'd like**
This task focuses on including the pre-req infra processes (PVS Workspace, Network, Port, etc) involved in creation of PowerVS K8s cluster.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

Karthik-K-N:
GOAL: Access Power VS cluster via loadbalancer


After some initial discussion the following things need to be done
1. Create Power VS workspace
2. Create VPC resources
3. Create Transit Gateway and attach power vs and vpc resources
4. Create Loadbalancer service.


Anything else is required @dharaneeshvrd @Prajyot-Parab @Amulyam24

dharaneeshvrd:
Need to keep below things in mind:
Currently transit gateway is available only in dal10, if we are implementing right now it's limited to use dal10 alone.
Need to attach private network only to tg or cc if you already not aware. DHCP service can be used for this.

Karthik-K-N:
> Need to keep below things in mind: Currently transit gateway is available only in dal10, if we are implementing right now it's limited to use dal10 alone. Need to attach private network only to tg or cc if you already not aware. DHCP service can be used for this.

Got it, Initial plan is to use only PER not to use CC.

Karthik-K-N:
@mkumatag would like to know your thoughts regarding using DHCP service, Is there any constraint on building capi image with that?

PR #1485, opened 2023-11-09
API changes to support infra creation

Karthik-K-N:

Contains the API changes to infra creation as a part of Power VS cluster creation

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
New fields are added to IBMPowerVSCluster resources to support infrastructure creation
```

- [ ] Squash commits before merge

Partially fixes: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/1450


api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -45,13 +43,53 @@ type IBMPowerVSClusterSpec struct {
...
+	// ResourceGroup name under which the resources will be created.
+	ResourceGroup string `json:"resourceGroup"`
+
+	// VPC contains information about IBM Cloud VPC resources
+	// +optional
+	VPC VPCResource `json:"vpc"`
+
+	// ControlPlaneLoadBalancer is optional configuration for customizing control plane behavior.
+	// +optional
+	ControlPlaneLoadBalancer *VPCLoadBalancerSpec `json:"controlPlaneLoadBalancer,omitempty"`
```
mkumatag, line 69: IMO everything should be optional and can have some default values if not supplied to have backward compatibility

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,11 +32,9 @@ const (
 
 // IBMPowerVSClusterSpec defines the desired state of IBMPowerVSCluster.
 type IBMPowerVSClusterSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-
 	// ServiceInstanceID is the id of the power cloud instance where the vsi instance will get deployed.
 	// +kubebuilder:validation:MinLength=1
+	// Deprecated: use ServiceInstance.ID instead
```
mkumatag, line 37: why just ID? there are other fields like name, regex

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -45,13 +43,53 @@ type IBMPowerVSClusterSpec struct {
...
 
 // IBMPowerVSClusterStatus defines the observed state of IBMPowerVSCluster.
 type IBMPowerVSClusterStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// Ready is true when the provider resource is ready.
 	Ready bool `json:"ready"`
+
+	// serviceInstance is the reference to the Power VS service on which the server instance(VM) will be created.
+	ServiceInstance IBMPowerVSResourceReference `json:"serviceInstance,omitempty"`
```
mkumatag, line 78: in status we can have just the id

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -45,13 +43,53 @@ type IBMPowerVSClusterSpec struct {
...
+
+	// Network is the reference to the Network to use for this cluster.
+	Network IBMPowerVSResourceReference `json:"network,omitempty"`
```
mkumatag, line 81: same here, lets have only id here

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -45,13 +43,53 @@ type IBMPowerVSClusterSpec struct {
...
+
+	// VPC holds the status of VPC resources
+	VPC VPCResource `json:"vpc,omitempty"`
```
mkumatag, line 84: again lets use only id in the status

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -81,6 +119,27 @@ type IBMPowerVSClusterList struct {
...
+	Name string `json:"name"`
+
+	// IBM Cloud VPC zone where the cluster's loadbalancers will be created
+	Zone string `json:"vpcZone"`
+
+	// Subnet contains the details about subnet
+	Subnet Subnet `json:"subnet"`
+
+	// TransitGateway holds the resource details of TransitGateway
+	TransitGateway TransitGateway `json:"transitGateway"`
```
mkumatag, line 135: I believe TransitGateway itself is a different component and works like connector between the vpc and classic n/w, so wondering if we need to define outside the vpc data structures

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -65,6 +65,18 @@ type VPCLoadBalancerSpec struct {
 	Name string `json:"name,omitempty"`
 }
 
+// VPCLoadBalancerStatus defines the status VPC load balancer.
+type VPCLoadBalancerStatus struct {
+	// Name of the VPC load balancer.
+	// +optional
+	Name string `json:"name,omitempty"`
```
mkumatag, line 72: in the status we shouldn't be mentioning the name

mkumatag (changes requested):

api/v1beta2/ibmpowervscluster_types.go
Karthik-K-N: agree, I think we can make use of webhooks to handle it.

api/v1beta2/ibmpowervscluster_types.go
Karthik-K-N: My intention was to refer ServiceInstanceID to ServiceInstance.ID, But yes we can use anything out of Name, ID, Regex, Update it accrodingly

Karthik-K-N:
Addressed first set of review comments.

api/v1beta2/ibmpowervscluster_types.go
mkumatag: we need a discussion about this, like what fields can be default set and what values.

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -45,13 +43,61 @@ type IBMPowerVSClusterSpec struct {
 	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
 	// +optional
 	ControlPlaneEndpoint capiv1beta1.APIEndpoint `json:"controlPlaneEndpoint"`
+
```
dharaneeshvrd, line 46: Now we need to use DHCP's private network here. Please see if you want to document that in the comment.

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -45,13 +43,61 @@ type IBMPowerVSClusterSpec struct {
...
+	// zone is the name of Power VS zone where the cluster will be created
+	Zone string `json:"zone"`
+
+	// resourceGroup name under which the resources will be created.
+	ResourceGroup string `json:"resourceGroup"`
+
+	// vpc contains information about IBM Cloud VPC resources
+	// +optional
+	VPC VPCResource `json:"vpc,omitempty"`
```
dharaneeshvrd, line 65: Just a nit to keep same pattern, please do change if it is ok
```suggestion
	VPC IBMVPCResourceReference `json:"vpc,omitempty"`
```

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -45,13 +43,61 @@ type IBMPowerVSClusterSpec struct {
...
+
+	// transitGateway contains information about IBM Cloud TransitGateway.
+	// +optional
+	TransitGateway TransitGateway `json:"transitGateway,omitempty"`
```
dharaneeshvrd, line 69: Same as VPCResource comment
```suggestion
	TransitGateway IBMTransitGatewayResource `json:"transitGateway,omitempty"`
```

api/v1beta2/ibmpowervscluster_types.go
dharaneeshvrd: Another design suggestion:
When there is a need for creating dhcp server you might need to check for existing dhcp server and compare the CIDR and chose a different one.
Rather you can use the existing one instead of creating new one which will keep things simpler.

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -45,13 +43,61 @@ type IBMPowerVSClusterSpec struct {
...
 	Ready bool `json:"ready"`
+
+	// serviceInstanceID is the reference to the Power VS service on which the server instance(VM) will be created.
+	ServiceInstanceID *string `json:"serviceInstanceID,omitempty"`
+
+	// networkID is the reference to the Power VS network to use for this cluster.
+	NetworkID *string `json:"networkID,omitempty"`
+
+	// vpcID is reference to IBM Cloud VPC resources.
+	VpcID *string `json:"vpcID,omitempty"`
```
dharaneeshvrd, line 88: like `VPCSubnetID`
```suggestion
	VPCID *string `json:"vpcID,omitempty"`
```

api/v1beta2/ibmpowervscluster_types.go
Karthik-K-N: Yeah, Makes sense will document it accordingly, may be the flow should be like
```
If the networkID is empty, check for existing DHCP server and use it or else create new one.
```

Karthik-K-N:
Addressed all the reivew comments

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,76 @@ const (
...
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-
 	// ServiceInstanceID is the id of the power cloud instance where the vsi instance will get deployed.
 	// +kubebuilder:validation:MinLength=1
+	// Deprecated: use ServiceInstance instead
 	ServiceInstanceID string `json:"serviceInstanceID"`
 
 	// Network is the reference to the Network to use for this cluster.
+	// Whenf the field is omitted, A DHCP service will be created in the service instance and its private network will be used.
```
dharaneeshvrd, line 41:
```suggestion
	// When the field is omitted, A DHCP service will be created in the service instance and its private network will be used.
```

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,76 @@ const (
...
 	Network IBMPowerVSResourceReference `json:"network"`
 
 	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
 	// +optional
```
dharaneeshvrd, line 45: Are we going to allow user to provide this as well?
Or we are going to take it from the vpc load balancer by a GET call?
Please see whether this needs to added in the comment

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,76 @@ const (
...
+
+	// vpc contains information about IBM Cloud VPC resources
+	// +optional
+	VPC IBMVPCResourceReference `json:"vpc,omitempty"`
+
+	// transitGateway contains information about IBM Cloud TransitGateway.
+	// +optional
+	TransitGateway IBMTransitGatewayResource `json:"transitGateway,omitempty"`
+
+	// controlPlaneLoadBalancer is optional configuration for customizing control plane behavior.
```
dharaneeshvrd, line 72: Might need to add info about that its a vpc load balancer like the comment in network

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,76 @@ const (
...
 	Ready bool `json:"ready"`
+
+	// serviceInstanceID is the reference to the Power VS service on which the server instance(VM) will be created.
+	ServiceInstanceID *string `json:"serviceInstanceID,omitempty"`
+
+	// networkID is the reference to the Power VS network to use for this cluster.
+	NetworkID *string `json:"networkID,omitempty"`
+
+	// dhcpServerID is the reference to the Power VS DHCP server.
+	DHCPServerID *string `json:"dhcpServerID,omitempty"`
```
dharaneeshvrd, line 89: Not sure how you are going to reverse eng to track the dhcp server id if the network is provided by the user.
Please see whether we need to get the dhcp server name too from the user?

api/v1beta2/ibmpowervscluster_types.go
Karthik-K-N: In case of using static ips as of today user can set  this.
Incase if user set this `controlPlaneLoadBalancer` then we will go ahead and provision loadbalancer and then update the values.

api/v1beta2/ibmpowervscluster_types.go
Karthik-K-N: We will set the DHCPServerID only in case we provision the DHCP server, that  will be done when `network` is empty.

api/v1beta2/ibmpowervscluster_types.go
dharaneeshvrd: In that case this can be in status instead of spec, what do you think?

Karthik-K-N: It cannot be in status because of two reasons
1. As per capi [contract](https://cluster-api.sigs.k8s.io/developer/providers/cluster-infrastructure#infracluster-resources) for infra providers it should be there in spec
2. Since its in spec user can set static ips and use the cluster as of today.

api/v1beta2/ibmpowervscluster_types.go
mkumatag: This package and file is all about IBM, so wondering if IBM everywhere is going to be repeated? btw: field name is already `transitGateway`

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,76 @@ const (
...
+	// https://cloud.ibm.com/docs/power-iaas?topic=power-iaas-creating-power-virtual-server
+	// when omitted system will dynamically create the service instance
+	// +optional
+	ServiceInstance IBMPowerVSResourceReference `json:"serviceInstance"`
+
+	// zone is the name of Power VS zone where the cluster will be created
+	Zone string `json:"zone"`
+
+	// resourceGroup name under which the resources will be created.
+	ResourceGroup string `json:"resourceGroup"`
```
mkumatag, line 62: lets have this as optional and use the default resource group when not supplied

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,76 @@ const (
...
+	// serviceInstance can be created via IBM Cloud catalog or CLI.
+	// supported serviceInstance identifier in PowerVSResource are Name and ID and that can be obtained from IBM Cloud UI or IBM Cloud cli.
+	// More detail about Power VS service instance.
+	// https://cloud.ibm.com/docs/power-iaas?topic=power-iaas-creating-power-virtual-server
+	// when omitted system will dynamically create the service instance
+	// +optional
+	ServiceInstance IBMPowerVSResourceReference `json:"serviceInstance"`
+
+	// zone is the name of Power VS zone where the cluster will be created
+	Zone string `json:"zone"`
```
mkumatag, line 59: we need have this as an optional to have a backward compatibility, lets also had a link which talks about all the possible values - https://cloud.ibm.com/docs/power-iaas?topic=power-iaas-creating-power-virtual-server

api/v1beta2/ibmpowervscluster_types.go
Karthik-K-N: so should we use `TransitGatewayResource`? and avoid IBM.

mkumatag: sgtm

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,77 @@ const (
...
+
+	// serviceInstance is the reference to the Power VS service on which the server instance(VM) will be created.
+	// Power VS service is a container for all Power VS instances at a specific geographic region.
+	// serviceInstance can be created via IBM Cloud catalog or CLI.
+	// supported serviceInstance identifier in PowerVSResource are Name and ID and that can be obtained from IBM Cloud UI or IBM Cloud cli.
+	// More detail about Power VS service instance.
+	// https://cloud.ibm.com/docs/power-iaas?topic=power-iaas-creating-power-virtual-server
+	// when omitted system will dynamically create the service instance
+	// +optional
+	ServiceInstance IBMPowerVSResourceReference `json:"serviceInstance"`
```
mkumatag, line 56:
```suggestion
	ServiceInstance IBMPowerVSResourceReference `json:"serviceInstance,omitempty"`
```

api/v1beta2/ibmvpcmachine_types.go
```diff
@@ -78,6 +78,9 @@ type IBMVPCResourceReference struct {
 	// +kubebuilder:validation:MinLength=1
 	// +optional
 	Name *string `json:"name,omitempty"`
+
+	// IBM Cloud VPC zone
+	Zone *string `json:"vpcZone,omitempty"`
```
mkumatag, line 83: As we have already a zone field in the `IBMVPCMachineSpec`, this entry may create confusion here

mkumatag (changes requested):

api/v1beta2/ibmpowervscluster_types.go
Karthik-K-N: What should be the default zone to use. Also I think we can validate this filed to have value when ServiceInstance info is not provided.

api/v1beta2/ibmvpcmachine_types.go
Karthik-K-N: Will remove it from here and create new resource VPCResource type with zone.

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -81,6 +159,58 @@ type IBMPowerVSClusterList struct {
...
+	ID *string `json:"id,omitempty"`
+
+	// Name of resource
+	// +kubebuilder:validation:MinLength=1
+	// +optional
+	Name *string `json:"name,omitempty"`
+
+	// IBM Cloud VPC zone
+	Zone *string `json:"zone,omitempty"`
+}
```
Karthik-K-N, line 184:
```suggestion
// IBM Cloud VPC region
	Region *string `json:"region,omitempty"`
}
```

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,104 @@ const (
...
+	// +optional
+	ResourceGroup *string `json:"resourceGroup,omitempty"`
+
+	// vpc contains information about IBM Cloud VPC resources
+	// +optional
+	VPC *VPCResourceReference `json:"vpc,omitempty"`
+
+	// vpcSubnet contains information about IBM Cloud VPC Subnet resources
+	// +optional
+	VPCSubnet *Subnet `json:"vpcSubnet,omitempty"`
```
Karthik-K-N, line 74:
```suggestion
	VPCSubnet []Subnet `json:"vpcSubnet,omitempty"`
```

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -63,6 +63,26 @@ type VPCLoadBalancerSpec struct {
 	// +kubebuilder:validation:Pattern=`^([a-z]|[a-z][-a-z0-9]*[a-z0-9])$`
 	// +optional
 	Name string `json:"name,omitempty"`
+
+	// Scheme sets the scheme of the load balancer (defaults to control-plane)
+	// +kubebuilder:default=control-plane
+	// +kubebuilder:validation:Enum=control-plane
+	// +optional
+	Scheme *LoadBalancerScheme `json:"scheme,omitempty"`
```
Karthik-K-N, line 71: change it to Public

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,104 @@ const (
...
+	// loadBalancers is optional configuration for configuring loadbalancers to control plane or data plane nodes.
+	// +optional
+	LoadBalancers []VPCLoadBalancerSpec `json:"loadBalancers,omitempty"`
+
+	// cosBucket contains options to configure a supporting IBM Cloud COS bucket for this
+	// cluster - currently used for nodes requiring Ignition
+	// (https://coreos.github.io/ignition/) for bootstrapping (requires
+	// BootstrapFormatIgnition feature flag to be enabled).
+	// +optional
+	CosBucket *CosBucket `json:"cosBucket,omitempty"`
```
Karthik-K-N, line 89:
```suggestion
	CosInstance *CosInstance `json:"cosInstance,omitempty"`
```

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -81,6 +159,58 @@ type IBMPowerVSClusterList struct {
...
+	//
+	// When enabled, the IAM instance profiles specified are not used.
+	// +optional
+	PresignedURLDuration *metav1.Duration `json:"presignedURLDuration,omitempty"`
+
+	// Name defines name of S3 Bucket to be created.
+	// +kubebuilder:validation:MinLength:=3
+	// +kubebuilder:validation:MaxLength:=63
+	// +kubebuilder:validation:Pattern=`^[a-z0-9][a-z0-9.-]{1,61}[a-z0-9]$`
+	Name string `json:"name,omitempty"`
```
Karthik-K-N, line 201:
```suggestion
	Name string `json:"name,omitempty"`
	BucketName string `json:"bucketName,omitempty"`
	BucketRegion string `json:"bucketName,omitempty"`
```

Karthik-K-N:
Use annotations to support existing way of creating cluster

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,104 @@ const (
...
 	// Network is the reference to the Network to use for this cluster.
+	// when the field is omitted, A DHCP service will be created in the service instance and its private network will be used.
 	Network IBMPowerVSResourceReference `json:"network"`
 
 	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
 	// +optional
 	ControlPlaneEndpoint capiv1beta1.APIEndpoint `json:"controlPlaneEndpoint"`
+
+	// serviceInstance is the reference to the Power VS service on which the server instance(VM) will be created.
+	// Power VS service is a container for all Power VS instances at a specific geographic region.
```
mkumatag, line 48:
```suggestion
	// Power Virtual Server workspace is a container for all Power Virtual Server instances at a specific geographic region.
```

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,104 @@ const (
...
+	ServiceInstance *IBMPowerVSResourceReference `json:"serviceInstance,omitempty"`
+
+	// zone is the name of Power VS zone where the cluster will be created
+	// possible values can be found here https://cloud.ibm.com/docs/power-iaas?topic=power-iaas-creating-power-virtual-server.
+	// when value for serviceInstance is omitted its required to set the value for zone.
+	// +optional
+	Zone *string `json:"zone,omitempty"`
+
+	// resourceGroup name under which the resources will be created.
+	// when omitted Default resource group will be used.
```
mkumatag, line 64: This is not a `Default` one by name, IBM cloud can set some other name as well for the default if I'm not wrong, lets modify the message accordingly.

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,104 @@ const (
...
+
+	// vpc contains information about IBM Cloud VPC resources
+	// +optional
+	VPC *VPCResourceReference `json:"vpc,omitempty"`
+
+	// vpcSubnets contains information about IBM Cloud VPC Subnet resources
+	// +optional
+	VPCSubnets []Subnet `json:"vpcSubnets,omitempty"`
+
+	// transitGateway contains information about IBM Cloud TransitGateway.
```
mkumatag, line 76: it will be great if you can add a link to the ibm cloud document for more information.

mkumatag: and the purpose of this entity here

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -32,26 +32,104 @@ const (
...
+	// +optional
+	TransitGateway *TransitGateway `json:"transitGateway,omitempty"`
+
+	// loadBalancers is optional configuration for configuring loadbalancers to control plane or data plane nodes.
```
mkumatag, line 80: information needs to added to to mention why is this optional and what happens user doesn't supply

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -81,6 +159,67 @@ type IBMPowerVSClusterList struct {
...
+	// +optional
+	ID *string `json:"id,omitempty"`
+
+	// Name of resource
+	// +kubebuilder:validation:MinLength=1
+	// +optional
+	Name *string `json:"name,omitempty"`
+
+	// IBM Cloud VPC zone
+	Zone *string `json:"zone,omitempty"`
```
mkumatag, line 183: why do we need this? can we just create subnets in all the zones by default.?

api/v1beta2/ibmpowervsmachine_types.go
```diff
@@ -121,6 +131,20 @@ type IBMPowerVSMachineSpec struct {
...
+	// +optional
+	Ignition *Ignition `json:"ignition,omitempty"`
+}
+
+// Ignition defines options related to the bootstrapping systems where Ignition is used.
+type Ignition struct {
+	// Version defines which version of Ignition will be used to generate bootstrap data.
+	//
+	// +optional
+	// +kubebuilder:default="2.3"
```
mkumatag, line 145: why it is set to 2.3 version?

api/v1beta2/ibmpowervsmachine_types.go
```diff
@@ -121,6 +131,20 @@ type IBMPowerVSMachineSpec struct {
 	// ProviderID is the unique identifier as specified by the cloud provider.
 	// +optional
 	ProviderID *string `json:"providerID,omitempty"`
+
+	// Ignition defined options related to the bootstrapping systems where Ignition is used.
+	// +optional
+	Ignition *Ignition `json:"ignition,omitempty"`
```
mkumatag, line 137: why a struct here? right now I see just a version, do we expect anything else in the future?

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -63,6 +63,25 @@ type VPCLoadBalancerSpec struct {
...
 	Name string `json:"name,omitempty"`
+
+	// public indicates that load balancer is public or private
+	// +kubebuilder:default=true
+	// +optional
+	Public bool `json:"public,omitempty"`
+}
+
+// LoadBalancerScheme defines the scheme of a load balancer.
+type LoadBalancerScheme string
```
mkumatag, line 74: where is this used?

mkumatag (changes requested):
Lets start using this terminology `PowerVS workspace` instead of service instance

mkumatag (approved):
/lgtm
/approve

/hold

for others to review

/cc @Prajyot-Parab

Prajyot-Parab (approved):
/unhold

PR #1592, opened 2024-01-25
Update api specs

Karthik-K-N:
Couple of fields were missed during earlier PR, Updating it here.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/api/v1beta2/ibmvpccluster_types.go
+++ b/api/v1beta2/ibmvpccluster_types.go
@@ -90,13 +90,14 @@ type VPCLoadBalancerStatus struct {
 	// id of VPC load balancer.
 	// +optional
 	ID *string `json:"id,omitempty"`
-	// +optional
-	Name *string `json:"name,omitempty"`
 	// State is the status of the load balancer.
 	State VPCLoadBalancerState `json:"state,omitempty"`
 	// hostname is the hostname of load balancer.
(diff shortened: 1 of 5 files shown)


api/v1beta2/ibmvpccluster_types.go
```diff
@@ -90,13 +90,14 @@ type VPCLoadBalancerStatus struct {
 	// id of VPC load balancer.
 	// +optional
 	ID *string `json:"id,omitempty"`
-	// +optional
-	Name *string `json:"name,omitempty"`
```
mkumatag, line 94: why are we removing the name?

mkumatag (changes requested):

api/v1beta2/ibmvpccluster_types.go
Karthik-K-N: Initially I thought of having [Loadbalaners](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/api/v1beta2/ibmpowervscluster_types.go#L134) in status as slice so name was required to identify a loadbalancer, During our api review we decided to have map instead of list hence the name is no more required.

mkumatag (approved):
/lgtm

PR #1488, opened 2023-11-09
Add proposal for Power VS infra creation

Karthik-K-N:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

/area provider/ibmcloud

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

PR #1608, opened 2024-02-09
Feature: PowerVS cluster creation with dynamic resource creation

Karthik-K-N:

This PR adds the ability to create required resources for creating Power VS cluster dynamically eliminating user to precreate resources before creating cluster. A reference to proposal has been added which contains the detailed implementation plan for this feature.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Add ability to create required resource dynamically for PowerVS cluster creation
```

Proposal: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1488
API Pr: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1485
            https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1592

TODO:
* [ ] Fix and add unit test cases
* [ ] Add default and validating webhooks
* [ ] Update cos object creation and deletion logic


mkumatag:
> /cc @mkumatag @dharaneeshvrd @Amulyam24 @Prajyot-Parab

team lets review this code by this weekend

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	s.Info("Created new service instance")
+	return serviceInstance, nil
+}
+
+// ReconcileNetwork reconciles network.
+func (s *PowerVSClusterScope) ReconcileNetwork() error {
+	// if DHCP server id is set means the server is already created
+	if s.GetDHCPServerID() != nil {
+		s.Info("DHCP server id is set")
+		err := s.checkDHCPServerStatus()
```
dharaneeshvrd, line 655:
```suggestion
		if err := s.checkDHCPServerStatus(); err != nil
```

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+}
+
+// checkDHCPServerStatus checks the DHCP server status.
+func (s *PowerVSClusterScope) checkDHCPServerStatus() error {
+	dhcpID := *s.GetDHCPServerID()
+	dhcpServer, err := s.IBMPowerVSClient.GetDHCPServer(dhcpID)
+	if err != nil {
+		return err
+	}
+	if dhcpServer == nil {
```
dharaneeshvrd, line 733: we can remove this check, dhcpclient.get returns either err or dhcpserver obj. If err is nil, dhcpserver obj won't be nil.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+}
+
+// TODO(karthik-k-n): Decide on proper naming format for services
+
+// GetServiceName returns name of given service type from spec or generate a name for it.
+func (s *PowerVSClusterScope) GetServiceName(resourceType ResourceType) *string {
+	switch resourceType {
+	case ServiceInstance:
+		if s.ServiceInstance() == nil || s.ServiceInstance().Name == nil {
+			return pointer.String(fmt.Sprintf("%s-serviceInstance", s.InfraCluster().GetName()))
```
dharaneeshvrd, line 1498: Regarding the suffix, sometime back had a discussion with IPI team and decided to use common suffix, for powervs and other few resources which hypershift infra creation also follows currently. Also keeping longer suffix would make difficult to find the cluster resources in ibm cloud portal. Keeping it short would be better IMO.
powervs - pvs
vpc - vpc
vpc subnet - vpc-sn
transit gateway - tg
cos - image-registry - this alone shared by ibmcloud platform as well
Please see whether using this would make sense!

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	options := &vpcv1.CreateSecurityGroupRuleOptions{}
+	options.SetSecurityGroupID(*vpcDetails.DefaultSecurityGroup.ID)
+	options.SetSecurityGroupRulePrototype(&vpcv1.SecurityGroupRulePrototype{
+		Direction: core.StringPtr("inbound"),
+		Protocol:  core.StringPtr("tcp"),
+		IPVersion: core.StringPtr("ipv4"),
+		PortMin:   core.Int64Ptr(int64(s.APIServerPort())),
+		PortMax:   core.Int64Ptr(int64(s.APIServerPort())),
+	})
+	_, _, err = s.IBMVPCClient.CreateSecurityGroupRule(options)
```
dharaneeshvrd, line 859:
```suggestion
	if _, _, err = s.IBMVPCClient.CreateSecurityGroupRule(options); err != nil {
```

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	}
+	return vpcDetails.ID, nil
+}
+
+// ReconcileVPCSubnet reconciles VPC subnet.
+func (s *PowerVSClusterScope) ReconcileVPCSubnet() error {
+	//TODO: Should we handle a case where there are no subnet specified
+	for _, subnet := range s.IBMPowerVSCluster.Spec.VPCSubnets {
+		if subnet.Name == nil {
+			subnet.Name = pointer.String(fmt.Sprintf("%s-vpcsubnet", s.InfraCluster().GetName()))
```
dharaneeshvrd, line 871: No need to keep this in `GetServiceName()`?

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	return nil
+}
+
+// checkVPCSubnet checks VPC subnet exist in cloud.
+func (s *PowerVSClusterScope) checkVPCSubnet(subnet infrav1beta2.Subnet) (string, error) {
+	vpcSubnet, err := s.IBMVPCClient.GetVPCSubnetByName(*subnet.Name)
+	if err != nil {
+		return "", err
+	}
+	if vpcSubnet == nil {
```
dharaneeshvrd, line 921: Not required like dhcp get

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+}
+
+// checkVPC checks VPC exist in cloud.
+func (s *PowerVSClusterScope) checkVPC() (string, error) {
+	vpcDetails, err := s.getVPC()
+	if err != nil {
+		s.Error(err, "failed to get vpc")
+		return "", err
+	}
+	if vpcDetails == nil {
```
dharaneeshvrd, line 815: Not required like dhcp get. I have checked, you can reverify once in vpc sdk get call.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+			return err
+		}
+		s.SetStatus(COSInstance, infrav1beta2.ResourceReference{ID: cosServiceInstanceStatus.GUID, ControllerCreated: pointer.Bool(true)})
+	}
+
+	apiKey := os.Getenv("IBMCLOUD_API_KEY")
+	if apiKey == "" {
+		fmt.Printf("ibmcloud api key is not provided, set %s environmental variable", "IBMCLOUD_API_KEY")
+	}
+	apiKey = APIKEY
```
dharaneeshvrd, line 1312: how APIKEY is set?
I m seeing an To-Do remove for this. You may need to properly handle here!

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	if err != nil {
+		return nil, fmt.Errorf("error getting id for resource group %s, %w", *s.ResourceGroup(), err)
+	}
+
+	//TODO(karthik-k-n)(Doubt): should this be fetched or hardcoded
+	//_, servicePlanID, err := s.CatalogClient.GetServiceInfo(powerVSService, powerVSServicePlan)
+	//if err != nil {
+	//	return nil, fmt.Errorf("error retrieving id info for powervs service %w", err)
+	//}
+
```
dharaneeshvrd, line 1407: IMO can be hardcoded! It is not going to change!

controllers/ibmpowervscluster_controller.go
```diff
@@ -101,25 +107,168 @@ func (r *IBMPowerVSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		conditions.MarkFalse(powerVSCluster, infrav1beta2.LoadBalancerReadyCondition, infrav1beta2.LoadBalancerReconciliationFailedReason, capiv1beta1.ConditionSeverityError, err.Error())
+		return reconcile.Result{}, err
+	}
+
+	// reconcile COSInstance
+	clusterScope.Info("Reconciling COSInstance")
+	if err := clusterScope.ReconcileCOSInstance(); err != nil {
+		conditions.MarkFalse(powerVSCluster, infrav1beta2.COSInstanceReadyCondition, infrav1beta2.COSInstanceReconciliationFailedReason, capiv1beta1.ConditionSeverityError, err.Error())
+		return reconcile.Result{}, err
+	}
```
dharaneeshvrd, line 184: don't infrav1beta2.COSInstanceReadyCondition set to be true here?

cloud/scope/powervs_cluster.go
Karthik-K-N: Its handled in latest changes by Prajyot

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	}
+
+	tgConnections, _, err := s.TransitGatewayClient.ListTransitGatewayConnections(&tgapiv1.ListTransitGatewayConnectionsOptions{
+		TransitGatewayID: transitGateway.ID,
+	})
+	if err != nil {
+		return fmt.Errorf("error listing transit gateway connections: %w", err)
+	}
+
+	for _, conn := range tgConnections.Connections {
```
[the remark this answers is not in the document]
Karthik-K-N: Handled the case when connections are 0 but added a TODO to verify.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+			s.SetVPCSubnetID(*subnet.Name, infrav1beta2.ResourceReference{ID: &vpcSubnetID, ControllerCreated: pointer.Bool(false)})
+			// check for next subnet
+			continue
+		}
+		subnetID, err = s.createVPCSubnet(subnet)
+		if err != nil {
+			s.Error(err, "error creating vpc subnet")
+			return err
+		}
+		if subnetID != nil {
```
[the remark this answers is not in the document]
Karthik-K-N: I think its required as the subnet may not be exist in status

cloud/scope/powervs_cluster.go
Karthik-K-N: I think its required as vpc may not be found in cloud

cloud/scope/powervs_cluster.go
Karthik-K-N: I think its required as vpcsubnet may not be found in cloud

Karthik-K-N:
Thanks for Dhraneesh for reviewing, I have addressed the comments and marked it as resolved whichever is possible, We will discuss for unresolved comments to finalize the approaches.

cloud/scope/powervs_machine.go
```diff
@@ -141,25 +153,38 @@ func NewPowerVSMachineScope(params PowerVSMachineScopeParams) (scope *PowerVSMac
...
+	serviceInstance, err := rc.GetServiceInstance(serviceInstanceID, serviceInstanceName)
 	if err != nil {
-		err = fmt.Errorf("failed to get resource instance: %w", err)
+		params.Logger.Error(err, "error failed to get service instance details", "name", serviceInstanceName, "id", serviceInstanceID)
 		return nil, err
 	}
+	if serviceInstance == nil {
+		return nil, fmt.Errorf("service instance %s is not yet created", serviceInstanceName)
+	}
+	if *serviceInstance.State != "active" {
```
Amulyam24, line 173:
```suggestion
	if *serviceInstance.State != string(infrav1beta2.ServiceInstanceStateActive) {
```

cloud/scope/powervs_machine.go
```diff
@@ -295,24 +485,89 @@ func (m *PowerVSMachineScope) DeleteMachine() error {
...
+	serviceInstance, err := m.ResourceClient.GetInstanceByName(cosInstanceName, cosResourceID, cosResourcePlanID)
+	if err != nil {
+		m.Error(err, "failed to get cos service instance", "name", cosInstanceName)
+		return nil, err
+	}
+	if serviceInstance == nil {
+		m.Info("cos service instance is nil")
+		return nil, err
+	}
+	if *serviceInstance.State != "active" {
```
Amulyam24, line 529:
```suggestion
	if *serviceInstance.State != string(infrav1beta2.ServiceInstanceStateActive) {
```

pkg/cloud/services/resourcecontroller/service.go
```diff
@@ -43,14 +70,114 @@ func (s *Service) GetResourceInstance(getResourceInstanceOptions *resourcecontro
...
+	if err := utils.PagingHelper(f); err != nil {
+		return nil, fmt.Errorf("error listing service instances %v", err)
+	}
+	switch len(serviceInstancesList) {
+	case 0:
+		return nil, nil
+	case 1:
+		return &serviceInstancesList[0], nil
+	default:
+		errStr := fmt.Errorf("there exist more than one service instance ID with with same name %s, Try setting serviceInstance.ID", name)
```
Amulyam24, line 128:
```suggestion
		errStr := fmt.Errorf("there exist more than one service instance ID with same name %s, Try setting serviceInstance.ID", name)
```

pkg/cloud/services/resourcecontroller/service.go
```diff
@@ -43,14 +70,114 @@ func (s *Service) GetResourceInstance(getResourceInstanceOptions *resourcecontro
...
+	if err := utils.PagingHelper(f); err != nil {
+		return nil, fmt.Errorf("error listing COS instances %v", err)
+	}
+	switch len(serviceInstancesList) {
+	case 0:
+		return nil, nil
+	case 1:
+		return &serviceInstancesList[0], nil
+	default:
+		errStr := fmt.Errorf("there exist more than one COS instance ID with with same name %s, Try setting serviceInstance.ID", name)
```
Amulyam24, line 173:
```suggestion
		errStr := fmt.Errorf("there exist more than one COS instance ID with same name %s, Try setting serviceInstance.ID", name)
```

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +292,1545 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+		ResourcePlanID: pointer.String(cosResourcePlanID),
+	})
+	if err != nil {
+		return nil, err
+	}
+	return serviceInstance, nil
+}
+
+// getResourceGroupID retrieving id of resource group.
+func (s *PowerVSClusterScope) getResourceGroupID() (string, error) {
```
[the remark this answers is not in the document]
Karthik-K-N: made changes to set resource group id in status.

cloud/scope/powervs_cluster.go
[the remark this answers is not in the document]
Karthik-K-N: You are correct its redundant check, I removed it.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+		} else {
+			return false, err
+		}
+	}
+	return true, nil
+}
+
+func (s *PowerVSClusterScope) createCOSBucket() error {
+	input := &s3.CreateBucketInput{
+		Bucket: pointer.String(s.COSInstance().BucketName),
```
Amulyam24, line 1335: There's a scenario where when we retry bucket creation immediately with the same name,  it throws an error saying bucket exists even though the service instance and bucket get deleted successfully. Maybe we can handle by adding a random string to the bucket name?
(check out `util.RandomString()` from `sigs.k8s.io/cluster-api/util`)

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+			s.Error(err, "error deleting COS service instance")
+			return err
+		}
+		s.Info("COS instance successfully deleted")
+	}
+	return nil
+}
+
+// deleteResource returns true or false to decide on deleting provided resource.
+func (s *PowerVSClusterScope) deleteResource(resourceType ResourceType) bool { //nolint:gocyclo
```
dharaneeshvrd, line 1753: or some other meaning func name. 
```suggestion
func (s *PowerVSClusterScope) resourceCreatedByController(resourceType ResourceType) bool { //nolint:gocyclo
```
same as other func, the name looks like, its deleting some resource

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+					return nil
+				}
+				return fmt.Errorf("error fetching the load balancer: %w", err)
+			}
+
+			if lb != nil && lb.ProvisioningStatus != nil && *lb.ProvisioningStatus != string(infrav1beta2.VPCLoadBalancerStateDeletePending) {
+				_, err = s.IBMVPCClient.DeleteLoadBalancer(&vpcv1.DeleteLoadBalancerOptions{
+					ID: lb.ID,
+				})
+				if err != nil {
```
dharaneeshvrd, line 1529:
```suggestion
				if _, err = s.IBMVPCClient.DeleteLoadBalancer(&vpcv1.DeleteLoadBalancerOptions{
					ID: lb.ID,
				}); err != nil {
```

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+		}
+
+		tgConnections, _, err := s.TransitGatewayClient.ListTransitGatewayConnections(&tgapiv1.ListTransitGatewayConnectionsOptions{
+			TransitGatewayID: tg.ID,
+		})
+		if err != nil {
+			return fmt.Errorf("error listing transit gateway connections: %w", err)
+		}
+
+		if tgConnections.Connections != nil && len(tgConnections.Connections) > 0 {
```
dharaneeshvrd, line 1627: this if is not required since Connections is not a pointer to the array, if the connection len is zero, it won't loop.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+					}
+				}
+			}
+		}
+
+		_, err = s.TransitGatewayClient.DeleteTransitGateway(&tgapiv1.DeleteTransitGatewayOptions{
+			ID: s.IBMPowerVSCluster.Status.TransitGateway.ID,
+		})
+
+		if err != nil {
```
dharaneeshvrd, line 1645:
```suggestion
		if _, err = s.TransitGatewayClient.DeleteTransitGateway(&tgapiv1.DeleteTransitGatewayOptions{
			ID: s.IBMPowerVSCluster.Status.TransitGateway.ID,
		}); err != nil {
```

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+			if strings.Contains(err.Error(), "dhcp server does not exist") {
+				return nil
+			}
+			return fmt.Errorf("error fetching DHCP server: %w", err)
+		}
+
+		err = s.IBMPowerVSClient.DeleteDHCPServer(*server.ID)
+		if err != nil {
+			return fmt.Errorf("error deleting the DHCP server: %w", err)
+		}
```
dharaneeshvrd, line 1671:
```suggestion
		if err = s.IBMPowerVSClient.DeleteDHCPServer(*server.ID); err != nil {
			return fmt.Errorf("error deleting the DHCP server: %w", err)
		}
```

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+			if strings.Contains(err.Error(), "Subnet not found") {
+				return nil
+			}
+			return fmt.Errorf("error fetching the subnet: %w", err)
+		}
+
+		_, err = s.IBMVPCClient.DeleteSubnet(&vpcv1.DeleteSubnetOptions{
+			ID: net.ID,
+		})
+		if err != nil {
```
dharaneeshvrd, line 1565: inline if ?

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+			ID: subnet.ID,
+		})
+
+		if err != nil {
+			if strings.Contains(err.Error(), "Subnet not found") {
+				return nil
+			}
+			return fmt.Errorf("error fetching the subnet: %w", err)
+		}
+
```
dharaneeshvrd, line 1561: I think you need to wait for load balancer to get deleted before proceeding to subnet deletion or else it will fail.
Please see whether we need a check here like how a check is there for dhcpserver before deleting powervs instance.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+			if strings.Contains(err.Error(), "VPC not found") {
+				return nil
+			}
+			return fmt.Errorf("error fetching the VPC: %w", err)
+		}
+
+		_, err = s.IBMVPCClient.DeleteVPC(&vpcv1.DeleteVPCOptions{
+			ID: vpc.ID,
+		})
+		if err != nil {
```
dharaneeshvrd, line 1594: inline if?

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+		servers, err := s.IBMPowerVSClient.GetAllDHCPServers()
+		if err != nil {
+			return fmt.Errorf("error fetching networks in the service instance: %w", err)
+		}
+
+		if len(servers) > 0 {
+			return fmt.Errorf("cannot delete service instance as network is not yet deleted")
+		}
+
+		_, err = s.ResourceClient.DeleteResourceInstance(&resourcecontrollerv2.DeleteResourceInstanceOptions{
```
dharaneeshvrd, line 1704: inline if

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+				return nil
+			}
+			return fmt.Errorf("error fetching COS instance: %w", err)
+		}
+
+		if cosInstance != nil && (*cosInstance.State == "pending_reclamation" || *cosInstance.State == string(infrav1beta2.ServiceInstanceStateRemoved)) {
+			return nil
+		}
+
+		_, err = s.ResourceClient.DeleteResourceInstance(&resourcecontrollerv2.DeleteResourceInstanceOptions{
```
dharaneeshvrd, line 1739: inline if

cloud/scope/powervs_cluster.go
Amulyam24: The idea behind this was that on `reconcileDelete`, we are not waiting for one resource to get deleted before proceeding to the other resource but rather call delete API on all the resources and return an aggregate of errors at the end. One thing is it logs the error before a successful deletion in the subsequent reconciliations.

I think it might get redundant adding check for LB here again since we check it with `DeleteLoadBalancer` in the previous step.

> Please see whether we need a check here like how a check is there for dhcpserver before deleting powervs instance.

For deleting powers instance, the check was added because even though the delete API returns with success, the service instance is not getting deleted when checked on UI. We thought it would be better to wait and delete any resources within it before deleting the workspace.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+			ID: s.IBMPowerVSCluster.Status.VPC.ID,
+		})
+
+		if err != nil {
+			if strings.Contains(err.Error(), "VPC not found") {
+				return nil
+			}
+			return fmt.Errorf("error fetching the VPC: %w", err)
+		}
+
```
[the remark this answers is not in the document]
Amulyam24: Replied via https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1608#discussion_r1494240913

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1549 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	}
+
+	tgConnections, _, err := s.TransitGatewayClient.ListTransitGatewayConnections(&tgapiv1.ListTransitGatewayConnectionsOptions{
+		TransitGatewayID: transitGateway.ID,
+	})
+	if err != nil {
+		return fmt.Errorf("error listing transit gateway connections: %w", err)
+	}
+
+	if len(tgConnections.Connections) == 0 {
```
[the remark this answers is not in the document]
Karthik-K-N: I have updated the logic, Please take a look

cloud/scope/powervs_cluster.go
dharaneeshvrd: Got it, only concerns me is that the intermediate error status user would see in ibmpowervscluster obj during destroy.
One example could be in future if some other external program query for this obj on an interval, not sure it would be good to throw these intermediate errors.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1557 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+			vpcAttached = true
+		}
+		if *conn.NetworkType == "power_virtual_server" && *conn.NetworkID == *pvsServiceInstanceCRN {
+			if *conn.Status != "attached" {
+				return fmt.Errorf("error powervs connection not attached to transit gateway, current status: %s", *conn.Status)
+			}
+			powerVSAttached = true
+		}
+	}
+	if !powerVSAttached || !vpcAttached {
```
dharaneeshvrd, line 1071: It seems it won't reach here if any one of the case is false since you are already returning from respective case.

Karthik-K-N: yeah, The idea is given both the cases are true, we want to make sure both vpc and powervs is attached in tg connection

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1558 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
 }
+
+// Name returns the CAPI cluster name.
+func (s *PowerVSClusterScope) Name() string {
+	return s.Cluster.Name
+}
+
+// Zone returns the cluster zone.
+func (s *PowerVSClusterScope) Zone() *string {
+	return s.IBMPowerVSCluster.Spec.Zone
```
mkumatag, line 272: what happens when someone calls when this zone is null?!

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1558 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+}
+
+// ResourceGroup returns the cluster resource group.
+func (s *PowerVSClusterScope) ResourceGroup() *infrav1beta2.IBMPowerVSResourceReference {
```
mkumatag, line 276: again this * may create issues when something is wrong, wondering if we can have a simple null check?

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1558 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+func (s *PowerVSClusterScope) InfraCluster() ClusterObject {
+	return s.IBMPowerVSCluster
+}
+
+// APIServerPort returns the APIServerPort to use when creating the ControlPlaneEndpoint.
+func (s *PowerVSClusterScope) APIServerPort() int32 {
+	if s.Cluster.Spec.ClusterNetwork != nil && s.Cluster.Spec.ClusterNetwork.APIServerPort != nil {
+		return *s.Cluster.Spec.ClusterNetwork.APIServerPort
+	}
+	return 6443
```
mkumatag, line 290: can we create a constant somewhere for this port and use it?

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1558 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+}
+
+// ServiceInstance returns the cluster ServiceInstance.
+func (s *PowerVSClusterScope) ServiceInstance() *infrav1beta2.IBMPowerVSResourceReference {
+	return s.IBMPowerVSCluster.Spec.ServiceInstance
+}
+
+// GetServiceInstanceID get the service instance id.
+func (s *PowerVSClusterScope) GetServiceInstanceID() string {
+	// support the existing worklfow
```
mkumatag, line 300: can you add more information here about the flow we written here?

mkumatag: or may be a generic information above which applies for all the Get calls

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1558 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+		return *s.IBMPowerVSCluster.Spec.ResourceGroup.ID, nil
+	}
+	if s.IBMPowerVSCluster.Status.ResourceGroup != nil && s.IBMPowerVSCluster.Status.ResourceGroup.ID != nil {
+		return *s.IBMPowerVSCluster.Status.ServiceInstance.ID, nil
+	}
+	resourceGroupID, err := s.fetchResourceGroupID()
+	if err != nil {
+		return "", err
+	}
+	s.SetStatus(ResourceGroup, infrav1beta2.ResourceReference{ID: &resourceGroupID, ControllerCreated: pointer.Bool(false)})
```
mkumatag, line 536: why new setting this value in the Get method? 🤔

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +261,1558 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+		if err != nil {
+			return err
+		}
+		if serviceInstance == nil {
+			return fmt.Errorf("error failed to get service instance with id %s", serviceInstanceID)
+		}
+		if *serviceInstance.State != string(infrav1beta2.ServiceInstanceStateActive) {
+			return fmt.Errorf("service instance not in active state, current state: %s", *serviceInstance.State)
+		}
+		s.Info("Found service instance and its in active state", "id", serviceInstanceID)
```
mkumatag, line 557: I see this can be a code you can run post creation of resource in the end as well.

mkumatag: applies this comment for all other reconciles as well

mkumatag (changes requested):
first set of review comments

cloud/scope/powervs_cluster.go
Karthik-K-N: For zone and resource group we are adding webhook to make it mandatory for user to set when using create infra workflow which is identified with help of annotations. 
But even if we want to be safe then we need to add nil check under each caller of this function.

cloud/scope/powervs_cluster.go
Karthik-K-N: Created a constant for it.

cloud/scope/powervs_cluster.go
Karthik-K-N: sure will do

cloud/scope/powervs_cluster.go
Karthik-K-N: We are expecting user to pass resource group name, So we only set the id. The GetResourceGroupID is called from various places but only for the first call the status will be set and in future it will be fetched from status itself.

cloud/scope/powervs_cluster.go
```diff
@@ -59,93 +98,158 @@ type PowerVSClusterScope struct {
...
+		return nil, fmt.Errorf("error failed to create resource client: %w", err)
 	}
-	scope.IBMPowerVSClient = c
 
-	return scope, nil
+	clusterScope := &PowerVSClusterScope{
+		session:              session,
+		Logger:               params.Logger,
+		Client:               params.Client,
+		patchHelper:          helper,
```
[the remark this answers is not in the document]
Karthik-K-N: I have moved defer after checking for error https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/blob/f62f4958cf40d3c2a9aad20c5f03e68e0ba816cd/controllers/ibmpowervscluster_controller.go#L88-L98, I think there is nothing to patch at that moment.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +237,1558 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+
+// GetServiceName returns name of given service type from spec or generate a name for it.
+func (s *PowerVSClusterScope) GetServiceName(resourceType infrav1beta2.ResourceType) *string {
+	switch resourceType {
+	case infrav1beta2.ResourceTypeServiceInstance:
+		if s.ServiceInstance() == nil || s.ServiceInstance().Name == nil {
+			return pointer.String(fmt.Sprintf("%s-serviceInstance", s.InfraCluster().GetName()))
+		}
+		return s.ServiceInstance().Name
+	case infrav1beta2.ResourceTypeNetwork:
```
Karthik-K-N, line 1474: Fix this: Currently DHCP server creates network and its of some format as given here, If user provides name this wont work. I think dhcp server name cannot be set as we think. Explore!!

util/util.go
```diff
@@ -39,3 +40,160 @@ func GetClusterByName(ctx context.Context, c client.Client, namespace, name stri
...
+	Description string
+	VPCRegion   string
+	COSRegion   string
+	Zones       []string
+	VPCZones    []string
+	SysTypes    []string
+}
+
+// Regions provides a mapping between Power VS and IBM Cloud VPC and IBM COS regions.
+var Regions = map[string]Region{
```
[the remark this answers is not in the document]
mkumatag: dependent's k8s version is dealt via go.sum file I feel

cloud/scope/powervs_cluster.go
mkumatag: but its better to call that separately and set the status instead of in this function

cloud/scope/powervs_cluster.go
mkumatag: its always safe to have null check :)

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -92,6 +97,22 @@ type IBMPowerVSClusterSpec struct {
 	CosInstance *CosInstance `json:"cosInstance,omitempty"`
 }
 
+// DHCPServer contains the DHCP server configurations
+type DHCPServer struct {
+	// Optional cidr for DHCP private network
+	Cidr *string `json:"cidr,omitempty"`
```
mkumatag, line 103: lets add some regex to validate this

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -92,6 +97,22 @@ type IBMPowerVSClusterSpec struct {
...
+
+	// Optional DNS Server for DHCP service
+	// +kubebuilder:default="1.1.1.1"
+	DNSServer *string `json:"dnsServer,omitempty"`
+
+	// Optional name of DHCP Service. Only alphanumeric characters and dashes are allowed (will be prefixed by DHCP identifier)
+	Name *string `json:"name,omitempty"`
```
mkumatag, line 110: what if user specifies the id as well?

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -92,6 +97,22 @@ type IBMPowerVSClusterSpec struct {
...
+
+	// Indicates if SNAT will be enabled for DHCP service
+	SnatEnabled *bool `json:"snatEnabled,omitempty"`
```
mkumatag, line 113: this should be true by default and you can rename this to `snat` to keep it simple as this is already boolean.

mkumatag (changes requested):

api/v1beta2/ibmpowervscluster_types.go
Karthik-K-N: Updated.

api/v1beta2/ibmpowervscluster_types.go
Karthik-K-N: Added ID field as well.

api/v1beta2/ibmpowervscluster_types.go
Karthik-K-N: Yet to find a valid regex, will update soon

cloud/scope/powervs_cluster.go
Karthik-K-N: Updated the logic to handle the scenarios when user provided DHCPserver name as well.

cloud/scope/powervs_cluster.go
Karthik-K-N: Updated.

cloud/scope/powervs_cluster.go
Karthik-K-N: Added check where ever the value is being referred.

cloud/scope/powervs_cluster.go
Karthik-K-N: Added check where ever the value is being referred.

util/util.go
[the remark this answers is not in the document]
Karthik-K-N: The issue I saw is, When I tried to use powervs utils it uses k8s version 1.29 but capi is using 1.28, It updated the verision in capi

capi go.mod file
```
        k8s.io/apiextensions-apiserver v0.28.4
-       k8s.io/apimachinery v0.28.4
+       k8s.io/apimachinery v0.29.1
        k8s.io/cli-runtime v0.28.4
        k8s.io/client-go v0.28.4
-       k8s.io/klog/v2 v2.100.1
-       k8s.io/utils v0.0.0-20230406110748-d93618cff8a2
+       k8s.io/klog/v2 v2.110.1
+       k8s.io/utils v0.0.0-20230726121419-3b25d923346b
```

Amulyam24:
Review comments from discussion today:

- [ ] 1. Add comments and documentation for easy understanding of code
- [ ] 2. Regex should be handled properly for all resources
- [ ] 3. CIDR regex validation for DHCP server
- [ ] 4. Handle Set() resource in ibmpowervscluster_types.go — complicating the code L243
- [ ] 5. Think about how to test the machine types in staging environment
- [ ] 6. Check why session is being used/added in PowerVSClusterScope L77
- [ ] 7. Why ClusterObject is being used?  just return object name, remove setter
- [ ] 8. Optimise zone handling in PowerVSClusterScope() L125
- [ ] 9. Add webhook validation for VPC region as well L190
- [ ] 10. Rename APIServerPort as DefaultAPIServerPort
- [ ] 11. SetNetworkStatus move to the common Set method
- [ ] 12. Change the order in GetDHCPServerID
- [ ] 13. Remove get for GetDHCPServer in parity with others
- [ ] 14. GetSubnetID() - instead of sending whole object, just send the name 
- [ ] 15. GetSubnetIDs() — no need to return array of string pointers
- [ ] 16. Remove additional check(controllerCreated) in SetVPCSubnetID() L416 L458
- [ ] 17. Check wherever pointers are not necessary(check for nil values)
- [ ] 19. GetLoadBanlancerID() L466 - just pass name 
- [ ] 21. GetLoadBalancerhost — add name to match the value being returned — check for others as well 
- [ ] 22. Move SetCosBucket to common Set method L499
- [ ] 23. Handle GetResourceGroupID()  — don’t call the fetchResourcegroupID()   L522

cloud/scope/powervs_cluster.go
```diff
@@ -59,93 +74,158 @@ type PowerVSClusterScope struct {
...
+			ServiceEndpoint:   params.ServiceEndpoint,
+			IBMPowerVSClient:  powerVSClient,
+		}, nil
 	}
 
-	// Fetch the service endpoint.
-	if svcEndpoint := endpoints.FetchPVSEndpoint(endpoints.CostructRegionFromZone(*res.RegionID), params.ServiceEndpoint); svcEndpoint != "" {
-		options.IBMPIOptions.URL = svcEndpoint
-		scope.Logger.V(3).Info("Overriding the default powervs service endpoint")
+	if params.IBMPowerVSCluster.Spec.VPC == nil || params.IBMPowerVSCluster.Spec.VPC.Region == nil {
```
Amulyam24, line 190: Add webhook validation for VPC region as well

api/v1beta2/types.go
```diff
@@ -16,6 +16,9 @@ limitations under the License.
 
 package v1beta2
 
+// APIServerPort is API server port number.
+const APIServerPort int32 = 6443
```
Amulyam24, line 20: Rename DefaultAPIServerPort

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +237,1610 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	}
+}
+
+// Network returns the cluster Network.
+func (s *PowerVSClusterScope) Network() *infrav1beta2.IBMPowerVSResourceReference {
+	return &s.IBMPowerVSCluster.Spec.Network
+}
+
+// SetNetworkStatus set the network status.
+func (s *PowerVSClusterScope) SetNetworkStatus(resource infrav1beta2.ResourceReference) {
```
Amulyam24, line 345: move to the common Set method

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +237,1610 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	if s.IBMPowerVSCluster.Status.Network == nil {
+		s.IBMPowerVSCluster.Status.Network = &resource
+	}
+	s.IBMPowerVSCluster.Status.Network.Set(resource)
+}
+
+// GetDHCPServerID returns the DHCP id.
+func (s *PowerVSClusterScope) GetDHCPServerID() *string {
```
Amulyam24, line 353: Change the order of checking status and spec

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +237,1610 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+		return s.IBMPowerVSCluster.Status.DHCPServer.ID
+	}
+	if s.IBMPowerVSCluster.Spec.DHCPServer != nil && s.IBMPowerVSCluster.Spec.DHCPServer.ID != nil {
+		return s.IBMPowerVSCluster.Spec.DHCPServer.ID
+	}
+	return nil
+}
+
+// GetDHCPServer returns the DHCP server details.
+func (s *PowerVSClusterScope) GetDHCPServer() *infrav1beta2.DHCPServer {
```
Amulyam24, line 364: Rename as DHCPServer

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +237,1610 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	}
+	return nil
+}
+
+// GetVPCSubnetIDs returns all the VPC subnet ids.
+func (s *PowerVSClusterScope) GetVPCSubnetIDs() []*string {
+	if s.IBMPowerVSCluster.Status.VPCSubnet == nil {
+		return nil
+	}
+	subnets := []*string{}
```
Amulyam24, line 403: maybe no need to return array of string pointers

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +237,1610 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	return subnets
+}
+
+// SetVPCSubnetID set the VPC subnet id.
+func (s *PowerVSClusterScope) SetVPCSubnetID(name string, resource infrav1beta2.ResourceReference) {
+	if s.IBMPowerVSCluster.Status.VPCSubnet == nil {
+		s.IBMPowerVSCluster.Status.VPCSubnet = make(map[string]infrav1beta2.ResourceReference)
+	}
+	if val, ok := s.IBMPowerVSCluster.Status.VPCSubnet[name]; ok {
+		if val.ControllerCreated != nil && *val.ControllerCreated {
```
Amulyam24, line 416: remove check wherever unecessary

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +237,1610 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	return nil
+}
+
+// SetLoadBalancerStatus set the loadBalancer id.
+func (s *PowerVSClusterScope) SetLoadBalancerStatus(name string, loadBalancer infrav1beta2.VPCLoadBalancerStatus) {
+	if s.IBMPowerVSCluster.Status.LoadBalancers == nil {
+		s.IBMPowerVSCluster.Status.LoadBalancers = make(map[string]infrav1beta2.VPCLoadBalancerStatus)
+	}
+	if val, ok := s.IBMPowerVSCluster.Status.LoadBalancers[name]; ok {
+		if val.ControllerCreated != nil && *val.ControllerCreated {
```
Amulyam24, line 458: remove check wherever unnecessary

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +237,1610 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+		return nil
+	}
+	if val, ok := s.IBMPowerVSCluster.Status.LoadBalancers[name]; ok {
+		return val.Hostname
+	}
+	return nil
+}
+
+// SetCOSBucketStatus set the COS bucket id.
+func (s *PowerVSClusterScope) SetCOSBucketStatus(cosInstanceID string, controllerCreated bool) {
```
Amulyam24, line 499: move to common Set method

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +237,1610 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+
+// GetResourceGroupID returns the resource group id.
+func (s *PowerVSClusterScope) GetResourceGroupID() (string, error) {
+	if s.IBMPowerVSCluster.Spec.ResourceGroup != nil && s.IBMPowerVSCluster.Spec.ResourceGroup.ID != nil {
+		return *s.IBMPowerVSCluster.Spec.ResourceGroup.ID, nil
+	}
+	if s.IBMPowerVSCluster.Status.ResourceGroup != nil && s.IBMPowerVSCluster.Status.ResourceGroup.ID != nil {
+		return *s.IBMPowerVSCluster.Status.ResourceGroup.ID, nil
+	}
+	resourceGroupID, err := s.fetchResourceGroupID()
```
Amulyam24, line 522: maybe handle this in a better way by avoiding calling `fetchResourceGroupID` here

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +237,1610 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+		return nil
+	}
+	if val, ok := s.IBMPowerVSCluster.Status.LoadBalancers[name]; ok {
+		return &val.State
+	}
+	return nil
+}
+
+// GetLoadBalancerHost will return the hostname of load balancer.
+func (s *PowerVSClusterScope) GetLoadBalancerHost(name string) *string {
```
Amulyam24, line 488: rename to GetLoadBalancerHostName

api/v1beta2/ibmpowervscluster_types.go
```diff
@@ -222,6 +239,14 @@ func (r *IBMPowerVSCluster) SetConditions(conditions capiv1beta1.Conditions) {
 	r.Status.Conditions = conditions
 }
 
+// Set sets the details of the resource.
+func (rf *ResourceReference) Set(resource ResourceReference) {
+	rf.ID = resource.ID
+	if !*rf.ControllerCreated {
```
Amulyam24, line 245: re-evaluate this check

cloud/scope/powervs_cluster.go
```diff
@@ -59,93 +74,158 @@ type PowerVSClusterScope struct {
 	logr.Logger
 	Client      client.Client
 	patchHelper *patch.Helper
+	session     *ibmpisession.IBMPISession
```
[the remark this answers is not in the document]
Karthik-K-N: Reused across multiple function for example https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/blob/969051ba88de2179b8f4ed9bec8728f0711b6dd4/cloud/scope/powervs_cluster.go#L1434, Need to explore on reusing it even furthur.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +231,1605 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+		return s.DHCPServer().Name
+	case infrav1beta2.ResourceTypeCOSInstance:
+		if s.COSInstance() == nil || s.COSInstance().Name == "" {
+			return pointer.String(fmt.Sprintf("%s-cosinstance", s.InfraCluster()))
+		}
+		return &s.COSInstance().Name
+	case infrav1beta2.ResourceTypeSubnet:
+		return pointer.String(fmt.Sprintf("%s-vpcsubnet", s.InfraCluster()))
+	}
+	return nil
```
Amulyam24, line 1540: handle error if type is not present

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +231,1605 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+			s.Error(err, "error deleting COS service instance")
+			return err
+		}
+		s.Info("COS instance successfully deleted")
+	}
+	return nil
+}
+
+// resourceCreatedByController helps to identify resource created by controller or not.
+func (s *PowerVSClusterScope) resourceCreatedByController(resourceType infrav1beta2.ResourceType) bool { //nolint:gocyclo
```
Amulyam24, line 1779: rename to `isResourceCreatedByController` wherever applicable

cloud/scope/powervs_cluster.go
Amulyam24: @dharaneeshvrd, will be handling it in next iteration.

cloud/scope/powervs_cluster.go
Karthik-K-N: We will handle this in next phase.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +231,1605 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+	var zone string
+	if subnet.Zone != nil {
+		zone = *subnet.Zone
+	} else {
+		powerVSZone := s.Zone()
+		if powerVSZone == nil {
+			return nil, fmt.Errorf("error creating vpc subnet, powervs zone is not set")
+		}
+		region := endpoints.ConstructRegionFromZone(*powerVSZone)
+		vpcZones, err := genUtil.VPCZonesForPowerVSRegion(region)
```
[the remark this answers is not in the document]
Karthik-K-N: The region mapping is currently from powervs region to vpc https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/blob/67e8057bad2a4fadf233d4e8636199ba4795b1d1/util/util.go#L77, so we need to fetch powervs region from zone

go.mod
```diff
@@ -12,18 +12,25 @@ replace (
 require (
 	github.com/IBM-Cloud/power-go-client v1.5.9
 	github.com/IBM/go-sdk-core/v5 v5.15.1
+	github.com/IBM/ibm-cos-sdk-go v1.10.2
+	github.com/IBM/networking-go-sdk v0.44.0
 	github.com/IBM/platform-services-go-sdk v0.59.0
 	github.com/IBM/vpc-go-sdk v0.48.0
+	github.com/blang/semver/v4 v4.0.0
+	github.com/coreos/ignition v0.35.0
```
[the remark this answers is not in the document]
mkumatag: here is the CVE link -- https://github.com/advisories/GHSA-hj57-j5cw-2mwp, as this is a very old release, I don't think they will even patch this one. Lets keep this schema offline somewhere in the pkg folder.

cloud/scope/powervs_cluster.go
```diff
@@ -157,3 +246,1660 @@ func (s *PowerVSClusterScope) PatchObject() error {
...
+		return nil, err
+	}
+	return serviceInstance, nil
+}
+
+// fetchResourceGroupID retrieving id of resource group.
+func (s *PowerVSClusterScope) fetchResourceGroupID() (string, error) {
+	if s.ResourceGroup() == nil || s.ResourceGroup().Name == nil {
+		return "", fmt.Errorf("resource group name is not set")
+	}
```
mkumatag, line 1517: TODO: Use default resource group for the account if not specified.

cloud/scope/powervs_machine.go
```diff
@@ -295,24 +506,100 @@ func (m *PowerVSMachineScope) DeleteMachine() error {
 	return nil
 }
 
-// GetBootstrapData returns the base64 encoded bootstrap data from the secret in the Machine's bootstrap.dataSecretName.
-func (m *PowerVSMachineScope) GetBootstrapData() (string, error) {
+// DeleteMachineIgnition deletes the ignition associated with machine.
+func (m *PowerVSMachineScope) DeleteMachineIgnition() error {
```
mkumatag, line 510: where is this called?

Prajyot-Parab: @Karthik-K-N did you drop it by mistake? I have it in my other PR, if need be I can split it in a diff PR and push.

Karthik-K-N: Updated it in the latest change to delete ignition as part of machine deletion.

go.mod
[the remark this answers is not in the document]
Prajyot-Parab: fix - https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/pull/12

mkumatag:
@Prajyot-Parab @Karthik-K-N I see linting issue for that file, see if we can somehow ignore that for that entire package

Prajyot-Parab:
> @Prajyot-Parab @Karthik-K-N I see linting issue for that file, see if we can somehow ignore that for that entire package

fixed.

mkumatag:
I have approved the PR, one of you @Amulyam24 @Prajyot-Parab give lgtm

Prajyot-Parab (approved):
/lgtm

PR #1637, opened 2024-02-28
Extend supported service endpoints

cjschaef:
Add additional services to endpoint override support for VPC related services.

**What this PR does / why we need it**:  Expands the supported IBM Cloud Services that can have endpoint overrides.

/area provider/ibmcloud

**Release note**:

```release-note
Extend supported IBM Cloud Service endpoint overrides
```

### DIFF
--- a/docs/book/src/getting-started.md
+++ b/docs/book/src/getting-started.md
@@ -47,7 +47,7 @@ it into a management cluster using `clusterctl`.
    > `${ServiceRegion1}:${ServiceID1}=${URL1},${ServiceID2}=${URL2};${ServiceRegion2}:${ServiceID1}=${URL1...}`.
    
 
-    Supported ServiceIDs include - `vpc, powervs, rc`
+    Supported ServiceIDs include - `CIS, COS, DNSServices, IAM, KeyProtect, ResourceManager, powervs, rc, vpc`
      ```console
       export SERVICE_ENDPOINT=us-south:vpc=https://us-south-stage01.iaasdev.cloud.ibm.com,powervs=https://dal.power-iaas.test.cloud.ibm.com,rc=https://resource-controller.test.cloud.ibm.com
      ```
--- a/pkg/endpoints/endpoints.go
+++ b/pkg/endpoints/endpoints.go
@@ -27,17 +27,29 @@ import (
 var ServiceEndpointFormat string
 
 const (
-	// VPC used to identify VPC service.
-	VPC serviceID = "vpc"
+	// CIS used to identify Cloud Internet Services.
+	CIS serviceID = "CIS"
+	// COS used to identify Cloud Object Storage service.
+	COS serviceID = "COS"
+	// DNSServices used to identify DNS Services.
+	DNSServices serviceID = "DNSServices"
+	// IAM used to identify Identity and Management service.
+	IAM serviceID = "IAM"
+	// KeyProtect used to identify Key Protect service.
+	KeyProtect serviceID = "KeyProtect"
 	// PowerVS used to identify PowerVS service.
 	PowerVS serviceID = "powervs"
 	// RC used to identify Resource-Controller service.
 	RC serviceID = "rc"
+	// ResourceManager used to identify Resource Manager service.
+	ResourceManager serviceID = "ResourceManager"
+	// VPC used to identify VPC service.
+	VPC serviceID = "vpc"
 )
 
 type serviceID string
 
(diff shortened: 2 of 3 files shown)


jeffnowicki:
Note: we're following OpenShift convention for service names being `PascalCase`. While we don't want to impose any specific consumer conventions, in this particular case it would ease our (VPC) OpenShift use of CAPI if we could follow similar conventions (where it makes sense). Let us know your thoughts and in you are open to refactoring the current names to `PascalCase` also.

mkumatag:
> Note: we're following OpenShift convention for service names being `PascalCase`. While we don't want to impose any specific consumer conventions, in this particular case it would ease our (VPC) OpenShift use of CAPI if we could follow similar conventions (where it makes sense). Let us know your thoughts and if you are open to refactoring the current names to `PascalCase` also.

I don't mind but just make sure we keep the backward compatibility as well, so that existing consumers shouldn't be suffering from this refactor. e.g: we can deprecate `vpc` in this release and introduce `VPC` and next release you can remove `vpc` from the code.

docs/book/src/getting-started.md
```diff
@@ -47,7 +47,7 @@ it into a management cluster using `clusterctl`.
    > `${ServiceRegion1}:${ServiceID1}=${URL1},${ServiceID2}=${URL2};${ServiceRegion2}:${ServiceID1}=${URL1...}`.
    
 
-    Supported ServiceIDs include - `vpc, powervs, rc`
+    Supported ServiceIDs include - `CIS, COS, DNSServices, IAM, KeyProtect, ResourceManager, powervs, rc, vpc`
```
Karthik-K-N, line 50: As of today, We dont user any of CIS, DNSServices, KeyProtect services. so should we add them now or add them when we start using these services later.

mkumatag: make sense, lets remove the entries which aren't used in this controller at the moment.

Prajyot-Parab: @Karthik-K-N is thee ResourceManager duplicate of rc here?

Prajyot-Parab: @Karthik-K-N also the `iam` endpoint is being set via `IBMCLOUD_AUTH_URL` param, so do we really need the IAM service ID?

Karthik-K-N: I think so its duplicate, if we want to use ResourceManager instead of rc, may be we can add support for both for a release and based on requirement we can deprecate later.

Prajyot-Parab: IMO its a duplicate and not required - https://pkg.go.dev/github.com/IBM/platform-services-go-sdk/resourcemanagerv2#pkg-constants

Karthik-K-N: Regarding iam, I think there is no way to use anything other than IBMCLOUD_AUTH_URL right, as its internally used by client?

Prajyot-Parab: Yes

> Regarding iam, I think there is no way to use anything other than IBMCLOUD_AUTH_URL right, as its internally used by client?

Yes, so having `iam` service id doesn't not make sense here i feel

jeffnowicki: additional information... we have our set of required endpoint services for OCP defined here: https://github.com/openshift/installer/blob/release-4.15/data/data/install.openshift.io_installconfigs.yaml#L3450-L3492

we default to public endpoints for these required services, but provide the ability to override via install-config arguments

that served our basis for this PR

mkumatag: Thanks for the information, not all those service endpoints from openshift will be used by the CAPI controller, e.g: CIS will be used by the ingress operator(part of OCP Cluster Operartors) which is out of scope for this project.

We will add which are really required for the CAPI controllers to run e.g: COS, IAM, ResourceController VPC etc..

cjschaef: I can add them later, but they could all possibly be required (depending on the Network design) as the Infrastructure resources get created/referenced in VPC related enhancements.

Or we can wait on merging this until I have completed the necessary enhancements to CAPI for VPC Infrastructure, if that appears to be a better option.

mkumatag: Let's have an incremental approach, you can merge the code with bare minimum required changes, later will add those additional endpoints as in when needed.

mkumatag:
we don't need this anymore, got it merged part of https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1648.

PR #1658, opened 2024-03-13
Update loadbalancer spec to configure private loadbalncer

Karthik-K-N:

Since Public is boolean variable , once we set it to false to create a private loadbalancer as false is unset value of boolean variable and using kubebuilder tag we are setting it default to true, so its making it not possible to create a private loadabalancer.

Now changing data type to boolean pointer so we can differentiate between unset value and false value.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note

```


mkumatag (approved):
/lgtm

PR #1668, opened 2024-03-18
VPC: Relocate COS and VPC types into common

cjschaef:
Relocate the CosInstance and VPCResourceReference
types from PowerVS code into common types, to
prevent duplicate code and allow VPC safe reuse.

**What this PR does / why we need it**:  Resources that could be shared between PowerVS and VPC, shouldn't be located in PowerVS code. Moving it to a shared location.

Fixes # N/A

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/api/v1beta2/ibmpowervscluster_types.go
+++ b/api/v1beta2/ibmpowervscluster_types.go
@@ -250,42 +250,6 @@ type TransitGateway struct {
 	ID *string `json:"id,omitempty"`
 }
 
-// VPCResourceReference is a reference to a specific VPC resource by ID or Name
-// Only one of ID or Name may be specified. Specifying more than one will result in
-// a validation error.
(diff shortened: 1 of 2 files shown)


Prajyot-Parab:
IMO we should not be touching the PowerVS resources at all at this point.

I suggest you go ahead and work on VPC part of things, even if it means we have some duplicacy. Lets get the VPC side of things working with new API specs.
Later we can work on optimising the common pieces once we have a confidence that both VPC and PowerVS flows are working fine with new API spec changes independently.

cc @mkumatag @Karthik-K-N 

Note - whenever you do any changes in API files, you should follow up by running `make generate`

cjschaef:
Okay, I will be attempting to make new definitions with unique names that hopefully do not collide with these PowerVS defined ones.

PR #1670, opened 2024-03-18
VPC: Create basic types for VPC

cjschaef:
Create a basic VPC resource type that VPC definitions will use.

**What this PR does / why we need it**: VPC updates will need references to generic VPC resources. This could be shared with PowerVS, if they wish to adopt them.

Fixes # N/A

**Special notes for your reviewer**: Replaces https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1668

/area provider/ibmcloud

**Release note**:

```release-note

```


Prajyot-Parab:
lgtm
@Karthik-K-N ptal

api/v1beta2/types.go
```diff
@@ -134,12 +134,42 @@ var (
...
+	// name defines the name of the COS Instance.
+	// +kubebuilder:validation:MinLength:=3
+	// +kubebuilder:validation:MaxLength:=63
+	// +kubebuilder:validation:Pattern=`^[a-z0-9][a-z0-9.-]{1,61}[a-z0-9]$`
+	// +required
+	Name string `json:"name"`
+
+	// id is the ID of the COS instance.
+	// +optional
+	ID *string `json:"id,omitempty"`
```
Prajyot-Parab, line 158: do we really need both Name and ID field?

api/v1beta2/types.go
```diff
@@ -134,12 +134,42 @@ var (
...
+
+	// bucketName is IBM Cloud COS bucket name.
+	// +optional
+	BucketName *string `json:"bucketName,omitempty"`
+
+	// bucketID is the IBM Cloud COS bucket ID.
+	// +optional
+	BucketID *string `json:"bucketID,omitempty"`
```
Prajyot-Parab, line 166: same comment as cos (do we really need both Name and ID field?)

Prajyot-Parab (changes requested):

api/v1beta2/types.go
Karthik-K-N: We usually have both Name and ID fields for resources, just incase want to speed up the look up when ID is known.

api/v1beta2/types.go
```diff
@@ -163,3 +193,20 @@ type VPCEndpoint struct {
...
+	// +optional
+	ID *string `json:"id,omitempty"`
+
+	// name of the resource.
+	// +kubebuilder:validation:MinLength=1
+	// +optional
+	Name *string `json:"name,omitempty"`
+
+	// type is the type of VPC resource.
+	Type *ResourceType `json:"type,omitempty"`
```
Karthik-K-N, line 211: Would like to understand how this will be used. any example?

api/v1beta2/types.go
cjschaef: I can remove, I simply copied the COSInstance from PowerVS code, as is.
If you have found you don't need name, I will remove it here as well.

api/v1beta2/types.go
cjschaef: Same as above, I can remove if you have found both to be unnecessary.

api/v1beta2/types.go
cjschaef: A VPCResource can be any kind of VPC resource (VSI, LB, SG, SGR, etc.).
Since I would only have a name or ID, it is helpful to have additional details as to what type of resource it is, IMHO.

api/v1beta2/types.go
cjschaef: Looks like the current PowerVS has no ID's, I kind of favor Id's over names and currently expect CAPI will not be creating COS Instances or resources....so I will see about using Id's over names ATM.
Will update.

api/v1beta2/types.go
cjschaef: Same update as above, will try only relying on ID's

api/v1beta2/types.go
cjschaef: I dropped Type for now.

api/v1beta2/types.go
cjschaef: I dropped BucketName for now.

api/v1beta2/types.go
cjschaef: I dropped `Name` for now.

Prajyot-Parab: @cjschaef It all depends on how you are going to setup the relevant controller code, will it need ID or just name or both or you can deduce ID from name in code itself or vice versa. We will leave taking call on which fields to have and not upto you however having clarity on how we plan to use them in future controller code helps us in reviewing/understanding better the entire flow.
/cc @Karthik-K-N 

Note - if we need both cant we reuse vpcResource struct for say cos and bucket obj? (just a thought) @Karthik-K-N your thoughts on this?

api/v1beta2/types.go
Prajyot-Parab: same comments as above, if you have a plan on how to use that ID in controller that might help us in understanding this better.

api/v1beta2/types.go
Karthik-K-N: We cannot drop Name as thats the only way user can specify the Name of COSInstance.

Also It wont be good to add VPCResource to COSInstance as both are two different entities in IBM Cloud. May be if we would rename VPCResource to something else to more appropriate as its just represents ID and Name in general, then only we can do.

cjschaef: My expectation at this time, is that the controller will not be creating COS resources, but it may be expected to create a VPC Custom Image, using a qcow2 file in a COS bucket, hence the need for this COS struct. I could rely only on a COS bucket CRN, but if this resource can be re-used, I expected Name/ID would be more beneficial, fields set depending on the requirements (PowerVS vs. VPC, etc.)

However, this is my best assumption on the workflow, as limitations may require changes and I am unsure if PowerVS creates COS resources (bucket maybe?) for the VSI Image. So now I wonder if VPC should deviate from the flow PowerVS uses.
It is early on in development, so I can mirror the flow PowerVS uses, for consistency, which would mean I follow similar field requirements (Name's)?

Perhaps you have an opinion on whether VPC should follow similar processes/expectations as PowerVS, when it makes sense. I am not opposed to relying on the controller to create COS resources if necessary, but I had expected the controller to rely on an existing COS Instance and Bucket and ideally the qcow2 Object holding the RHCOS image, and not be responsible for creating it.

Karthik-K-N: I think this is what causing confusion, There are two things
1. In PowerVS we create cos instance and bucket to [upload the ignition](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/cloud/scope/powervs_machine.go#L398) data as a object and pass the object URL as [userdata](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/cloud/scope/powervs_machine.go#L282) while creating machine.

2. We have ibmpowervs_image_controller which takes care of importing the RHCOS image from respective COS bucket to where redhat will push the image into the PowerVS service instance.

api/v1beta2/types.go
```diff
@@ -142,6 +142,21 @@ var (
...
+// COSInstanceReference represents an IBM Cloud COS instance.
+type COSInstanceReference struct {
+	// id is the ID of the COS instance.
+	// +required
+	ID *string `json:"id"`
+
+	// bucketID is the IBM Cloud COS bucket ID.
+	// +optional
+	BucketID *string `json:"bucketID,omitempty"`
+
```
Karthik-K-N, line 154: So if we want to reuse the same struct for PowerVS as well then lets add BucketName and Name here. 
So this allows us to create and use the COSinstance and Bucket with user provided name.

Just for clarification, In PowerVS we create these resources to upload the ignition data.

api/v1beta2/types.go
cjschaef: The question I guess I still have is, do the COS instance and bucket need to be created as a prerequisite? I hadn't seen anything in the CAPI code that creates the instance and the bucket (only the Object/File), but I certainly could have missed that.

If this is the case, and I would follow the same flow as PowerVS (create the COS instance and bucket outside CAPI), that will influence what values I want, although I have seen I am limited by the IBM COS SDK, as far as values as well.

api/v1beta2/types.go
```diff
@@ -142,6 +142,21 @@ var (
...
 
+// COSInstanceReference represents an IBM Cloud COS instance.
+type COSInstanceReference struct {
+	// id is the ID of the COS instance.
+	// +required
+	ID *string `json:"id"`
+
+	// bucketID is the IBM Cloud COS bucket ID.
+	// +optional
+	BucketID *string `json:"bucketID,omitempty"`
```
cjschaef, line 153: I need to swap back to BucketName, I don't think the IBM COS SDK provides additional details I want to use this.
Regardless of whether I expect CAPI to create COS instances and buckets.

mkumatag (changes requested):
@cjschaef I think we are mixing 2 things here - COS bucket for the image import and the ignition data. I recommend you @Karthik-K-N @Prajyot-Parab and @cjschaef please setup a meeting offline to discuss on the spec considering the amount of time we are left with for the implementation.

api/v1beta2/types.go
Karthik-K-N: We do create COS instance and bucket as a part of IBMPowerVSCluster reconciler https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/2a0c2811d02cd003c56327e485c54e262057c408/controllers/ibmpowervscluster_controller.go#L200-L203

cjschaef:
@mkumatag @Karthik-K-N 
Based on the findings I've had, reviewing the PowerVS code, expectations, and dependencies, my expectations for the VPC path, is not to rely on CAPI controllers to create COS resources (COS Instance, Bucket, or Object).

Instead, I plan to rely on the COS Instance ID, the Bucket Name, and likely the COS Object Name, with the alternative being I can rely on the Object URL instead, but that depends on IBM Cloud VPC API's, which I have to re-review.

If PowerVS does not wish or plan to share this `common` COS Instance type, I will not add any additional fields to accommodate PowerVS (e.g., COS Instance Name).

api/v1beta2/types.go
cjschaef: I plan to handle the process for VSI Custom Image using a VPC unique type instead, and since Ignition data management is expected to occur elsewhere, I don't think COS details like this are necessary. I will divert from the PowerVS path in respect to COS.

I will be removing this `COSInstanceReference` type as I don't expect to need it for VPC based on my findings.

Prajyot-Parab:
@Karthik-K-N ptal, lets merge if all okay

api/v1beta2/types.go
```diff
@@ -396,3 +396,17 @@ type VPCEndpoint struct {
 	// +optional
 	LBID *string `json:"loadBalancerIPID,omitempty"`
 }
+
+// VPCResource represents a specific VPC resource.
+// +kubebuilder:validation:XValidation:rule="!has(self.id) && !has(self.name)",message="an id or name must be provided"
```
cjschaef, line 401: This validation is written inversely, I will correct it.

cjschaef: fixed.

cjschaef:
Closing this PR in favor of more recent PR's for extended VPC support.

PR #1669, opened 2024-03-18
VPC: Add Security Group resources and constants/enums

cjschaef:
Add resource definitions for managing VPC Security Groups and Rules, along with a set of enumerations to simplify validation and options.

**What this PR does / why we need it**:  Adds support to define SecurityGroups in VPC for VPCCluster definitions.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes # N/A

**Special notes for your reviewer**:

/area provider/ibmcloud

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

PR #1683, opened 2024-03-25
VPC: VPCCluster definition extension

cjschaef:
Extend the VPCCluster definition to include additional VPC resources and expand configuration options.

**What this PR does / why we need it**: Expands the supported resource creation/management for VPC clusters

Fixes # N/A

/area provider/ibmcloud

**Release note**:

```release-note
Users will want to migrate to use the new definitions and configuration to take advantage of the extended support for VPCCluster infrastructure.
```


api/v1beta2/ibmvpccluster_types.go
```diff
@@ -32,28 +32,39 @@ const (
...
+	// +optional
+	ControlPlaneEndpoint capiv1beta1.APIEndpoint `json:"controlPlaneEndpoint"`
+
+	// ControlPlaneLoadBalancer is optional configuration for customizing control plane behavior.
+	// +optional
+	// dep: use LoadBalancers instead
+	ControlPlaneLoadBalancer *VPCLoadBalancerSpec `json:"controlPlaneLoadBalancer,omitempty"`
+
+	// cosInstance is the IBM COS instance to use for cluster resources.
+	COSInstance *COSInstance `json:"cosInstance,omitempty"`
```
cjschaef, line 47: If the shared basic types is added in time, this should be `COSInstanceReference`, from `types.go`.

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -103,25 +114,118 @@ type VPCLoadBalancerStatus struct {
...
+// TODO(cjschaef): Dummy COSInstance until it is defined in a common location.
+type COSInstance struct {
+	Name string `json:"name"`
+}
+
+// VPCResource dummy.
+// TODO(cjschaef): Dummy VPCResource until it is defined in a common location.
+type VPCResource struct {
+	Name string `json:"name"`
+}
```
cjschaef, line 152: I'll remove these dummy types if the other PR's adding them are merged before.
https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1670
https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1669

Prajyot-Parab:
/hold 
let the others PR be merged first, to avoid any confusion

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -103,25 +114,118 @@ type VPCLoadBalancerStatus struct {
...
+
+	// vpc references the IBM Cloud VPC.
+	// +optional
+	VPC *VPCResourceStatus `json:"vpc,omitempty"`
+}
+
+// VPCResourceStatus identifies a resource by crn and type and whether it was created by the controller.
+type VPCResourceStatus struct {
+	// controllerCreated indicates whether the resource is created by the CAPI controller.
+	ControllerCreated bool `json:"controllerCreated,omitempty"`
```
Karthik-K-N, line 220: Since most of VPC resources can be tagged, better to make use of tags instead of ControllerCreated. tags will be very easy comapared to managing state of this variable.

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -103,25 +114,118 @@ type VPCLoadBalancerStatus struct {
...
+
+	// crn defines the IBM Cloud CRN of the resource.
+	// +required
+	CRN string `json:"crn"`
+
+	// type defines the type of IBM Cloud resource.
+	// +required
+	Type ResourceType `json:"type"`
```
Karthik-K-N, line 228: I think its usage will be like 
````
VPC *VPCResourceStatus `json:"vpc,omitempty"
```

so do you think we need to again set VPC.Type =vpc

cjschaef: Yes I think that would be the simple case, I am concerned about more complex cases.

I will drop the type for now then, and attempt to add it back in if/when I need to when I start making changes to `reconcile` logic changes.

api/v1beta2/ibmvpccluster_types.go
cjschaef: I may need to think on this one more. I could move to rely on tags, but I know in the past we could not, or simply did not, in terms of performing resource management, perhaps a limitation with VPC resources at the time, or the technology/component used to create such resources.
Perhaps CAPI relies on tags information, if possible, and outside layers and logic transform that information as necessary. I may not know more on how to handle this until I start getting deeper into interdependencies, expectancies, etc.

Do you have a suggestion for this, or simply added a `Tags []string ...`? I didn't see a PowerVS equivalent, and I think the `Status` results will be necessary for post-Infrastructure-provisioning processing.

cjschaef: Or are you expecting that CAPI (or other components/automation) would look for specific Tags on resources (say a VSI with `CAPI-created` tag or something like that)?

api/v1beta2/ibmvpccluster_types.go
cjschaef: I dropped the Type for now.

api/v1beta2/ibmvpccluster_types.go
cjschaef: I dropped the bool for now.

api/v1beta2/ibmvpccluster_types.go
cjschaef: I'm dropping this reference, as I don't plan to manage COS resources.

Instead, I will incorporate a new `ImageSpec`, which will have details about the expected VSI Custom Image, which could include a COS ObjectURL, to create a new Custom Image from, if necessary.

api/v1beta2/ibmvpccluster_types.go
Karthik-K-N: Apologies some how I missed it. Yes what I am thinking is once the capi creates the resource in cloud say VPC, lets tag that resource with tag like CAPI-created or something, so it would help us in deciding on deleting that resource.

In PowerVS we cannot make use of tags as few services doesnot support tagging like DHCP, Network so on

mkumatag:
some of the changes are already merged part of other PRs recently, lets close this and push remaining ones in a different one.

PR #1707, opened 2024-04-03
Enhance security exposure patch

Prajyot-Parab:
- Enhance security exposure patch

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Enhance security exposure patch
```

### DIFF
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -142,7 +142,7 @@ prerequisites_vpc(){
     export PROVIDER_ID_FORMAT=v2
     export EXP_CLUSTER_RESOURCE_SET=true
     export IBMACCOUNT_ID=${IBMACCOUNT_ID:-"7cfbd5381a434af7a09289e795840d4e"}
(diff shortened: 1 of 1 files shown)


scripts/ci-e2e.sh
```diff
@@ -199,5 +199,9 @@ main(){
     [ -z "${BOSKOS_HOST:-}" ] || release_account >> "$ARTIFACTS/logs/boskos.log" 2>&1
 }
 
+if [[ $- =~ x ]]; then
```
Amulyam24, line 202: @Prajyot-Parab , has this been added as an additional check even after adding the snippet in boskos.sh?

Amulyam24 (approved):
/lgtm

scripts/ci-e2e.sh
Prajyot-Parab: Yes, to ensure we don't end up exposing anything in future even if boskos is removed/disabled. Also this applies to overall ci-e2e scope itself. (the other one was specifically for boskos)

PR #1710, opened 2024-04-03
Fix kubebuilder validation for SecurityGroup resources

cjschaef:
Some of the kubebuilder validation logic was written inversely and needs to be fixed so validation is done correctly.

**What this PR does / why we need it**: Some kubebuilder validation is written inversely, or is incorrect, causing issues with valid definitions.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Fix kubebuiler validation for new SecurityGroup related resources
```

### DIFF
--- a/api/v1beta2/types.go
+++ b/api/v1beta2/types.go
@@ -269,7 +269,7 @@ type PortRange struct {
 }
 
 // SecurityGroup defines a VPC Security Group that should exist or be created within the specified VPC, with the specified Security Group Rules.
-// +kubebuilder:validation:XValidation:rule="!has(self.id) && !has(self.name)",message="either an id or name must be specified"
+// +kubebuilder:validation:XValidation:rule="has(self.id) || has(self.name)",message="either an id or name must be specified"
 type SecurityGroup struct {
 	// id of the Security Group.
 	// +optional
@@ -298,8 +298,10 @@ type SecurityGroup struct {
 
 // SecurityGroupRule defines a VPC Security Group Rule for a specified Security Group.
 // +kubebuilder:validation:XValidation:rule="(has(self.destination) && !has(self.source)) || (!has(self.destination) && has(self.source))",message="both destination and source cannot be provided"
-// +kubebuilder:validation:XValidation:rule="has(self.destination) && self.direction == 'inbound'",message="destinationis not valid for SecurityGroupRuleDirectionInbound direction"
-// +kubebuilder:validation:XValidation:rule="has(self.source) && self.direction == 'outbound'",message="source is not valid for SecurityGroupRuleDirectionOutbound direction"
+// +kubebuilder:validation:XValidation:rule="self.direction == 'inbound' ? has(self.source) : true",message="source must be set for SecurityGroupRuleDirectionInbound direction"
+// +kubebuilder:validation:XValidation:rule="self.direction == 'inbound' ? !has(self.destination) : true",message="destination is not valid for SecurityGroupRuleDirectionInbound direction"
+// +kubebuilder:validation:XValidation:rule="self.direction == 'outbound' ? has(self.destination) : true",message="destination must be set for SecurityGroupRuleDirectionOutbound direction"
+// +kubebuilder:validation:XValidation:rule="self.direction == 'outbound' ? !has(self.source) : true",message="source is not valid for SecurityGroupRuleDirectionOutbound direction"
 type SecurityGroupRule struct {
 	// action defines whether to allow or deny traffic defined by the Security Group Rule.
 	// +required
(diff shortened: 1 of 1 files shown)


api/v1beta2/types.go
```diff
@@ -269,7 +269,7 @@ type PortRange struct {
 }
 
 // SecurityGroup defines a VPC Security Group that should exist or be created within the specified VPC, with the specified Security Group Rules.
-// +kubebuilder:validation:XValidation:rule="!has(self.id) && !has(self.name)",message="either an id or name must be specified"
+// +kubebuilder:validation:XValidation:rule="has(self.id) || has(self.name)",message="either an id or name must be specified"
```
Karthik-K-N, line 272: Could you please help me understand this, How it would affect, User should set either id or name right?

https://playcel.undistro.io/?content=H4sIAAAAAAAAA3VRS0sDMRD%2BK%2BN6aAu1xdeltyJFkapFkSLsZbo73Q0kkyXJti7W%2F%2B5ku6hVSg4Z8mW%2Bx8xHkmPAZJKcwh05AuUhlHJxVQeIkJTwNn2Yg3Vw%2F%2FL0CGvrDIZRyil7ymqnQnPrbF1NUgZQ%2BQTOLy6vrlNOhgm9V468V5ZFYDyGJenMGoJgW5Gb2RwWGptC2jk%2FSVm%2BHL5FO8hiIZDDLKgNCcdq6j2ZlW6gslvxnAPxRjnLhjhEbpHVVrKgEEhNTrXIVoVyr2uNsQyzb3cwRy5qLAj6oj8YRSetmzNYSj6CxtautfaTCEqR7j69emqJ0RGCQLH2Kqc4q1%2BjHB6ZZceyiMzQe66516bYoK7xiDgWqNiHP7vqiGZd%2FtiXWa1JJic9di0MaCpNvjPmK%2BUwYml3SvT9g6WOVD6A3Q7%2BA4yGBsnnFwKmXNFAAgAA

cjschaef: Yes, a name or an id must be provided.

Prajyot-Parab: @Karthik-K-N isn't this appropriate `!has(self.id) && !has(self.name)` ?

Karthik-K-N: No, Sometimes CEL expressions are confusing but the main logic is, The message will be displayed when the expression evaluates to false, In the above case if we set `!has(self.id) && !has(self.name)`, it does not allow us to set either one of id or name.
Example when we set id `!true && !false` --> `false && true` --> false = `either an id or name must be specified`

More info: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules

api/v1beta2/types.go
```diff
@@ -298,8 +298,7 @@ type SecurityGroup struct {
 
 // SecurityGroupRule defines a VPC Security Group Rule for a specified Security Group.
 // +kubebuilder:validation:XValidation:rule="(has(self.destination) && !has(self.source)) || (!has(self.destination) && has(self.source))",message="both destination and source cannot be provided"
-// +kubebuilder:validation:XValidation:rule="has(self.destination) && self.direction == 'inbound'",message="destinationis not valid for SecurityGroupRuleDirectionInbound direction"
-// +kubebuilder:validation:XValidation:rule="has(self.source) && self.direction == 'outbound'",message="source is not valid for SecurityGroupRuleDirectionOutbound direction"
+// +kubebuilder:validation:XValidation:rule="(self.direction == 'inbound' && !has(self.destination) && has(self.source)) || (self.direction == 'outbound' && has(self.destination) && !has(self.source))",message="destination is not valid for SecurityGroupRuleDirectionInbound direction and source is not valid for SecruityGroupRuleDirectionOutbound"
```
Karthik-K-N, line 301: I have spent some time in understanding validations, Here are my thoughts

The above expression may confuse with very generic error , example
```
  securityGroupRule:
    - action: "action"
      direction: "inbound"
      
spec.securityGroupRule[0]: Invalid value: "object": destination is not valid for SecurityGroupRuleDirectionInbound direction and source is not valid for SecruityGroupRuleDirectionOutbound
```

Instead of combining them into one line expression with combined error messages, It should be split to display appropriate validation failure messages, for example
```
// +kubebuilder:validation:XValidation:rule="self.direction == 'inbound' ? has(self.source) : true",message="source should be set for SecurityGroupRuleDirectionInbound direction"
// +kubebuilder:validation:XValidation:rule="self.direction == 'inbound' ? (!has(self.destination)) : true",message="destination is not valid for SecurityGroupRuleDirectionInbound direction"
// +kubebuilder:validation:XValidation:rule="self.direction == 'outbound' ? has(self.destination) : true",message="destination should be set for SecruityGroupRuleDirectionOutbound direction"
// +kubebuilder:validation:XValidation:rule="self.direction == 'outbound' ? (!has(self.source)) : true",message="source is not valid for SecruityGroupRuleDirectionOutbound direction"
```

Karthik-K-N (commented):
The expressions can be optimised to provide better error messages to user, But its not a blocking issue but an improvisation and good to have
Either we can add a TODO to fix it later or spend sometime to fix it now itself its upto you.

api/v1beta2/types.go
cjschaef: Sure I can split the validation logic up.

cjschaef: Updated, and performed some basic (non-exhaustive) validation.

Karthik-K-N: Thanks for updating

Karthik-K-N (commented):
/lgtm

Prajyot-Parab:
@cjschaef can you please update the release note section as well in this PR.

cjschaef:
Updated

Prajyot-Parab (approved):

PR #1812, opened 2024-05-28
Update log format while checking the state of a resource

Amulyam24:
This PR updates the log format while checking for the resource by printing the name or ID of the resource.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Update log format while checking the state of a resource
```

### DIFF
--- a/cloud/scope/powervs_cluster.go
+++ b/cloud/scope/powervs_cluster.go
@@ -680,7 +680,7 @@ func (s *PowerVSClusterScope) ReconcilePowerVSServiceInstance() (bool, error) {
 			return false, fmt.Errorf("failed to get PowerVS service instance with ID %s", serviceInstanceID)
 		}
 
-		requeue, err := s.checkServiceInstanceState(serviceInstance.State)
+		requeue, err := s.checkServiceInstanceState(*serviceInstance)
 		if err != nil {
 			return false, err
 		}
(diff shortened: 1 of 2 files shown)


cloud/scope/powervs_cluster.go
```diff
@@ -879,7 +879,7 @@ func (s *PowerVSClusterScope) isDHCPServerActive() (bool, error) {
 		return false, err
 	}
 
-	requeue, err := s.checkDHCPServerStatus(dhcpServer.Status)
+	requeue, err := s.checkDHCPServerStatus(*dhcpServer)
```
Karthik-K-N, line 882: How about logging the name here before calling the function? So we can avoid passing whole object.

cloud/scope/powervs_cluster.go
```diff
@@ -2550,7 +2553,7 @@ func (s *PowerVSClusterScope) DeleteServiceInstance() (bool, error) {
 		return false, err
 	}
 	s.Info("PowerVS service instance successfully deleted")
-	return false, nil
+	return true, nil
```
Karthik-K-N, line 2556: Why to re-queue in case of successful delete?

Amulyam24: As per the logic, we are requeuing after successful delete of a resource and wait for it to get deleted before checking again in the next reconcile to confirm if it has been removed successfully.

Karthik-K-N (commented):
/lgtm

Prajyot-Parab (approved):

PR #1816, opened 2024-05-29
Fetch service instance id from spec of IBMPowerVSCluster object while setting provider id

Karthik-K-N:

When we are using external [cloud-provider template](https://cluster-api-ibmcloud.sigs.k8s.io/topics/powervs/external-cloud-provider) we explicitly pass the service instance id.
Internally in the code that wont be set under status field of IBMPowerVSCluster, so the in the [provider id of machine](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/e49f8ee8a7bfb2d627461f2b1e415eedbee09297/cloud/scope/powervs_machine.go#L942) the service instance will be set as empty string which is incorrect.

In this PR we try to fetch the service instance id from spec of object if its not present under the status.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Fetch Service instance id from spec field of IBMPowerVSCluster object if its not present under status.
```

### DIFF
--- a/cloud/scope/powervs_machine.go
+++ b/cloud/scope/powervs_machine.go
@@ -922,10 +922,16 @@ func (m *PowerVSMachineScope) GetZone() string {
 
 // GetServiceInstanceID returns the service instance id.
 func (m *PowerVSMachineScope) GetServiceInstanceID() string {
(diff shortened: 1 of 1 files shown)


Amulyam24 (approved):
/lgtm

Prajyot-Parab (approved):

PR #1825, opened 2024-06-04
Improve IBMPowerVSCluster deletion

### LINKED ISSUE #1805: Enhance deletion logic
/kind feature
/area provider/ibmcloud

We are currently fetching the resource and checking against a [text phrase](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/cloud/scope/types.go#L20-L42) if it is successfully deleted(eg: [VPC](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/cloud/scope/powervs_cluster.go#L2390-L2400)), leverage using http status codes instead.
This is applicable to

- [ ] VPC resources(VPC, load balancer, subnet, security groups)
- [ ] Transit gateway

dharaneeshvrd:

Fixes #1805 

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/cloud/scope/powervs_cluster.go
+++ b/cloud/scope/powervs_cluster.go
@@ -2285,12 +2285,12 @@ func (s *PowerVSClusterScope) DeleteLoadBalancer() (bool, error) {
 			continue
 		}
 
-		lb, _, err := s.IBMVPCClient.GetLoadBalancer(&vpcv1.GetLoadBalancerOptions{
+		lb, resp, err := s.IBMVPCClient.GetLoadBalancer(&vpcv1.GetLoadBalancerOptions{
 			ID: lb.ID,
 		})
 
 		if err != nil {
-			if strings.Contains(err.Error(), string(VPCLoadBalancerNotFound)) {
+			if resp != nil && resp.StatusCode == ResourceNotFoundCode {
 				s.Info("VPC load balancer successfully deleted")
 				continue
 			}
@@ -2324,10 +2324,10 @@ func (s *PowerVSClusterScope) DeleteVPCSecurityGroups() error {
 			s.Info("Skipping VPC security group deletion as resource is not created by controller", "ID", *securityGroup.ID)
 			continue
 		}
-		if _, _, err := s.IBMVPCClient.GetSecurityGroup(&vpcv1.GetSecurityGroupOptions{
+		if _, resp, err := s.IBMVPCClient.GetSecurityGroup(&vpcv1.GetSecurityGroupOptions{
 			ID: securityGroup.ID,
 		}); err != nil {
-			if strings.Contains(err.Error(), string(VPCSecurityGroupNotFound)) {
+			if resp != nil && resp.StatusCode == ResourceNotFoundCode {
 				s.Info("VPC security group has been already deleted", "ID", *securityGroup.ID)
 				continue
 			}
@@ -2356,12 +2356,12 @@ func (s *PowerVSClusterScope) DeleteVPCSubnet() (bool, error) {
 			continue
 		}
 
-		net, _, err := s.IBMVPCClient.GetSubnet(&vpcv1.GetSubnetOptions{
+		net, resp, err := s.IBMVPCClient.GetSubnet(&vpcv1.GetSubnetOptions{
 			ID: subnet.ID,
 		})
 
 		if err != nil {
-			if strings.Contains(err.Error(), string(VPCSubnetNotFound)) {
+			if resp != nil && resp.StatusCode == ResourceNotFoundCode {
 				s.Info("VPC subnet successfully deleted")
 				continue
 			}
@@ -2398,12 +2398,12 @@ func (s *PowerVSClusterScope) DeleteVPC() (bool, error) {
 		return false, nil
 	}
 
-	vpc, _, err := s.IBMVPCClient.GetVPC(&vpcv1.GetVPCOptions{
+	vpc, resp, err := s.IBMVPCClient.GetVPC(&vpcv1.GetVPCOptions{
 		ID: s.IBMPowerVSCluster.Status.VPC.ID,
 	})
 
 	if err != nil {
-		if strings.Contains(err.Error(), string(VPCNotFound)) {
+		if resp != nil && resp.StatusCode == ResourceNotFoundCode {
 			s.Info("VPC successfully deleted")
 			return false, nil
 		}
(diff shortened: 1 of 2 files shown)


controllers/ibmpowervscluster_controller.go
```diff
@@ -311,7 +311,8 @@ func (r *IBMPowerVSClusterReconciler) reconcileDelete(ctx context.Context, clust
 
 	clusterScope.Info("Deleting DHCP server")
 	if err := clusterScope.DeleteDHCPServer(); err != nil {
-		allErrs = append(allErrs, errors.Wrapf(err, "failed to delete DHCP server"))
+		clusterScope.Error(err, "failed to delete DHCP server")
+		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
```
Amulyam24, line 315: @dharaneeshvrd, if the DHCP server deletion constantly fails, won't this keep requeuing the request and not proceed with PowerVS service instance deletion?

dharaneeshvrd: Yes, currently also same thing will happen with this block https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/cloud/scope/powervs_cluster.go#L2543toL2554
We need to delete before proceeding to the powervs instance right? or we can try deleting powervs instance after few retries anyway?

Amulyam24: > Yes, currently also same thing will happen with this block https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/cloud/scope/powervs_cluster.go#L2543toL2554 

Yes, currently it's doing the same. I'm trying to understand how this change will help

> we can try deleting powervs instance after few retries anyway?

right, I feel we need to proceed with PowerVS service instance deletion despite failure with DHCP, wdyt?

Amulyam24: I have not tested the scenario but in case of DHCP failure, will the service instance deletion also get stuck?

dharaneeshvrd: Yes if the dhcp server did not get deleted, I don't think powervs instance will get deleted successfully.

Karthik-K-N: So is workspace deletion trigger DHCP server deletion? if workspace deletion internally triggers DHCP server deletion then it expects the DHCP server deletion to be successful  ,in case of error it will block workspace deletion as well. 
Is my understanding correct?

dharaneeshvrd: We are not sure about this, tried to reproduce this issue in past week, somehow did nt reproduce at all. But I believe we can try the powervs workspace deletion and retry it until it succeeds like Amulya mentioned. No point in blocking the powervs workspace deletion when dhcp server deletion fails.

dharaneeshvrd: Btw, in hypershift we were nt trying to delete the dhcp server separately we will only delete only the powervs workspace and retry the deletion until it succeeds.

Karthik-K-N: Got it, Thanks. @Amulyam24 was there any particular reason for explicitly deleting DHCP server in CAPI?

Amulyam24: In the initial flow before adding proper state management of resources, we added the DHCP deletion and check to make sure we PowerVS workspace deletion succeeds. We can consider remove it now and proceed with workspace deletion and retry until it is deleted. We need to test it thoroughly to make sure it is cleaned up. @dharaneeshvrd can you please try this out and let us know?

dharaneeshvrd: Sure will try and see!

dharaneeshvrd: Tried many times in different region, the deletion went through without explicitly deleting the DHCP server.
Modified the code accordingly, please check!

Karthik-K-N: Nice, Thanks for verifying.

cloud/scope/powervs_cluster.go
```diff
@@ -2492,32 +2490,6 @@ func (s *PowerVSClusterScope) deleteTransitGatewayConnections(tg *tgapiv1.Transi
 	return requeue, nil
 }
 
-// DeleteDHCPServer deletes DHCP server.
-func (s *PowerVSClusterScope) DeleteDHCPServer() error {
-	if !s.isResourceCreatedByController(infrav1beta2.ResourceTypeDHCPServer) {
```
Karthik-K-N, line 2497: Do you think there will be scenario where user provides existing workspace and expects the DHCP server to be created? If so in that case we need to delete only DHCP server.

dharaneeshvrd: True, we need to handle that. Will do the changes.

dharaneeshvrd: Addressed this, ptal!

cloud/scope/powervs_cluster.go
```diff
@@ -2498,6 +2498,10 @@ func (s *PowerVSClusterScope) DeleteDHCPServer() error {
 		s.Info("Skipping DHP server deletion as resource is not created by controller")
 		return nil
 	}
+	if s.isResourceCreatedByController(infrav1beta2.ResourceTypeServiceInstance) {
+		s.Info("Skipping DHCP server deletion as PowerVS service instance is created by controller, will directly delete the PowerVS service instance")
```
Karthik-K-N, line 2502: I think user may not know PowerVS service instance delete will also delete DHCP server and may get confused after seeing work Skipping, I think its better to either use V(3) or update the log message.

cloud/scope/powervs_cluster.go
```diff
@@ -2356,12 +2356,12 @@ func (s *PowerVSClusterScope) DeleteVPCSubnet() (bool, error) {
...
 		}
 
-		net, _, err := s.IBMVPCClient.GetSubnet(&vpcv1.GetSubnetOptions{
+		net, resp, err := s.IBMVPCClient.GetSubnet(&vpcv1.GetSubnetOptions{
 			ID: subnet.ID,
 		})
 
 		if err != nil {
-			if strings.Contains(err.Error(), string(VPCSubnetNotFound)) {
+			if resp.StatusCode == ResourceNotFoundCode {
```
Karthik-K-N, line 2364: So is it safe to assume that resp wont be nil?

cloud/scope/powervs_cluster.go
dharaneeshvrd: Updated the log, let's have this log IMO, since its skipping a major action in delete flow.

cloud/scope/powervs_cluster.go
dharaneeshvrd: It seems there is a possibility for it to be nil, added a nil check.

Karthik-K-N (commented):
/lgtm

dharaneeshvrd:
@Karthik-K-N @Amulyam24 addressed the comments, please take a look!

Amulyam24 (approved):
/lgtm

PR #1858, opened 2024-06-26
VPC: Create v2 path for new Infrastructure implementation

cjschaef:
Create a v2 path that will be used for the new Infrastructure implementation for VPC Clusters. All new functionality will be placed in these new v2 paths, based on the new NetworkSpec field, to prevent breaking existing implementation.

**What this PR does / why we need it**:  adds new path to fully implement VPC Cluster Infrastructure support w/o breaking current functionality

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
adding new VPC Infrastructure pathways
```

### DIFF
--- a/cloud/scope/vpc_cluster.go
+++ b/cloud/scope/vpc_cluster.go
@@ -1,0 +1,168 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scope
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/go-logr/logr"
+
+	"github.com/IBM/go-sdk-core/v5/core"
+	"github.com/IBM/platform-services-go-sdk/resourcecontrollerv2"
+
+	"k8s.io/klog/v2/textlogger"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	capiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/patch"
+
+	infrav1beta2 "sigs.k8s.io/cluster-api-provider-ibmcloud/api/v1beta2"
+	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/cloud/services/authenticator"
+	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/cloud/services/cos"
+	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/cloud/services/resourcecontroller"
+	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/cloud/services/resourcemanager"
+	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/cloud/services/vpc"
+	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/endpoints"
+)
+
+const (
+	// LOGDEBUGLEVEL indicates the debug level of the logs.
+	LOGDEBUGLEVEL = 5
+)
+
+// VPCClusterScopeParams defines the input parameters used to create a new VPCClusterScope.
+type VPCClusterScopeParams struct {
+	Client          client.Client
+	Cluster         *capiv1beta1.Cluster
+	IBMVPCCluster   *infrav1beta2.IBMVPCCluster
+	Logger          logr.Logger
+	ServiceEndpoint []endpoints.ServiceEndpoint
+
+	IBMVPCClient vpc.Vpc
+}
+
+// VPCClusterScope defines a scope defined around a VPC Cluster.
+type VPCClusterScope struct {
+	logr.Logger
+	Client      client.Client
+	patchHelper *patch.Helper
+
+	COSClient                cos.Cos
+	ResourceControllerClient resourcecontroller.ResourceController
+	ResourceManagerClient    resourcemanager.ResourceManager
+	VPCClient                vpc.Vpc
+
+	Cluster         *capiv1beta1.Cluster
+	IBMVPCCluster   *infrav1beta2.IBMVPCCluster
+	ServiceEndpoint []endpoints.ServiceEndpoint
+}
+
+// NewVPCClusterScope creates a new VPCClusterScope from the supplied parameters.
+func NewVPCClusterScope(params VPCClusterScopeParams) (*VPCClusterScope, error) {
+	if params.Client == nil {
+		err := errors.New("error failed to generate new scope from nil Client")
+		return nil, err
+	}
+	if params.Cluster == nil {
+		err := errors.New("error failed to generate new scope from nil Cluster")
+		return nil, err
+	}
+	if params.IBMVPCCluster == nil {
+		err := errors.New("error failed to generate new scope from nil IBMVPCCluster")
+		return nil, err
+	}
+	if params.Logger == (logr.Logger{}) {
+		params.Logger = textlogger.NewLogger(textlogger.NewConfig())
+	}
+
+	helper, err := patch.NewHelper(params.IBMVPCCluster, params.Client)
+	if err != nil {
+		return nil, fmt.Errorf("error failed to init patch helper: %w", err)
+	}
+
+	vpcEndpoint := endpoints.FetchVPCEndpoint(params.IBMVPCCluster.Spec.Region, params.ServiceEndpoint)
+	vpcClient, err := vpc.NewService(vpcEndpoint)
+	if err != nil {
+		return nil, fmt.Errorf("error failed to create IBM VPC client: %w", err)
+	}
+
+	if params.IBMVPCCluster.Spec.Network == nil || params.IBMVPCCluster.Spec.Region == "" {
+		return nil, fmt.Errorf("error failed to generate vpc client as Network or Region is nil")
+	}
+
+	if params.Logger.V(LOGDEBUGLEVEL).Enabled() {
+		core.SetLoggingLevel(core.LevelDebug)
+	}
+
+	auth, err := authenticator.GetAuthenticator()
+	if err != nil {
+		return nil, fmt.Errorf("error failed to create authenticator: %w", err)
+	}
+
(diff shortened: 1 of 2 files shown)


api/v1beta2/conditions_consts.go
```diff
@@ -120,4 +120,8 @@ const (
 	// CreateInfrastructureAnnotation is the name of an annotation that indicates if
 	// Power VS infrastructure should be created as a part of cluster creation.
 	CreateInfrastructureAnnotation = "powervs.cluster.x-k8s.io/create-infra"
+
+	// CreateVPCInfrastructureAnnotation is the name of an annotation that indicates if
+	// VPC infrastructure should be created as part of cluster creation.
+	CreateVPCInfrastructureAnnotation = "vpc.cluster.x-k8s.io/create-infra"
```
Karthik-K-N, line 126: @cjschaef what will be used to differentiate v2 vs v1 code path. I think in VPC we create infra by default so this annotation might not be required. Are you planning to use this as a differentiator?

cjschaef: Yes, I plan on using that and flagging off the NetworkSpec field in the Cluster resource for the v2 path.

Karthik-K-N: ok, If I understand correctly, in V2 path you are expecting `NetworkSpec` to be mandatory?

cjschaef: That is correct. `NetworkSpec` is part of the new Infrastructure deployment support, and is required for that support.

Karthik-K-N: Got it thanks for the clarification, 
May be we can get started with using that annotation if we can't find any alternative suitable name which matches for what we are doing but eventually we need to remove it as it may confuse the users because now also by default we create the resource.

cjschaef: I could remove the annotation altogether, if we wish to start the so-called "v2" VPC support without requiring an annotation (rely only on using `NetworkSpec` to flag the new (v2) VPC support, creating all of the required Infrastructure.

I had expected that annotation was required in order to get CAPI to perform Infrastructure Creation/Management. But if we expect for our new (v2) support, CAPI will always be attempting to create/manage Infrastructure resources, I can drop that annotation completely.

cjschaef: Another question, can CAPI be run within a cluster with only the Machine Controller running (not the Cluster/Infra Controller), so as to not attempt to reprovision Infrastructure post cluster creation/install? I had expected this annotation would help with that situation as well, to prevent additional churn by CAPI by continuing to monitor (or deleting) Infrastructure post cluster create.

Karthik-K-N: In PowerVS v1 flow we were not creating any infra rather using existing infra and expecting user to provide them. We introduced this annotation in for v2 to make sure we need to create infra and not expecting anything from user.

This is not true for VPC, in vpc v1 flow as well we create required resources so this annotation does imply anything. But for you to differentiate V2 path as you suggested its better to go with NetworkSpec. If user set NetworkSpec you can assume its V2 flow.

Karthik-K-N: >Another question, can CAPI be run within a cluster with only the Machine Controller running (not the Cluster/Infra Controller), so as to not attempt to reprovision Infrastructure post cluster creation/install? I had expected this annotation would help with that situation as well, to prevent additional churn by CAPI by continuing to monitor (or deleting) Infrastructure post cluster create.

No this annotation is nothing to do with that, But to achieve what you are expecting capi provides annotation named `[cluster.x-k8s.io/managed-by](https://github.com/Karthik-K-N/cluster-api/blob/e2bf2b7d108ecd51bf7b23bf50e2432e82d8f2fe/api/v1beta1/common_types.go#L154-L160)` and our controllers are configured to use them [here](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/01e1771ca8cbf8b9969bb64b6956b3d470219313/controllers/ibmpowervscluster_controller.go#L477).

cjschaef: Okay, thanks for the details, I will drop the annotation. and rely only on the NetworkSpec.

cjschaef:
Re-pushing to try to get CLA check to pass (no idea why it failed all of a sudden).

cloud/scope/vpc_cluster.go
```diff
@@ -0,0 +1,168 @@
...
+	ResourceManagerClient    resourcemanager.ResourceManager
+	VPCClient                vpc.Vpc
+
+	Cluster         *capiv1beta1.Cluster
+	IBMVPCCluster   *infrav1beta2.IBMVPCCluster
+	ServiceEndpoint []endpoints.ServiceEndpoint
+}
+
+// NewVPCClusterScope creates a new VPCClusterScope from the supplied parameters.
+func NewVPCClusterScope(params VPCClusterScopeParams) (*VPCClusterScope, error) {
```
Karthik-K-N, line 78: I am still not clear on how this will be invoked and embedded into existing flow, Do you mind adding that snippet of code to this PR. Also if possible we can move api definition to seperate PR if possible.

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -94,6 +98,19 @@ type AdditionalListenerSpec struct {
...
+type VPCNetworkSpec struct {
+	// computeSubnetsSpec is a set of Subnet's which define the Compute subnets.
+	ComputeSubnetsSpec []Subnet `json:"computeSubnetsSpec,omitempty"`
+	// controlPlaneSubnetsSpec is a set of Subnet's which define the Control Plane subnets.
+	ControlPlaneSubnetsSpec []Subnet `json:"controlPlaneSubentsSpec,omitempty"`
+	// resourceGroup is the name of the Resource Group containing all of the newtork resources.
+	// This can be different than the Resource Group containing the remaining cluster resources.
+	ResourceGroup *string `json:"resourceGroup,omitempty"`
+
+	// TODO(cjschaef): Complete spec definition (SecurityGroups, VPC)
```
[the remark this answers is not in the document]
cjschaef: Yes, I plan to add additional details in more functional PR's.

cloud/scope/vpc_cluster.go
cjschaef: It gets utilized here, in this new v2 path, which allows me to modify the `Cluster` resource without affecting the existing VPC (legacy) support:
https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1858/files#diff-647a58f1c11f771d3ec32c60c28c4b8b29442f9915129e16eeddb59120dcc7d8R140-R146

I can attempt to break up this PR, and maybe see if adding anything more for the new VPCClusterScope use that doesn't balloon the PR too much.

Karthik-K-N: >It gets utilized here, in this new v2 path, which allows me to modify the Cluster resource without affecting the existing VPC (legacy) support:
I got that but where do you call reconcilerv2?, Something like you need to add a check to see whether vpcclusterspec.NetworkSpec is set or not, If not set go with v1 flow and if set go with new v2 flow?


If this is the case, now if you are expecting user to go with v2 flow say for IPI case but user misses to set networkspec now the controller by default with v1 flow which is not you are expecting and we cannot make networkspec mandatory also, So we dont have any way to add a validation for v2 flow. Its up-to user intelligence and documentation to go with v2 flow..

cjschaef: Yes, that is my expectation, the burden is on the user to provide details in order to utilize the v2 flow.
The architecture deployment for the v2 flow is going to be drastically different than the existing deployment.
In order to achieve and configure this, the user will be on the hook to configure the Cluster resource properly.

cjschaef: >I got that but where do you call reconcilerv2?, Something like you need to add a check to see whether vpcclusterspec.NetworkSpec is set or not, If not set go with v1 flow and if set go with new v2 flow?

That is right here
https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1858/files#diff-647a58f1c11f771d3ec32c60c28c4b8b29442f9915129e16eeddb59120dcc7d8R74-R77

Otherwise, it continues with the legacy (v1) flow

Karthik-K-N: got it thanks

Karthik-K-N (commented):
overall LGTM, may be we can merge the api pr then merge this,

Also I see lots of code duplication can be avoided like in NewClusterScope, May be once we have clear picture of v2 flow, we can combine whereever possible.

Thanks.

cjschaef:
Rebased off API updates:  #1875 

Yes, I imagine some refactoring could be done to handle duplicate code, but the intention is to avoid interacting/modifying the existing VPC flow to simplify testing and support (I am not familiar with the expectations with the current VPC flow). Depending on whether we expect to deprecate and remove the current flow, in favor of the extended VPC support (v2).

controllers/ibmvpccluster_controller.go
```diff
@@ -71,6 +71,11 @@ func (r *IBMVPCClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, err
 	}
 
+	// Determine whether the Cluster is designed for extended Infrastructure support, implemented in a separate path.
+	if ibmCluster.Spec.Network != nil {
```
mkumatag, line 75: this code block can above line 64 to avoid one extra Get call..

cloud/scope/vpc_cluster.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	vpcEndpoint := endpoints.FetchVPCEndpoint(params.IBMVPCCluster.Spec.Region, params.ServiceEndpoint)
+	vpcClient, err := vpc.NewService(vpcEndpoint)
+	if err != nil {
+		return nil, fmt.Errorf("error failed to create IBM VPC client: %w", err)
+	}
+
+	if params.IBMVPCCluster.Spec.Network == nil || params.IBMVPCCluster.Spec.Region == "" {
+		return nil, fmt.Errorf("error failed to generate vpc client as NetworkSpec info is nil")
```
mkumatag, line 107: if statement is checking for Region but it is not reported in the error, this may mislead the user..

mkumatag (changes requested):

cloud/scope/vpc_cluster.go
cjschaef: updated error info

controllers/ibmvpccluster_controller.go
cjschaef: I think this needs to be here, after the `ibmCluster` variable is defined and then populated via `r.Get` from the Cluster resource on line 65-66.

mkumatag: yep, my bad..

cjschaef:
Repushing to get CLA to pass.

mkumatag (approved):
/lgtm

PR #1869, opened 2024-07-03
Refactor infra creation to improve the overall infra setup time

### LINKED ISSUE #1837: Experiment to rearrange resource creation invocation in recocile loop to find best minimal time for overall infra resource creation
/kind feature
/area provider/ibmcloud

Currently as a part of [reconcile](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/395cb8e75e4c180c2d9ff673cb45ce9357828033/controllers/ibmpowervscluster_controller.go#L115) while creating infra resources we create loadbalancer as the last resources, Its observed that loadbalancer takes lot more time to become active state. 

We want to try and figure out best possible order for infra resource creation so that we can create all resources in minimal time.
example Create all vpc resources prior to power vs resources so on.

dharaneeshvrd:

Fixes #1837 

With the changes, infra creation takes around ~8 mins.
Most time taking resource is DHCP server VM.

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/controllers/ibmpowervscluster_controller.go
+++ b/controllers/ibmpowervscluster_controller.go
@@ -20,10 +20,12 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
 
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


controllers/ibmpowervscluster_controller.go
```diff
@@ -111,142 +113,228 @@ func (r *IBMPowerVSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-		return reconcile.Result{}, err
-	}
+	conditions.MarkFalse(update.cluster, conditionArgs[0].(capiv1beta1.ConditionType), conditionArgs[1].(string), conditionArgs[2].(capiv1beta1.ConditionSeverity), conditionArgs[3].(string), conditionArgs[4:]...)
+}
 
-	powerVSCluster := clusterScope.IBMPowerVSCluster
+func (r *IBMPowerVSClusterReconciler) reconcilePowerVSResources(clusterScope *scope.PowerVSClusterScope, updatePowerVSCluster *updatePowerVSCluster, ch chan reconcileResult, wg *sync.WaitGroup) {
+	defer wg.Done()
 	// reconcile PowerVS service instance
 	clusterScope.Info("Reconciling PowerVS service instance")
```
Karthik-K-N, line 140: since PowerVS and VPC resource creations will happen concurrently, we can prefix log with some specific keys so we can trace the log of individual flow, something like
vpcLog := clusterScope.WithName("vpc")
vpcLog.Info("Creating vpc) , I hope it will work

controllers/ibmpowervscluster_controller.go
```diff
@@ -111,142 +113,228 @@ func (r *IBMPowerVSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-			conditions.MarkFalse(powerVSCluster, infrav1beta2.COSInstanceReadyCondition, infrav1beta2.COSInstanceReconciliationFailedReason, capiv1beta1.ConditionSeverityError, err.Error())
+			conditions.MarkFalse(updatePowerVSCluster.cluster, infrav1beta2.COSInstanceReadyCondition, infrav1beta2.COSInstanceReconciliationFailedReason, capiv1beta1.ConditionSeverityError, err.Error())
 			return reconcile.Result{}, err
 		}
-		conditions.MarkTrue(powerVSCluster, infrav1beta2.COSInstanceReadyCondition)
+		conditions.MarkTrue(updatePowerVSCluster.cluster, infrav1beta2.COSInstanceReadyCondition)
+	}
+
+	var networkReady, loadBalancerReady bool
+	for _, cond := range clusterScope.IBMPowerVSCluster.Status.Conditions {
```
Karthik-K-N, line 312: By this time network should be ready right, When do you think it may not be ready also if its not ready we cannot even attach with TransitGateway right?

Karthik-K-N (commented):
Thank you for this PR, Overall approach looks good to me, Whats the time difference between old vs new approach?

controllers/ibmpowervscluster_controller.go
Amulyam24: should we consider moving this check before calling `ReconcileTransitGateway` to avoid failure?

dharaneeshvrd: With transit gateway we are directly attaching the powervs with CRN right? It is not dependent on the network. And I m not requeuing when network and load balancer is not ready since it would block the TG creation even though it is not dependant for the TG. But before proceeding to retrieve the VPC LB details and set the overall Ready status as true, wanted to validate their status has reached Ready, that's why added the condition to validate both of these resource's Ready status.

dharaneeshvrd:
@Karthik-K-N 
> Whats the time difference between old vs new approach?
Currently it's taking around 15 mins and with these changes it is taking around 8 to 9 mins.

mkumatag:
@dharaneeshvrd as per the discussion, lets get the early testing done with this PR to ensure it is working fine

mkumatag:
> @dharaneeshvrd as per the discussion, lets get the early testing done with this PR to ensure it is working fine

Any update on this?

/cc @Karthik-K-N 

please take a look and give lgtm if no more comments

Karthik-K-N (commented):
One last question, Otherwise overall LGTM

controllers/ibmpowervscluster_controller.go
```diff
@@ -111,6 +113,114 @@ func (r *IBMPowerVSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+
+func (update *updatePowerVSCluster) condition(condition bool, conditionArgs ...interface{}) {
+	update.mu.Lock()
+	defer update.mu.Unlock()
+	if condition {
+		conditions.MarkTrue(update.cluster, conditionArgs[0].(capiv1beta1.ConditionType))
+		return
+	}
+
+	conditions.MarkFalse(update.cluster, conditionArgs[0].(capiv1beta1.ConditionType), conditionArgs[1].(string), conditionArgs[2].(capiv1beta1.ConditionSeverity), conditionArgs[3].(string), conditionArgs[4:]...)
```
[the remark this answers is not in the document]
dharaneeshvrd: If I pass it as a straight types, we would get a golint error for exceeding the max number of args which is 5. Also I would need to create a separate func for true and false condition. It would complicate things IMO.

>Should we add a check for verifying the min length to avoid index out of range?

Again, need to add a check on many places where we are updating the condition and won't be able to take major action if it errored. Ex [here](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1869/files#diff-11c67fcffb6910d7d88d6c399511174f09fbd5098e22e3aa37ee4f14f2302a8cR176) I want to return the main error instead of the error I m getting from this, it does n't look good to me. Scope of this is within the reconcile func and only two possible flow. I think the current way is ok IMO.

dharaneeshvrd:
@mkumatag 

>Any update on this?

I have tested personally and Ashwin Hendre also able to test this and reported the improvement of 12% overall cluster setup time.

Karthik-K-N (commented):
/lgtm

controllers/ibmpowervscluster_controller.go
```diff
@@ -111,6 +113,114 @@ func (r *IBMPowerVSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	clusterScope.IBMPowerVSClient.WithClients(powervs.ServiceOptions{CloudInstanceID: clusterScope.GetServiceInstanceID()})
+
+	// reconcile network
+	powerVSLog.Info("Reconciling network")
+	if requeue, err := clusterScope.ReconcileNetwork(); err != nil {
+		powerVSLog.Error(err, "failed to reconcile PowerVS network")
+		powerVSCluster.updateCondition(false, infrav1beta2.NetworkReadyCondition, infrav1beta2.NetworkReconciliationFailedReason, capiv1beta1.ConditionSeverityError, err.Error())
+		ch <- reconcileResult{reconcile.Result{}, err}
+		return
+	} else if requeue {
```
mkumatag, line 163: we aren't doing any `requeue` here, hence it will be good if we can rename or rewrite the logic to have meaningful name here

mkumatag: also please add why there is no requeue here?

mkumatag (changes requested):
few minor nits, otherwise lgtm

controllers/ibmpowervscluster_controller.go
dharaneeshvrd: I have refactored the code, please take a look!

mkumatag (approved):
/lgtm

PR #1883, opened 2024-07-15
VPC: Extend VPC related API's for v2 VPC reconcile

cjschaef:
Extend the API's used in the v2 support of reconciling VPC for the VPCCluster support.

**What this PR does / why we need it**:  implements the v2 support of reconciling a VPC for VPCCluster support

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
implement new VPC reconcile for VPCClusters
```


api/v1beta2/types.go
```diff
@@ -400,3 +400,53 @@ type VPCEndpoint struct {
 	// +optional
 	LBID *string `json:"loadBalancerIPID,omitempty"`
 }
+
+// GenericResourceReference represents a generic IBM Cloud resource.
+type GenericResourceReference struct {
```
mkumatag, line 405: I don't see this is used anywhere!

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -163,6 +178,45 @@ type IBMVPCClusterStatus struct {
...
+	// +optional
+	ResourceGroup *ResourceStatus `json:"resourceGroup,omitempty"`
+
+	// vpc references the IBM Cloud VPC.
+	// +optional
+	VPC *VPCResourceStatus `json:"vpc,omitempty"`
+}
+
+// VPCResourceStatus identifies a VPC resource by id (and name) and whether it is ready.
+type VPCResourceStatus struct {
```
mkumatag, line 194: `ResourceStatus` also look the same, can we reuse?

mkumatag (changes requested):

api/v1beta2/ibmvpccluster_types.go
cjschaef: I had designed the two status' initially, but can not recall why I had a unique one for VPC resources ATM.
I'll remove and use `ResourceStatus` for all resources instead.

api/v1beta2/types.go
cjschaef: Sorry, too early for that resource, I'll remove it for this PR.

cjschaef: Removed

api/v1beta2/ibmvpccluster_types.go
cjschaef: Removed to use `ResourceStatus`

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -113,7 +113,11 @@ type VPCNetworkSpec struct {
 	// +optional
 	ResourceGroup *string `json:"resourceGroup,omitempty"`
 
-	// TODO(cjschaef): Complete spec definition (SecurityGroups, VPC)
+	// vpc defines the IBM Cloud VPC.
+	// +optional
```
Karthik-K-N, line 117: We already have a field named VPC under IBMVPCClusterSpec, so may be lets add a comment whereever it looks appropriate on whats the difference or when to use which field.

Karthik-K-N (commented):
I think its also ok to have one big PR with various small commits, so it gives overall good picture of intent of changes.

api/v1beta2/ibmvpccluster_types.go
cjschaef: I added a little more detail to this field and some other fields/structs, as they related to the extended VPC Infra support.

cjschaef:
Repushing to get CLA to pass.

mkumatag (approved):
/lgtm

Issue #1868, opened 2024-07-02
Deprecate provider-id format v1


Amulyam24:
/kind feature
/area provider/ibmcloud

**Describe the solution you'd like**
When deploying the cluster without cloud provider, it was discovered that the first control plane which is provisioned fails certain health checks(connecting to pods on control plane node failing). It was due to the nodes not having IP addresses assigned which was introduced due to the change in https://github.com/kubernetes/kubernetes/pull/121028 from k8s v 1.29

As a result, we need to switch to using cloud provider template by default when deploying PowerVS clusters

Hence, we need to deprecate the provider-id format v1 and remove [cluster-template-powervs.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/templates/cluster-template-powervs.yaml) which will not be used going forward.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

PR #1897, opened 2024-07-30
Deprecate provider-id format v1

KeerthanaAP:
Deprecate provider-id format v1

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/1868

/area provider/ibmcloud

**Release note**:

```release-note
Deprecate provider-id format v1
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -103,11 +103,10 @@ func initFlags(fs *pflag.FlagSet) {
 		10*time.Minute,
 		"The minimum interval at which watched resources are reconciled.",
 	)
-
 	fs.StringVar(
 		&options.ProviderIDFormat,
 		"provider-id-fmt",
-		string(options.ProviderIDFormatV1),
+		string(options.ProviderIDFormatV2),
 		"ProviderID format is used set the Provider ID format for Machine",
 	)
 
@@ -132,8 +131,9 @@ func initFlags(fs *pflag.FlagSet) {
 
 func validateFlags() error {
 	switch options.ProviderIDFormatType(options.ProviderIDFormat) {
+	// Deprecated: ProviderIDFormatV1 is deprecated and will be removed in a future release.
 	case options.ProviderIDFormatV1:
-		setupLog.Info("Using v1 version of ProviderID format")
+		setupLog.Info("Using v1 version of ProviderID format.V1 is deprecated and will be removed in a future release.Instead use V2.")
 	case options.ProviderIDFormatV2:
 		setupLog.Info("Using v2 version of ProviderID format")
 	default:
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -103,11 +103,10 @@ func initFlags(fs *pflag.FlagSet) {
 		10*time.Minute,
 		"The minimum interval at which watched resources are reconciled.",
 	)
-
 	fs.StringVar(
 		&options.ProviderIDFormat,
 		"provider-id-fmt",
-		string(options.ProviderIDFormatV1),
+		string(options.ProviderIDFormatV2),
```
Karthik-K-N, line 109: Since we are making v2 by default, Lets also update the doc to not to explicitly set the provider id format to v2 for example here https://cluster-api-ibmcloud.sigs.k8s.io/topics/powervs/external-cloud-provider.

Also lets check in the whole doc or wherever to update the same. Like manager.yaml https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/config/manager/manager.yaml#L30

Karthik-K-N (commented):
Thanks for submitting the PR, I know its in draft but just an early feedback for making more changes.

main.go
KeerthanaAP: Updated the docs accordingly.

main.go
```diff
@@ -132,6 +131,7 @@ func initFlags(fs *pflag.FlagSet) {
 
 func validateFlags() error {
 	switch options.ProviderIDFormatType(options.ProviderIDFormat) {
+	// Deprecated: ProviderIDFormatV1 is deprecated and will be removed in a future release.
 	case options.ProviderIDFormatV1:
 		setupLog.Info("Using v1 version of ProviderID format")
```
Karthik-K-N, line 136: Lets update this log line to mention whatever you added in comments. So the user will get to know about the deprecation.

KeerthanaAP: Sure will do it.

pkg/options/options.go
```diff
@@ -23,6 +23,8 @@ const (
 	// ProviderIDFormatV1 will set provider id to machine as follows
 	// For VPC machines: ibmvpc://<cluster_name>/<vm_hostname>
 	// For Power VS machines: ibmpowervs://<cluster_name>/<vm_hostname>
+	// Deprecated: ProviderIDFormatV1 is deprecated and will be removed in a future release. see https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/1868 for more details.
```
Karthik-K-N, line 26: Sorry for late update, as you mentioned the provider id format v1 is used for vpc as well and we are deprecating it only for PowerVS, so Lets explicitly mention that its deprecated for PowerVS

main.go
KeerthanaAP: Done.

Karthik-K-N (commented):
/lgtm

Issue #1933, opened 2024-08-30
Validate DHCP server status before reusing existing DHCP network


Karthik-K-N:
/kind feature
/area provider/ibmcloud

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

Today, If user sets spec.dhcpServer.name and while checking for network we assume that there might exist a network associated with the server. If the network exist we are simply considering it. Code ref: https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/blob/3eb2b524c65167565c595d01e475497b066ef22a/cloud/scope/powervs_cluster.go#L884-L891

The request is to validate the DHCP server status before considering the associated network.

Reconcile Network workflow

1. spec.DHCPserver.id is set - BUG, This case is not hanlded
		Comments:  Need to discuss what should be done on how to select network, may be expect network id to be set?

2. spec.DHCPServer.name is set 
	a. if spec.network.id is set - Fetch for network with id and use 
		Comments: DHCPServer name is ignored and network id got preference, 
	b. Try to find network with name DHCPSERVER<spec.DHCPServer.name>_Private and if found use

3. spec.DHCPServer.name is not set  
	a. if spec.network.id is set - Fetch for network with id and use 
		Comments: DHCPServer name is ignored and network id got preference, 

	*. if spec.network.name is set
		a. Try to find network with name spec.network.name

	*. if spec.network.name is not set 
	    a. Try to find network with name DHCPSERVER<cluster_name>_Private and if found use

4. If the network is not found, Create DHCP server with name spec.DHCPServer.Name if set else with clustername
	set both dhcpserverid and network id in status.

5. status.DHCPServer.id is set, then the dhcp server will be created by controller and status.network.id will also be set by controller

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

Karthik-K-N:
@Amulyam24 @dharaneeshvrd also we cannot check DHCP server status when user provides name of existing network right, Because no where we can know, if it has associated DHCP server or not.
Just wanted to know is there any way we can validate a network is functional?

Karthik-K-N:
@Amulyam24 we also need to check whats the flow when we set spec.network.name.

Karthik-K-N:
What will happen when a user sets a network name in spec say "capiNetwork" and which does not exist in cloud, 

I think currently controller does following things, 
[...]

PR #1901, opened 2024-08-01
Attach connections to existing transit gateway when its not there

dharaneeshvrd:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1887 

**Special notes for your reviewer**:

/area provider/ibmcloud

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

Issue #1726, opened 2024-04-15
Document and update logic in controllers for handling the resources identified by either ID or Name.


Karthik-K-N:
/kind feature
/area provider/ibmcloud

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

Currently for any cloud resource for example [PowerVS workspace](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/fbbd4e6bec827bcd9dc5e4d9eab15823d403027f/api/v1beta2/ibmpowervsmachine_types.go#L141), TransitGateway, VPC user can specify it either via ID, Name and in some cases regular expressions and there are some assumptions we made and which needs to be documented
1. If user sets ID we expect it to be exist in cloud or lese fails the and update it via Event/condition on resource.
2. If user sets Name and the resource does not exist with given name we create it.

Along with documentation we need to revisit the controller code to ensure that this flow is achieved.

Few other things to note
1. For VPC subnets and loadbalancers we should not allow users to set duplicate names.
2. Document that if both name and id provided id will get precedence and update logic accordingly.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

Karthik-K-N:
@Amulyam24 I think this is handled as part of your refactor PR right?

Amulyam24:
yes, this has been handled by https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1937. It was referenced in the PR
Closing the issue.
/close

PR #1937, opened 2024-09-02
Refactor reconcilation of resources

### LINKED ISSUE #1907: Refactor Power VS reconcile helpers for better readbility and better code grouping
/kind feature
/area provider/ibmcloud

Currently in any Power VS service reconcilation are broadly classified into 3 different functionalities

1. Fetch service instance id from either spec or status 
   Ref: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/01e1771ca8cbf8b9969bb64b6956b3d470219313/cloud/scope/powervs_cluster.go#L703-L722

2. Check for existing resources in cloud
   Ref: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/01e1771ca8cbf8b9969bb64b6956b3d470219313/cloud/scope/powervs_cluster.go#L725-L733

3. Create the service
  Ref: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/01e1771ca8cbf8b9969bb64b6956b3d470219313/cloud/scope/powervs_cluster.go#L736-L747

Though its working as expected, since we are adding more complex logic as the feature requests, it would be better to refactor to have particular function names rather than generic in most of the cases.

Since we are relaying more on controllerCreated variable to decide upon user passed  vs controller created resources, Eventually we are moving agaist K8s generic princile of level triggered events. IT has its own drawbacks against advantages, We also need to think of better way to avoid this.

Amulyam24:

**This PR refactors the reconciliation of resources to follow a unit format as below**
1. Fetch ID from IBMPowerVSCluster status object. If ID exists, check the status of the resource in cloud and return
2. If ID is not present in the status, try to fetch it from the IBMPowerVSCluster spec and check if the resource exists in cloud. Here we can check either via name or ID passed in the spec. If resource with given ID doesn't exist, return err, or else check the status of the resource in cloud and set IBMPowerVSCluster.status accordingly
3. If resource with given name doesn't exist, proceed with creating the resource. If it exists, check the status of the resource in cloud. Set IBMPowerVSCluster.status accordingly in both cases

**The changes in this PR has been tested with the following scenarios**
1. Resource names are set in spec and they do not exist
2. Resource names are set in spec and they exist
3. Resource IDs are set in spec and they do not exist
4. Resource IDs are set in spec and they exist
5. Spec with only required fields

**Note:**
1. Adjustments were made to support if ID or Name are set in spec
2. The changes in this PR were also tested with TG changes from existing PR https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1901
3. Caveats noticed where in terms of handling existing DHCP server ID set in spec and it has been documented in https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/1933

Fixes #1907 #1726

/area provider/ibmcloud

**Release note**:

```release-note
Refactor reconcilation of resources
```

### DIFF
--- a/api/v1beta2/ibmpowervscluster_webhook.go
+++ b/api/v1beta2/ibmpowervscluster_webhook.go
@@ -102,11 +102,34 @@ func (r *IBMPowerVSCluster) validateIBMPowerVSClusterNetwork() *field.Error {
 	return nil
 }
 
+func (r *IBMPowerVSCluster) validateIBMPowerVSClusterLoadBalancers() (allErrs field.ErrorList) {
+	if err := r.validateIBMPowerVSClusterLoadBalancerNames(); err != nil {
+		allErrs = append(allErrs, err...)
+	}
+
+	if len(r.Spec.LoadBalancers) == 0 {
+		return allErrs
+	}
+
+	for _, loadbalancer := range r.Spec.LoadBalancers {
+		if *loadbalancer.Public {
+			return allErrs
+		}
+	}
+
+	return append(allErrs, field.Invalid(field.NewPath("spec.LoadBalancers"), r.Spec.LoadBalancers, "Expect atleast one of the load balancer to be public"))
+}
+
 func (r *IBMPowerVSCluster) validateIBMPowerVSClusterLoadBalancerNames() (allErrs field.ErrorList) {
 	found := make(map[string]bool)
 	for i, loadbalancer := range r.Spec.LoadBalancers {
+		if loadbalancer.Name == "" {
+			continue
+		}
+
 		if found[loadbalancer.Name] {
 			allErrs = append(allErrs, field.Duplicate(field.NewPath("spec", fmt.Sprintf("loadbalancers[%d]", i)), map[string]interface{}{"Name": loadbalancer.Name}))
+			continue
 		}
 		found[loadbalancer.Name] = true
 	}
@@ -117,8 +140,12 @@ func (r *IBMPowerVSCluster) validateIBMPowerVSClusterLoadBalancerNames() (allErr
 func (r *IBMPowerVSCluster) validateIBMPowerVSClusterVPCSubnetNames() (allErrs field.ErrorList) {
 	found := make(map[string]bool)
 	for i, subnet := range r.Spec.VPCSubnets {
+		if subnet.Name == nil {
+			continue
+		}
 		if found[*subnet.Name] {
 			allErrs = append(allErrs, field.Duplicate(field.NewPath("spec", fmt.Sprintf("vpcSubnets[%d]", i)), map[string]interface{}{"Name": *subnet.Name}))
+			continue
 		}
 		found[*subnet.Name] = true
 	}
@@ -128,6 +155,9 @@ func (r *IBMPowerVSCluster) validateIBMPowerVSClusterVPCSubnetNames() (allErrs f
 
 func (r *IBMPowerVSCluster) validateIBMPowerVSClusterTransitGateway() *field.Error {
 	if r.Spec.Zone == nil && r.Spec.VPC == nil {
+		return nil
+	}
+	if r.Spec.TransitGateway == nil {
 		return nil
 	}
 	if _, globalRouting, _ := genUtil.GetTransitGatewayLocationAndRouting(r.Spec.Zone, r.Spec.VPC.Region); r.Spec.TransitGateway.GlobalRouting != nil && !*r.Spec.TransitGateway.GlobalRouting && globalRouting != nil && *globalRouting {
@@ -183,7 +213,7 @@ func (r *IBMPowerVSCluster) validateIBMPowerVSClusterCreateInfraPrereq() (allErr
 		allErrs = append(allErrs, err...)
 	}
 
-	if err := r.validateIBMPowerVSClusterLoadBalancerNames(); err != nil {
+	if err := r.validateIBMPowerVSClusterLoadBalancers(); err != nil {
 		allErrs = append(allErrs, err...)
 	}
 
--- a/cloud/scope/powervs_cluster.go
+++ b/cloud/scope/powervs_cluster.go
@@ -384,14 +384,8 @@ func (s *PowerVSClusterScope) ServiceInstance() *infrav1beta2.IBMPowerVSResource
 	return s.IBMPowerVSCluster.Spec.ServiceInstance
 }
 
-// GetServiceInstanceID get the service instance id.
+// GetServiceInstanceID returns service instance id set in status field of IBMPowerVSCluster object. If it doesn't exist, returns empty string.
 func (s *PowerVSClusterScope) GetServiceInstanceID() string {
-	if s.IBMPowerVSCluster.Spec.ServiceInstanceID != "" {
-		return s.IBMPowerVSCluster.Spec.ServiceInstanceID
-	}
-	if s.IBMPowerVSCluster.Spec.ServiceInstance != nil && s.IBMPowerVSCluster.Spec.ServiceInstance.ID != nil {
-		return *s.IBMPowerVSCluster.Spec.ServiceInstance.ID
-	}
 	if s.IBMPowerVSCluster.Status.ServiceInstance != nil && s.IBMPowerVSCluster.Status.ServiceInstance.ID != nil {
 		return *s.IBMPowerVSCluster.Status.ServiceInstance.ID
 	}
@@ -470,11 +464,8 @@ func (s *PowerVSClusterScope) Network() *infrav1beta2.IBMPowerVSResourceReferenc
 	return &s.IBMPowerVSCluster.Spec.Network
 }
 
-// GetDHCPServerID returns the DHCP id from spec or status of IBMPowerVSCluster object.
+// GetDHCPServerID returns the DHCP id from status of IBMPowerVSCluster object. If it doesn't exist, returns nil.
 func (s *PowerVSClusterScope) GetDHCPServerID() *string {
-	if s.IBMPowerVSCluster.Spec.DHCPServer != nil && s.IBMPowerVSCluster.Spec.DHCPServer.ID != nil {
-		return s.IBMPowerVSCluster.Spec.DHCPServer.ID
-	}
 	if s.IBMPowerVSCluster.Status.DHCPServer != nil {
 		return s.IBMPowerVSCluster.Status.DHCPServer.ID
 	}
@@ -491,11 +482,8 @@ func (s *PowerVSClusterScope) VPC() *infrav1beta2.VPCResourceReference {
 	return s.IBMPowerVSCluster.Spec.VPC
 }
 
-// GetVPCID returns the VPC id.
+// GetVPCID returns the VPC id set in status field of IBMPowerVSCluster object. If it doesn't exist, returns nil.
 func (s *PowerVSClusterScope) GetVPCID() *string {
-	if s.IBMPowerVSCluster.Spec.VPC != nil && s.IBMPowerVSCluster.Spec.VPC.ID != nil {
-		return s.IBMPowerVSCluster.Spec.VPC.ID
-	}
 	if s.IBMPowerVSCluster.Status.VPC != nil {
 		return s.IBMPowerVSCluster.Status.VPC.ID
 	}
@@ -516,15 +504,6 @@ func (s *PowerVSClusterScope) GetVPCSubnetID(subnetName string) *string {
 // GetVPCSubnetIDs returns all the VPC subnet ids.
 func (s *PowerVSClusterScope) GetVPCSubnetIDs() []*string {
 	subnets := []*string{}
-	// use the vpc subnet id set by user.
-	for _, subnet := range s.IBMPowerVSCluster.Spec.VPCSubnets {
-		if subnet.ID != nil {
-			subnets = append(subnets, subnet.ID)
-		}
-	}
-	if len(subnets) != 0 {
-		return subnets
-	}
 	if s.IBMPowerVSCluster.Status.VPCSubnet == nil {
 		return nil
 	}
@@ -534,8 +513,8 @@ func (s *PowerVSClusterScope) GetVPCSubnetIDs() []*string {
 	return subnets
 }
 
-// SetVPCSubnetID set the VPC subnet id.
-func (s *PowerVSClusterScope) SetVPCSubnetID(name string, resource infrav1beta2.ResourceReference) {
+// SetVPCSubnetStatus set the VPC subnet id.
+func (s *PowerVSClusterScope) SetVPCSubnetStatus(name string, resource infrav1beta2.ResourceReference) {
 	s.V(3).Info("Setting status", "name", name, "resource", resource)
 	if s.IBMPowerVSCluster.Status.VPCSubnet == nil {
 		s.IBMPowerVSCluster.Status.VPCSubnet = make(map[string]infrav1beta2.ResourceReference)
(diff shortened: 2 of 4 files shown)


cloud/scope/powervs_cluster.go
```diff
@@ -1125,6 +1127,7 @@ func (s *PowerVSClusterScope) ReconcileVPCSubnets() (bool, error) {
 				return false, fmt.Errorf("failed to get VPC subnet with ID %s", *subnetID)
 			}
 			// check for next subnet
+			s.SetVPCSubnetStatus(*subnetDetails.Name, infrav1beta2.ResourceReference{ID: subnetDetails.ID})
```
Amulyam24, line 1130: This was added mainly to handle the scenario where existing resource ID is passed in spec. Might not be the best way but found this easier as of now with the current code flow as changes otherwise was making the code complicated

Karthik-K-N: When id is set in spec, why do you think we need to set it in status as well? Do we follow same thing for other resources as well.

cloud/scope/powervs_cluster.go
```diff
@@ -1100,7 +1101,7 @@ func (s *PowerVSClusterScope) ReconcileVPCSubnets() (bool, error) {
 		}
 	}
 	for index, subnet := range s.IBMPowerVSCluster.Spec.VPCSubnets {
-		if subnet.Name == nil {
+		if subnet.ID == nil && subnet.Name == nil {
```
Karthik-K-N, line 1104: If I understand correctly, You want to fetch the name of associated subnet id correct?

cloud/scope/powervs_cluster.go
```diff
@@ -1136,9 +1139,7 @@ func (s *PowerVSClusterScope) ReconcileVPCSubnets() (bool, error) {
 		}
 		if vpcSubnetID != "" {
 			s.V(3).Info("Found VPC subnet in IBM Cloud", "id", vpcSubnetID)
-			s.SetVPCSubnetID(*subnet.Name, infrav1beta2.ResourceReference{ID: &vpcSubnetID, ControllerCreated: ptr.To(false)})
-			// check for next subnet
-			continue
+			s.SetVPCSubnetStatus(*subnet.Name, infrav1beta2.ResourceReference{ID: &vpcSubnetID, ControllerCreated: ptr.To(false)})
```
Karthik-K-N, line 1142: why continue removed from here? We are supposed to move to next subnet?

cloud/scope/powervs_cluster.go
```diff
@@ -1791,18 +1791,22 @@ func (s *PowerVSClusterScope) createTransitGateway() (*string, error) {
 		return nil, fmt.Errorf("failed to fetch resource group ID for resource group %v, ID is empty", s.ResourceGroup())
 	}
 
+	if s.IBMPowerVSCluster.Status.ServiceInstance == nil || s.IBMPowerVSCluster.Status.VPC == nil {
+		return nil, fmt.Errorf(" failed to proeceed with transit gateway creation as either one of VPC or PowerVS service instance reconciliation is not successful")
```
Karthik-K-N, line 1795:
```suggestion
		return nil, fmt.Errorf("failed to proeceed with transit gateway creation as either one of VPC or PowerVS service instance reconciliation is not successful")
```

cloud/scope/powervs_cluster.go
```diff
@@ -1899,6 +1903,7 @@ func (s *PowerVSClusterScope) ReconcileLoadBalancers() (bool, error) {
 			return false, err
 		}
 		if loadBalancerStatus != nil {
+			s.V(3).Info("Found VPC load balancer in IBM Cloud", "id", *loadBalancerStatus.ID)
 			s.SetLoadBalancerStatus(loadBalancer.Name, *loadBalancerStatus)
```
Karthik-K-N, line 1907: I think here loadBalancer.Name will not be set properly if its nil. Scope of loadBalancer in line 1880 will end after if block.
You also added check loadBalancer.ID == nil && loadBalancer.Name there will be an issue when id is not nil but name is nil.

cloud/scope/powervs_cluster.go
```diff
@@ -2284,7 +2298,7 @@ func (s *PowerVSClusterScope) GetServiceName(resourceType infrav1beta2.ResourceT
 		if s.Network() == nil || s.Network().Name == nil {
 			return ptr.To(fmt.Sprintf("DHCPSERVER%s_Private", s.InfraCluster()))
 		}
-		return s.Network().Name
+		return ptr.To(fmt.Sprintf("DHCPSERVER%s_Private", *s.Network().Name))
```
Karthik-K-N, line 2301: What made you add this, I am not able to recollect!

cloud/scope/powervs_cluster.go
Amulyam24: We want to set the subnet name only is ID and name are nil. With the previous condition, in a case where we set subnet ID in spec, name will be empty and here we will be setting the name which would set different subnet name than the one passed via ID. Though `subnet.Name` will not be used in flow when subnet ID is set, thought better to set name only when appropriate.

cloud/scope/powervs_cluster.go
Amulyam24: Currently, we are following this for all resources. As `controllerCreated` will be false in this case, ideally, we wouldn't need ID which would be used during deletion.
Should we consider setting ID only when name is passed?

cloud/scope/powervs_cluster.go
Amulyam24: Removed by mistake. Rectified!

cloud/scope/powervs_cluster.go
Amulyam24: The loadBalancer variable here is referring to this right? https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/1f473c2bc73e3daadd4a2c0747d462ef8e8bbccd/cloud/scope/powervs_cluster.go#L1867 

and the flow will reach here when resource is checked via given name and it will be set in either case to the user provided name or default name as per the flow

> You also added check loadBalancer.ID == nil && loadBalancer.Name there will be an issue when id is not nil but name is nil.

can you elaborate more on this?

cloud/scope/powervs_cluster.go
Amulyam24: In a case where `network.Name` is set in spec to a DHCP based n/w and resource exists in cloud, the flow wasn't working as expected as we check for resource with spec.name(for example: capi-test) against the name in cloud which will be in a different format(DHCPSERVERcapi-test_Private). Hence added this change to convert the name accordingly.

However, this would be wrong if user uses a different network type. Reverted this change and will need to revisit this.

cloud/scope/powervs_cluster.go
Karthik-K-N: Can you check the flow when loadbalancer.id is set and loadbalancer.name is nil?

controllers/ibmpowervscluster_controller.go
```diff
@@ -331,8 +331,8 @@ func (r *IBMPowerVSClusterReconciler) reconcile(clusterScope *scope.PowerVSClust
 	}
 
 	// update cluster object with loadbalancer host
-	loadBalancer := clusterScope.PublicLoadBalancer()
-	if loadBalancer == nil {
+	loadBalancer, err := clusterScope.PublicLoadBalancer()
+	if err != nil || loadBalancer == nil {
```
Karthik-K-N, line 335: I think in case of err, its better to log error explicitly.

cloud/scope/powervs_cluster.go
Amulyam24: When loadbalancer.ID is set we check the resource with given ID in cloud and set the status [here](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1937/files#diff-a12914f75cce0ea4c617f28f71fcfbd158441b44cb39595fe88f16c69c269b79R1897)

Karthik-K-N (commented):
/lgtm

Amulyam24:
/hold

Testing the changes after rebase.

mkumatag (changes requested):

cloud/scope/powervs_cluster.go
```diff
@@ -600,20 +579,29 @@ func (s *PowerVSClusterScope) GetTransitGatewayID() *string {
...
 	// if the user did not specify any loadbalancer then return the public loadbalancer created by the controller.
 	if len(s.IBMPowerVSCluster.Spec.LoadBalancers) == 0 {
 		return &infrav1beta2.VPCLoadBalancerSpec{
 			Name:   *s.GetServiceName(infrav1beta2.ResourceTypeLoadBalancer),
 			Public: ptr.To(true),
-		}
+		}, nil
 	}
 	for _, lb := range s.IBMPowerVSCluster.Spec.LoadBalancers {
+		if lb.ID != nil && lb.Name == "" {
```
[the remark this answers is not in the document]
Amulyam24: Refactored the logic, PTAL

Amulyam24:
/unhold

PR is ready for review, PTAL

dharaneeshvrd (commented):
Overall LGTM

cloud/scope/powervs_cluster_test.go
```diff
@@ -868,166 +758,337 @@ func TestGetLoadBalancerState(t *testing.T) {
...
-								Public: ptr.To(true),
-							},
-							{
-								Name:   "lb2",
-								Public: ptr.To(true),
-							},
+		}
+
+		hostName, err := clusterScope.GetPublicLoadBalancerHostName()
+		g.Expect(hostName).To(Equal(ptr.To("lb1-hostname")))
```
Karthik-K-N, line 938: Hope it wont flake with order mismatch?

cloud/scope/powervs_cluster_test.go
```diff
@@ -1234,20 +1321,51 @@ func TestReconcilePowerVSServiceInstance(t *testing.T) {
...
-			mockResourceController.EXPECT().GetServiceInstance(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil)
-			mockResourceController.EXPECT().CreateResourceInstance(gomock.Any()).Return(nil, nil, errors.New("failed to create resource instance"))
-			return mockResourceController, nil
+		clusterScope := PowerVSClusterScope{
+			ResourceClient: mockResourceController,
+			IBMPowerVSCluster: &infrav1beta2.IBMPowerVSCluster{
+				Spec: infrav1beta2.IBMPowerVSClusterSpec{
+					ServiceInstance: &infrav1beta2.IBMPowerVSResourceReference{
+						ID: ptr.To(serviceInstanceID),
+					},
```
Karthik-K-N, line 1335: Id is same in both spec and status, It is supposed to be different for this case right?

cloud/scope/powervs_cluster_test.go
```diff
@@ -1234,20 +1321,51 @@ func TestReconcilePowerVSServiceInstance(t *testing.T) {
...
+				},
+			},
+		}
+
+		instance := &resourcecontrollerv2.ResourceInstance{
+			Name:  ptr.To("test-instance"),
+			State: ptr.To("active"),
 		}
-		clusterScope, err := NewPowerVSClusterScope(clusterScopeParams)
+		mockResourceController.EXPECT().GetResourceInstance(gomock.Any()).Return(instance, nil, nil)
```
Karthik-K-N, line 1349: May be we update the mock to take the status.id and return the instance?

cloud/scope/powervs_cluster_test.go
Amulyam24: No, it will return the first public load balancer as expected.

cloud/scope/powervs_cluster.go
```diff
@@ -716,7 +705,7 @@ func (s *PowerVSClusterScope) ReconcileResourceGroup() error {
 
 // ReconcilePowerVSServiceInstance reconciles Power VS service instance.
 func (s *PowerVSClusterScope) ReconcilePowerVSServiceInstance() (bool, error) {
-	// Verify if service instance id is set in spec or status field of IBMPowerVSCluster object.
+	// Verify if service instance id is set in status field of IBMPowerVSCluster object.
 	serviceInstanceID := s.GetServiceInstanceID()
 	if serviceInstanceID != "" {
 		s.V(3).Info("PowerVS service instance ID is set, fetching details", "id", serviceInstanceID)
```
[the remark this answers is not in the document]
Amulyam24: Updated

cloud/scope/powervs_cluster_test.go
Amulyam24: Yes, updated it

controllers/ibmpowervscluster_controller.go
```diff
@@ -330,20 +330,18 @@ func (r *IBMPowerVSClusterReconciler) reconcile(clusterScope *scope.PowerVSClust
...
-		return reconcile.Result{}, fmt.Errorf("failed to fetch public loadbalancer")
-	}
-
 	clusterScope.Info("Getting load balancer host")
-	hostName := clusterScope.GetLoadBalancerHostName(loadBalancer.Name)
+	hostName, err := clusterScope.GetPublicLoadBalancerHostName()
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to fetch public loadbalancer: %w", err)
+	}
 	if hostName == nil || *hostName == "" {
```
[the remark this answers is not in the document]
Amulyam24: > Currently we expect at least one public loadbalancer right, Otherwise we will endup in infinite loop here

With the current flow, this might not occur but added a check just to be safe.

Karthik-K-N (commented):
One suggestion , Otherwises LGTM

api/v1beta2/ibmpowervscluster_webhook.go
```diff
@@ -102,11 +102,36 @@ func (r *IBMPowerVSCluster) validateIBMPowerVSClusterNetwork() *field.Error {
...
 
+func (r *IBMPowerVSCluster) validateIBMPowerVSClusterLoadBalancers() (allErrs field.ErrorList) {
+	if err := r.validateIBMPowerVSClusterLoadBalancerNames(); err != nil {
+		allErrs = append(allErrs, err...)
+	}
+
+	isPublic := false
+	for _, loadbalancer := range r.Spec.LoadBalancers {
+		if *loadbalancer.Public {
+			isPublic = true
```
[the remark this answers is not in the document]
Amulyam24: Addressed.

api/v1beta2/ibmpowervscluster_webhook.go
```diff
@@ -102,11 +102,36 @@ func (r *IBMPowerVSCluster) validateIBMPowerVSClusterNetwork() *field.Error {
...
+	}
+
+	return allErrs
+}
+
 func (r *IBMPowerVSCluster) validateIBMPowerVSClusterLoadBalancerNames() (allErrs field.ErrorList) {
 	found := make(map[string]bool)
 	for i, loadbalancer := range r.Spec.LoadBalancers {
+		if loadbalancer.Name == "" {
+			continue
```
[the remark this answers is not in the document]
Amulyam24: It checks if LB with given ID is present and for the second time(with same ID), it will override the already set status and adds only one LB to the status field.

Amulyam24:
@Karthik-K-N, added a couple of [minor changes](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/compare/d4694566f1a2ac547918d8748749ca028dc80fe4..56b874e1087617ab964f7051b3a6bbecb719b2c2). PTAL!

Karthik-K-N (commented):
Thank you.
/lgtm

Amulyam24:
@mkumatag, PTAL!
The following/open UT PRs will require changes as per refactoring being added in this PR.

Issue #1896, opened 2024-07-25
Extend VPC (x86_64) Infrastructure reconciliation support


cjschaef:
/kind feature
/area provider/ibmcloud

**Describe the solution you'd like**
The current implementation for reconciling VPC (x86_64) Infrastructure support is very basic and lacks full feature/resource support for creating VPC Resources to be used for cluster Infrastructure. Expending the support for VPC to reconcile more resources and in more complex configurations can help with a more complete set of Infrastructure for clusters.

**NOTE: Implementation for extended Machine reconciliation will be covered separately**

**Anything else you would like to add:**
In order to prevent breaking existing (legacy) functionality, a new code path (a v2 path) will be developed, for more complex field parsing, configuration and resource definitions. We'd expect to eventually migrate all VPC Infrastructure reconciliation to this new path and deprecate and remove the old functionality.

---

### Expected Functional changes
Major items identified, but not limited to:
- New API's
    - Multiple Subnets
    - Custom Image
    - Security Group and Security Group Rules
    - Load Balancer resources (pools, etc.)
- New Cluster Controller Reconciliation for
    - VPC
    - Custom Image
    - Subnets
    - Security Groups and Security Group Rules
    - Load Balancers

### Expected Prerequisites
- Existing Resource Group (RG for any existing resources and the Cluster's Resource Group)
- Existing COS Instance, Bucket, Object containing qcow image for VPC Custom Image generation (or provided existing VPC Custom Image)
- IAM Authorization for VPC to COS for creating VPC Custom Image using COS Object

### Expected Resource Specifications
- Existing resources can be designated by Name or ID (where applicable)
- Resources specified by ID that don't exist or cannot be found results in an error
- Resources specified by Name will trigger a lookup (treated as existing), or get created if not found (some limitations include ResourceGroup, which will not be created regardless of how it was designated)
- Certain resources are dependent on other resources, requiring a hierarchy of reconciliation/creation, but also, limits the way a resource is specified, even if it is expected to exist (runtime resource ID's are not available prior to Infrastructure creation, meaning names may be designated instead, but if those resources don't already exist at that point of resource reconciliation, an error will result rather than try to create). Some examples of this:  SecurityGroupRules can depend on Subnet CIDR's or other SecurityGroups which must already exist;  Load Balancers can have Subnets and SecurityGroups attached during creation, so those resources must already exist

mkumatag:
[...]

PR #1942, opened 2024-09-04
VPC: Add Custom Image reconciliation

cjschaef:
Add support to reconcile a VPC Custom Image for the new v2 VPC Infrastructure reconcile logic.

Related: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/1896

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Add support to reconcile VPC Custom Images during Infrastructure creation.
```


cloud/scope/vpc_cluster.go
```diff
@@ -23,6 +23,7 @@ import (
 
 	"github.com/go-logr/logr"
 
+	"github.com/IBM-Cloud/bluemix-go/crn"
```
dharaneeshvrd, line 26: Can we please not use bluemix just for parsing crn? Since we are not using the main functionality(connecting to ibmcloud), it may create dependency issues in future for capi.
I can see they have just split the crn and directly access the id by index. We can add this into some util func and use it locally.

cloud/scope/vpc_cluster.go
```diff
@@ -486,9 +497,186 @@ func (s *VPCClusterScope) createVPC() (*vpcv1.VPC, error) {
 	} else if vpcDetails == nil {
 		return nil, fmt.Errorf("no vpc details after creation")
 	}
-	if err = s.TagResource(s.IBMVPCCluster.Name, *vpcDetails.CRN); err != nil {
+
+	// NOTE: This tagging is only attempted once. We may wish to refactor in case this single attempt fails.
+	if err = s.TagResource(s.Name(), *vpcDetails.CRN); err != nil {
```
dharaneeshvrd, line 502: Do we need to make this change as part of this PR?
It seems not relevant to me. Please see whether we can create a separate issue and handle it separately.

cloud/scope/vpc_cluster.go
```diff
@@ -486,9 +497,186 @@ func (s *VPCClusterScope) createVPC() (*vpcv1.VPC, error) {
...
+
+	// Check status of VPC Custom Image.
+	if imageID != nil {
+		image, _, err := s.VPCClient.GetImage(&vpcv1.GetImageOptions{
+			ID: imageID,
+		})
+		if err != nil {
+			return false, fmt.Errorf("error retrieving vpc custom image by id: %w", err)
+		}
+		if image == nil {
```
dharaneeshvrd, line 547: Since you are passing id won't it return an error if the image is not found?
Please see whether we need this nil check?

cloud/scope/vpc_cluster.go
```diff
@@ -486,9 +497,186 @@ func (s *VPCClusterScope) createVPC() (*vpcv1.VPC, error) {
...
+		Name: image.Name,
+		// We must wait for the image to be ready, on followup reconciliation loops.
+		Ready: false,
+	})
+	return true, nil
+}
+
+// createCustomImage will create a new VPC Custom Image.
+func (s *VPCClusterScope) createCustomImage() (*vpcv1.Image, error) {
+	if s.IBMVPCCluster.Spec.Image == nil {
```
dharaneeshvrd, line 594: This nil check already happening in main Reconcile func, please remove here.

cloud/scope/vpc_cluster.go
```diff
@@ -486,9 +497,186 @@ func (s *VPCClusterScope) createVPC() (*vpcv1.VPC, error) {
...
+				ID: resourceGroupID,
+			},
+		},
+	}
+
+	imageDetails, _, err := s.VPCClient.CreateImage(options)
+	if err != nil {
+		return nil, fmt.Errorf("error unknown failure creating vpc custom image: %w", err)
+	}
+	if imageDetails == nil || imageDetails.ID == nil || imageDetails.Name == nil || imageDetails.CRN == nil {
```
dharaneeshvrd, line 653: Why do we need to check all the params separately?
If the error is nil, would imageDetails still be not nil?

cloud/scope/vpc_cluster.go
```diff
@@ -486,9 +497,186 @@ func (s *VPCClusterScope) createVPC() (*vpcv1.VPC, error) {
...
+		id, err := s.GetResourceGroupID()
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving resource group id: %w", err)
+		}
+		resourceGroupID = ptr.To(id)
+	}
+
+	// We must have an OperatingSystem value supplied in order to create the Custom Image.
+	// NOTE(cjschaef): Perhaps we could try defaulting this value, so it isn't required for Custom Image creation.
+	if s.IBMVPCCluster.Spec.Image.OperatingSystem == nil {
```
dharaneeshvrd, line 622: If you are planning to keep this, you can add this check into webhook.

dharaneeshvrd (changes requested):

cloud/scope/vpc_cluster.go
mkumatag: +1, we purposefully migrated from bluemix-go client to other individual clients from IBM org, wondering if there is any other alternative for the crn validation?!

cjschaef: Is there an alternative golang module I can use instead for parsing CRN's?
Is this a valid alternative?
https://github.com/IBM-Cloud/ibm-cloud-cli-sdk/tree/master/bluemix/crn

It would be helpful to keep the CRN for the image, for future enhancements regarding account details, etc. from a Custom Image CRN that gets supplied.

Should I reimplement the CRN logic within CAPI then?

cloud/scope/vpc_cluster.go
cjschaef: I can create a second PR, if that PR can be reviewed and merged prior to this PR, to fix this invalid reference.

cloud/scope/vpc_cluster.go
cjschaef: If the image is not found, I thought we got a 404 DetailedResponse back. I am defensively not expecting a non-error (`error == nil`) to result in `image` always being set.

cloud/scope/vpc_cluster.go
cjschaef: Sorry, I'm not sure what the webhook expected change would be, can you provide more details?

cloud/scope/vpc_cluster.go
cjschaef: This is done defensively, to make sure the required Image fields are populated from the API response.

cloud/scope/vpc_cluster.go
```diff
@@ -486,9 +497,186 @@ func (s *VPCClusterScope) createVPC() (*vpcv1.VPC, error) {
...
+	imageDetails, _, err := s.VPCClient.CreateImage(options)
+	if err != nil {
+		return nil, fmt.Errorf("error unknown failure creating vpc custom image: %w", err)
+	}
+	if imageDetails == nil || imageDetails.ID == nil || imageDetails.Name == nil || imageDetails.CRN == nil {
+		return nil, fmt.Errorf("error failed creating custom image")
+	}
+
+	// NOTE: This tagging is only attempted once. We may wish to refactor in case this single attempt fails.
+	if err := s.TagResource(s.Name(), *imageDetails.CRN); err != nil {
```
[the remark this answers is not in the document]
cjschaef: If tagging fails, after creating the Custom Image (or any resource with this similar logic, like VPC), we return an error. Next loop through could then find the resource that was created the first attempt, and update the resource Status (in the VPC example, a VPC is likely "ready" in the second attempt), with the expectation it is found by the name used during creation. This will monitor and/or update the resource Status accordingly. As we expect to not try to recreate a resource that was already created, we don't expect to try to retag the resource.

Logic could be added to check tags and apply if missing, but that logic will be more cumbersome and time consuming that desired currently, and I only plan to worry about tags during Delete actions, which are a secondary concern currently.

cloud/scope/vpc_cluster.go
cjschaef: Is this additional API changes, such as enhancing this function?
https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/5f99eed4ed831535c7f238a570d77494b5c98540/api/v1beta2/ibmvpccluster_webhook.go#L72

cloud/scope/vpc_cluster.go
cjschaef: I have the separate PR opened for the necessary fixes.
https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1949

cloud/scope/vpc_cluster.go
mkumatag: > Should I reimplement the CRN logic within CAPI then?

lets do it..

cloud/scope/vpc_cluster.go
dharaneeshvrd: Yes. But you can take that as an separate issue.

cloud/scope/vpc_cluster.go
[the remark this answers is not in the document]
dharaneeshvrd: I can see you have handled this in ReconcileVPC, but here, you don't have a func like GetVPCID. You are checking either from status or from spec crn. If it's there you are validating the image and set the status based on that. Please take care of this here.

dharaneeshvrd (changes requested):

cloud/scope/vpc_cluster.go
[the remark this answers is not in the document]
cjschaef: Not sure I am following, but if the recommendation here is to update status (SetResourceStatus) prior to tagging, I can move that for Custom Images.

I can handle VPC in a similar fashion then, in a separate PR I expect.

cloud/scope/vpc_cluster.go
```diff
@@ -486,9 +497,186 @@ func (s *VPCClusterScope) createVPC() (*vpcv1.VPC, error) {
...
+			Name: image.Name,
+			// Ready status will be invert of the need to requeue.
+			Ready: !requeue,
+		})
+		return requeue, nil
+	}
+
+	// No VPC Custom Image exists or was found.
+	// So, check if the Image spec was defined, as it contains all the data necessary to reconcile.
+	if s.IBMVPCCluster.Spec.Image == nil {
```
cjschaef, line 567: I have moved this check earlier, will be reflected once I push updates.

Karthik-K-N (commented):
For using this CustomImage for machine creation, will you be submitting another PR or planning to add to this as separate commit.

cloud/scope/vpc_cluster.go
```diff
@@ -486,9 +497,186 @@ func (s *VPCClusterScope) createVPC() (*vpcv1.VPC, error) {
...
+		s.V(3).Info("No VPC Custom Image defined, skipping reconciliation")
+		return false, nil
+	}
+
+	// Create Custom Image.
+	s.Info("Creating a VPC Custom Image")
+	image, err := s.createCustomImage()
+	if err != nil {
+		return false, fmt.Errorf("error failure trying to create vpc custom image: %w", err)
+	} else if image == nil {
```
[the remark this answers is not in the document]
cjschaef: I've refactored the function and return values, so this block has been removed (it will be reflected once I push up changes).

cloud/scope/vpc_cluster.go
cjschaef: I can do that until it gets removed in favor of webhook validation then.

cloud/scope/vpc_cluster.go
```diff
@@ -486,9 +497,186 @@ func (s *VPCClusterScope) createVPC() (*vpcv1.VPC, error) {
...
+	// NOTE(cjschaef): Perhaps we could try defaulting this value, so it isn't required for Custom Image creation.
+	if s.IBMVPCCluster.Spec.Image.OperatingSystem == nil {
+		return nil, fmt.Errorf("error failed to create vpc custom image due to missing operatingSystem")
+	}
+
+	// Build the COS Object URL using the ImageSpec
+	fileHRef, err := s.buildCOSObjectHRef()
+	if err != nil {
+		return nil, fmt.Errorf("error building vpc custom image file href: %w", err)
+	} else if fileHRef == nil {
```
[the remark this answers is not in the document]
cjschaef: I will remove.

cloud/scope/vpc_cluster.go
```diff
@@ -486,9 +497,186 @@ func (s *VPCClusterScope) createVPC() (*vpcv1.VPC, error) {
...
+	if err := s.TagResource(s.Name(), *imageDetails.CRN); err != nil {
+		return nil, fmt.Errorf("error failure tagging vpc custom image: %w", err)
+	}
+	return imageDetails, nil
+}
+
+// buildCOSObjectHRef will build the HRef path to a COS Object that can be used for VPC Custom Image creation.
+func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
+	// We need COS details in order to create the Custom Image from.
+	if s.IBMVPCCluster.Spec.Image.COSInstance == nil || s.IBMVPCCluster.Spec.Image.COSBucket == nil || s.IBMVPCCluster.Spec.Image.COSObject == nil {
```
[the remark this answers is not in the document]
cjschaef: I can create an issue.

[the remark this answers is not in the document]
cjschaef: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/1957

cloud/scope/vpc_cluster.go
```diff
@@ -494,3 +504,185 @@ func (s *VPCClusterScope) createVPC() (*vpcv1.VPC, error) {
...
+		if err != nil {
+			return false, fmt.Errorf("error parsing vpc custom image crn: %w", err)
+		}
+		// If the value provided isn't a CRN or is missing the Resource ID, raise an error.
+		if imageCRN == nil || imageCRN.Resource == "" {
+			return false, fmt.Errorf("error parsing vpc custom image crn, missing resource id")
+		}
+		// If we didn't hit an error during parsing, and Resource was set, set that as the Image ID.
+		imageID = ptr.To(imageCRN.Resource)
+	}
```
dharaneeshvrd, line 557:
```suggestion
        if s.IBMVPCCluster.Spec.Image == nil {
		// If no Image spec was defined, we expect it is maintained externally and continue without reconciling. For example, using a Catalog Offering Custom Image, which may be in another account, which means it cannot be looked up, but can be used when creating Instances.
		s.V(3).Info("No VPC Custom Image defined, skipping reconciliation")
		return false, nil
	} else if s.IBMVPCCluster.Spec.Image.Name != nil {
		// Attempt to retrieve the image details via the name, if it already exists
		imageDetails, err := s.VPCClient.GetImageByName(*s.IBMVPCCluster.Spec.Image.Name)
		if err != nil {
			return false, fmt.Errorf("error checking vpc custom image by name: %w", err)
		} else if imageDetails != nil && imageDetails.ID != nil {
			// Prevent relookup (API request) of VPC Custom Image if we already have the necessary data
			requeue := true
			if imageDetails.Status != nil && *imageDetails.Status == string(vpcv1.ImageStatusAvailableConst) {
				requeue = false
			}
			s.SetResourceStatus(infrav1beta2.ResourceTypeCustomImage, &infrav1beta2.ResourceStatus{
				ID:   *imageDetails.ID,
				Name: s.IBMVPCCluster.Spec.Image.Name,
				// Ready status will be invert of the need to requeue.
				Ready: !requeue,
			})
			return requeue, nil
		}
	}
	
	var imageID *string
	// Attempt to collect VPC Custom Image info from Status.
	if s.IBMVPCCluster.Status.Image != nil {
		if s.IBMVPCCluster.Status.Image.ID != "" {
			imageID = ptr.To(s.IBMVPCCluster.Status.Image.ID)
		}
	} else if s.IBMVPCCluster.Spec.Image.CRN != nil {
		// Parse the supplied Image CRN for Id, to perform image lookup.
		imageCRN, err := ParseCRN(*s.IBMVPCCluster.Spec.Image.CRN)
		if err != nil {
			return false, fmt.Errorf("error parsing vpc custom image crn: %w", err)
		}
		// If the value provided isn't a CRN or is missing the Resource ID, raise an error.
		if imageCRN == nil || imageCRN.Resource == "" {
			return false, fmt.Errorf("error parsing vpc custom image crn, missing resource id")
		}
		// If we didn't hit an error during parsing, and Resource was set, set that as the Image ID.
		imageID = ptr.To(imageCRN.Resource)
	}
```

Can we please rearrange something like this? 
Checking existing image based on name can be a separate block and using id to check the image status in a different block.

pkg/cloud/services/vpc/service.go
```diff
@@ -213,6 +223,45 @@ func (s *Service) GetVPCByName(vpcName string) (*vpcv1.VPC, error) {
...
+		if err != nil {
+			return false, "", err
+		}
+
+		if imagesList == nil {
+			return false, "", fmt.Errorf("image list returned is nil")
+		}
+
+		for i, v := range imagesList.Images {
+			if (*v.Name) == imageName {
```
Karthik-K-N, line 246:
```suggestion
			if *v.Name == imageName {
```

controllers/ibmvpccluster_controller.go
```diff
@@ -245,8 +245,22 @@ func (r *IBMVPCClusterReconciler) reconcileCluster(clusterScope *scope.VPCCluste
...
 		return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
 	}
+	clusterScope.Info("Reconciliation of VPC complete")
 	conditions.MarkTrue(clusterScope.IBMVPCCluster, infrav1beta2.VPCReadyCondition)
 
+	// Reconcile the cluster's VPC Custom Image.
+	clusterScope.Info("Reconciling VPC Custom Image")
+	if requeue, err := clusterScope.ReconcileVPCCustomImage(); err != nil {
+		clusterScope.Error(err, "failed to reconcile VPC Custom Image")
+		conditions.MarkFalse(clusterScope.IBMVPCCluster, infrav1beta2.ImageReadyCondition, infrav1beta2.VPCReconciliationFailedReason, capiv1beta1.ConditionSeverityError, err.Error())
```
Karthik-K-N, line 255: Lets define another reason specifically for CustomImage rather than using VPCReconciliationFailedReason

Karthik-K-N (commented):
Couple small changes, Otherwise LGTM. Thank you.

cloud/scope/vpc_cluster.go
cjschaef: No, I do not want to rearrange to check Spec prior to Status in this way.

I could move `s.IBMVPCCluster.Spec.Image == nil` first, prior to Status check, but that would be the only check I'd want to move prior to Status check.

controllers/ibmvpccluster_controller.go
cjschaef: oops, yes I will add a new reconciliation failed reason.

pkg/cloud/services/vpc/service.go
cjschaef: Sure, then I'd recommend the other `Get*ByName` functions get updated later as well.

dharaneeshvrd:
LGTM!

Karthik-K-N (commented):
/lgtm

mkumatag (approved):

PR #1970, opened 2024-09-18
VPC: Add subnet reconciliation

cjschaef:
Add support to reoncile VPC subnets for the new v2 VPC Infrastructure reoncile logic.

Related: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/1896

Add new support for reconciling VPC subnets for VPC Infra

Fixes #

/area provider/ibmcloud

**Release note**:  

```release-note
Add support to new (v2) reconciliation of Subnets for VPC Infrastructure
```

### DIFF
--- a/cloud/scope/vpc_cluster.go
+++ b/cloud/scope/vpc_cluster.go
@@ -50,6 +50,9 @@ import (
 const (
 	// LOGDEBUGLEVEL indicates the debug level of the logs.
 	LOGDEBUGLEVEL = 5
+
+	// vpcSubnetIPVersion4 defines the IP v4 string used for VPC Subnet generation.
+	vpcSubnetIPVersion4 = "ipv4"
 )
 
 // VPCClusterScopeParams defines the input parameters used to create a new VPCClusterScope.
@@ -307,10 +310,42 @@ func (s *VPCClusterScope) GetServiceName(resourceType infrav1beta2.ResourceType)
 		if s.NetworkSpec().VPC.Name != nil {
 			return s.NetworkSpec().VPC.Name
 		}
+	case infrav1beta2.ResourceTypeSubnet:
+		// Generate a generic subnet name based off the cluster name, which can be extended as necessary (for Zones).
+		return ptr.To(fmt.Sprintf("%s-subnet", s.IBMVPCCluster.Name))
+	case infrav1beta2.ResourceTypePublicGateway:
+		// Generate a generic public gateway name based off the cluster name, which can be extedned as necessary (for Zone).
+		return ptr.To(fmt.Sprintf("%s-pgateway", s.IBMVPCCluster.Name))
 	default:
 		s.V(3).Info("unsupported resource type", "resourceType", resourceType)
 	}
 	return nil
+}
+
+// GetSubnetID returns the ID of a subnet, provided the name.
+func (s *VPCClusterScope) GetSubnetID(name string) (*string, error) {
+	// Check Status first
+	if s.NetworkStatus() != nil {
+		if s.NetworkStatus().ControlPlaneSubnets != nil {
+			if subnet, ok := s.NetworkStatus().ControlPlaneSubnets[name]; ok {
+				return &subnet.ID, nil
+			}
+		}
+		if s.NetworkStatus().WorkerSubnets != nil {
+			if subnet, ok := s.NetworkStatus().WorkerSubnets[name]; ok {
+				return &subnet.ID, nil
+			}
+		}
+	}
+	// Otherwise, if no Status, or not found, attempt to look it up via IBM Cloud API.
+	subnet, err := s.VPCClient.GetVPCSubnetByName(name)
+	if err != nil {
+		return nil, fmt.Errorf("failed subnet id lookup by name %s: %w", name, err)
+	}
+	if subnet == nil {
+		return nil, nil
+	}
+	return subnet.ID, nil
 }
 
 // GetVPCID returns the VPC id, if available.
@@ -377,6 +412,30 @@ func (s *VPCClusterScope) SetResourceStatus(resourceType infrav1beta2.ResourceTy
 			return
 		}
 		s.IBMVPCCluster.Status.Image.Set(*resource)
+	case infrav1beta2.ResourceTypeControlPlaneSubnet:
+		if s.NetworkStatus() == nil {
+			s.IBMVPCCluster.Status.Network = &infrav1beta2.VPCNetworkStatus{}
+		}
+		if s.NetworkStatus().ControlPlaneSubnets == nil {
+			s.IBMVPCCluster.Status.Network.ControlPlaneSubnets = make(map[string]*infrav1beta2.ResourceStatus)
+		}
+		if subnet, ok := s.NetworkStatus().ControlPlaneSubnets[*resource.Name]; ok {
+			subnet.Set(*resource)
+		} else {
+			s.IBMVPCCluster.Status.Network.ControlPlaneSubnets[*resource.Name] = resource
+		}
+	case infrav1beta2.ResourceTypeWorkerSubnet:
+		if s.NetworkStatus() == nil {
+			s.IBMVPCCluster.Status.Network = &infrav1beta2.VPCNetworkStatus{}
+		}
+		if s.NetworkStatus().WorkerSubnets == nil {
+			s.IBMVPCCluster.Status.Network.WorkerSubnets = make(map[string]*infrav1beta2.ResourceStatus)
+		}
+		if subnet, ok := s.NetworkStatus().WorkerSubnets[*resource.Name]; ok {
+			subnet.Set(*resource)
+		} else {
+			s.IBMVPCCluster.Status.Network.WorkerSubnets[*resource.Name] = resource
+		}
 	default:
 		s.V(3).Info("unsupported resource type", "resourceType", resourceType)
 	}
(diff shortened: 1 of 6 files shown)


cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,321 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+		if err != nil {
+			return false, fmt.Errorf("error failed building control plane subnets: %w", err)
+		}
+	} else {
+		subnets = s.IBMVPCCluster.Spec.Network.ControlPlaneSubnets
+	}
+
+	// Reconcile Control Plane subnets.
+	requeue := false
+	for _, subnet := range subnets {
```
Karthik-K-N, line 763: If no inter dependency, In future for performance optimization we can think of reconciling data and control plane subnets concurrently.

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,321 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+	// Add a tag to the subnet for the cluster.
+	err = s.TagResource(s.IBMVPCCluster.Name, *subnetDetails.CRN)
+	if err != nil {
+		return fmt.Errorf("error failed to tag subnet %s: %w", *subnetDetails.Name, err)
+	}
+
+	return nil
+}
+
+// findOrCreatePublicGateway will attempt to find if there is an existing Public Gateway for a specific zone, for the cluster (in cluster's Resource Group and VPC), or create a new one. Only one Public Gateway is required in each zone, for any subnets in that zone.
```
Karthik-K-N, line 1006: For my better understanding, Do you plan to support multiple subnets in same zone, Do you have any use case in mind when we will need to do this.

Karthik-K-N (commented):
Overall LGTM

cloud/scope/vpc_cluster.go
cjschaef: That is likely possible, if subnet creation is found to take longer than desired I can revisit that. I know Custom Image and LB's are the larger time sinks (in terms of waiting for them to be `Ready`, and SG/SGR reconciliation is heavily complex and can take time as well.

cloud/scope/vpc_cluster.go
cjschaef: Yes, the simple path here, is when using different subnets for Control Plane and Data Plane, distributed across the same zones (one CP subnet per zone and one DP subnet per zone).
Otherwise, if the user desires to include edge subnets, or distribute nodes within smaller subnets IP CIDR's, multiple subnets could be used per zone as well.

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,321 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+
+// reconcileSubnet will attempt to find the existing subnet, or create it if necessary.
+// The logic can handle either Control Plane or Worker subnets, but must distinguish between them for Status updates.
+func (s *VPCClusterScope) reconcileSubnet(subnet infrav1beta2.Subnet, isControlPlane bool) (bool, error) { //nolint: gocyclo
+	// If subnet already has a Name defined, use that for lookup.
+	if subnet.Name != nil {
+		// If we have the subnet name, we can easily check Network Stauts on the subnet's status.
+		if isControlPlane {
+			// If we find the subnet in Control Plane subnet status and that it is ready, we can return, with no requeue required.
+			if subnet, ok := s.NetworkStatus().ControlPlaneSubnets[*subnet.Name]; ok && subnet.Ready {
```
dharaneeshvrd, line 807: Even if we find it in status, don't we need to check the cloud that its still in desired state?
Please take care of this in all the similar blocks.

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,321 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+		subnets = append(subnets, infrav1beta2.Subnet{
+			Name: ptr.To(name),
+			Zone: ptr.To(zone),
+		})
+	}
+	return subnets, nil
+}
+
+// checkSubnetStatus will check the status of a IBM Cloud Subnet and update the Network Status.
+func (s *VPCClusterScope) checkSubnetStatus(subnetDetails *vpcv1.Subnet, isControlPlane bool) (bool, error) {
```
dharaneeshvrd, line 904: I feel we need to rename this if you are going to update the status also as part of this func.

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,321 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+		Ready: false,
+	}
+	if isControlPlane {
+		s.SetResourceStatus(infrav1beta2.ResourceTypeControlPlaneSubnet, resourceStatus)
+	} else {
+		s.SetResourceStatus(infrav1beta2.ResourceTypeWorkerSubnet, resourceStatus)
+	}
+
+	// Add a tag to the subnet for the cluster.
+	err = s.TagResource(s.IBMVPCCluster.Name, *subnetDetails.CRN)
```
dharaneeshvrd, line 998: Hope tagging subnet is tested? We have nt tried before.

dharaneeshvrd (changes requested):

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,321 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+		} else if requiresRequeue {
+			// If the reconcile of the subnet requires further reconciliation, plan to requeue entire ReconcileSubnets call, but attempt to further reconcile additional Subnets (attempt all subnet reconciliation).
+			requeue = true
+		}
+	}
+
+	// If no Worker subnets were supplied, attempt to create one in each zone.
+	if s.IBMVPCCluster.Spec.Network.WorkerSubnets == nil || len(s.IBMVPCCluster.Spec.Network.WorkerSubnets) == 0 {
+		// If neither Control Plane nor Worker subnets were supplied, we rely on both Planes using the same subnet per zone, and we will re-reconcile those subnets below, for IBMVPCCluster Status updates.
+		if len(s.IBMVPCCluster.Spec.Network.ControlPlaneSubnets) != 0 {
```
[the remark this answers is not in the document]
cjschaef: I could reword, but no, in this case if some Control Plane subnets were supplied, we will auto-generate some Worker subnets.
If neither got provided, then Worker subnets will use the Control Plane subnets that were auto-generated.

cloud/scope/vpc_cluster.go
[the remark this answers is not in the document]
cjschaef: Yeah, that is a good idea to be safe.

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,321 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+		// If we have the subnet name, we can easily check Network Stauts on the subnet's status.
+		if isControlPlane {
+			// If we find the subnet in Control Plane subnet status and that it is ready, we can return, with no requeue required.
+			if subnet, ok := s.NetworkStatus().ControlPlaneSubnets[*subnet.Name]; ok && subnet.Ready {
+				s.V(3).Info("found subnet is ready", "subnetName", subnet.Name)
+				return false, nil
+			}
+		} else {
+			// If we find the subnet in Worker subnet status and that it is ready, we can return, with no requeue required.
+			if subnet, ok := s.NetworkStatus().WorkerSubnets[*subnet.Name]; ok && subnet.Ready {
```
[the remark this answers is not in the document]
cjschaef: will update.

cloud/scope/vpc_cluster.go
cjschaef: The discussion I had previously, was that with certain VPC resources, if the details are already in Status, checking the realtime status of said resource would not need to happen. For instance, handling cases where CAPI created a resource, and someone/thing deleted the resource, having CAPI attempt to re-reconcile that case, even though Status had a name and ID (which are no longer valid, like ID) would not be desirable. 

As VPC and Subnets, do not typically change status, unlike Load Balancers, the expectation is that once these resources are Ready (or the equivalent VPC status value/constant) it should stay that way until deletion (either requested or done maliciously). Load Balancers, due to `updating` being a frequent occurrence, likely would not be able to rely on previous status (instead checking status in realtime).

If this expectation has changed, and CAPI is expected to recreate all VPC resources (if they disappear, API fails to find them, etc.), I can change all resource logic to no rely on Status details (such as ID) if that changes in realtime.

cloud/scope/vpc_cluster.go
cjschaef: I will change to `update`.

cloud/scope/vpc_cluster.go
cjschaef: Yes it does. I don't perform exhaustive testing, due to time constraints, but I test all the reconciliation logic after each set of changes for the standard path. So tagging works on subnets (based on the API I don't think GlobalTagging cares what a resource is, as long as it has a CRN), even though the UI leaves a lot to be desired in terms of tagging support.
```
# ibmcloud resource search "tags:\"us-east-capi-subnet-1-v2vzf\"" | grep "Name\|Resource Type"
Name:                us-east-capi-subnet-1-v2vzf-pgateway-us-east-1
Resource Type:       public-gateway
Name:                us-east-capi-subnet-1-v2vzf-subnet-compute-us-east-1
Resource Type:       subnet
Name:                us-east-capi-subnet-1-v2vzf-subnet-control-plane-us-east-1
Resource Type:       subnet
Name:                us-east-capi-subnet-1-v2vzf-pgateway-us-east-2
Resource Type:       public-gateway
Name:                us-east-capi-subnet-1-v2vzf-subnet-control-plane-us-east-2
Resource Type:       subnet
Name:                us-east-capi-subnet-1-v2vzf-subnet-compute-us-east-2
Resource Type:       subnet
Name:                us-east-capi-subnet-1-v2vzf-pgateway-us-east-3
Resource Type:       public-gateway
Name:                us-east-capi-subnet-1-v2vzf-subnet-compute-us-east-3
Resource Type:       subnet
Name:                us-east-capi-subnet-1-v2vzf-subnet-control-plane-us-east-3
Resource Type:       subnet
Name:                us-east-capi-subnet-1-v2vzf-rhcos
Resource Type:       image
Name:                us-east-capi-subnet-1-v2vzf-vpc
Resource Type:       vpc
```

cjschaef:
I don't see this error is local linting, and other instances don't appear to fail
```
controllers/ibmvpccluster_controller.go:268:176: printf: non-constant format string in call to sigs.k8s.io/cluster-api/util/conditions.MarkFalse (govet)
		conditions.MarkFalse(clusterScope.IBMVPCCluster, infrav1beta2.VPCSubnetReadyCondition, infrav1beta2.VPCSubnetReconciliationFailedReason, capiv1beta1.ConditionSeverityError, err.Error())
		                                                                                                                                                                             ^
```
Going to assume this is a test flake and get the test to rebuild.

cjschaef:
ah, think I found the error and have fixed it now.

cloud/scope/vpc_cluster.go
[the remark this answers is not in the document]
dharaneeshvrd: Ok got it.

cloud/scope/vpc_cluster.go
dharaneeshvrd: @Karthik-K-N @mkumatag could you please share your opinion here?

Karthik-K-N: We usually follow this pattern of fetching id from status and checking its actual status in cloud, Its to make sure that the observed resource status is correct and up to date with real world.

I think its better to follow this pattern as its helps in early detection of any potential problems.

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,324 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+
+	// Find or create a Public Gateway in this zone for the subnet, only one Public Gateway is required for each zone, for this cluster.
+	// NOTE(cjschaef): We may need to add support to not attach Public Gateways to subnets.
+	publicGateway, err := s.findOrCreatePublicGateway(*subnet.Zone)
+	if err != nil {
+		return fmt.Errorf("error failed to find or create public gateway for subnet %s: %w", *subnet.Name, err)
+	}
+
+	options := &vpcv1.CreateSubnetOptions{}
+	options.SetSubnetPrototype(&vpcv1.SubnetPrototype{
```
Karthik-K-N, line 961: Currently in existing VPC flow we create subnet with these properties https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/blob/83e571fe6e85a6edd94d46b762a1844f9f7dd195/cloud/scope/cluster.go#L219-L236. Do you think its not required to set CIDR block or is it not required when you set TotalIpv4AddressCount?

cjschaef: It seems like the two are mutually exclusive (not completely sure though), so you are able to define one or the other (total IPs or a CIDR).
Given the history we have, I opted to keep using the total IP's for now, but once we consider allowing this to be configurable, we could provide both as options (total IP's or CIDR's).

cloud/scope/vpc_cluster.go
cjschaef: Should reconciliation completely stop (panic) if a resource cannot be found then?
Or should resources be recreated and update Status with new ID's, etc.?

Karthik-K-N: As of now reconciler will throw error. For [reference](https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/blob/2c25714f46ca572018419ebb3cb5ad6fe93793ca/cloud/scope/powervs_cluster.go#L994-L1016). Its upto the user to resolve the error.

cjschaef: I will attempt to refactor, with the expectation that anything in Status that cannot be found in IBM Cloud, etc. results in an continuous error. Other resources will be updated in the same fashion in other PR's then.

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,326 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+				return false, fmt.Errorf("error retrieving existing subnet by name %s: %w", *subnetName, err)
+			} else if subnetDetails == nil {
+				return false, fmt.Errorf("error failed to find existing subnet by name: %s", *subnetName)
+			}
+			return s.updateSubnetStatus(subnetDetails, isControlPlane)
+		}
+	}
+
+	// Otherwise, if these is an ID or name, attempt to lookup the subnet and update status as necessary.
+	if subnet.ID != nil {
```
dharaneeshvrd, line 849: If you found an existing subnet with id or name, wondering how are you planning to differentiate this during delete phase? Just asking for my understanding!
If it via tags, hope all the resources involved are having CRN like you mentioned, to attach the tags.

cjschaef: At this moment, I don't have solid plans for deletion logic. Tagging is going to likely be involved/required to properly handle all of the edge cases involved with bring your own 'resource' (vpc, subnet, image, etc.).
Yes, I have already gone through and tested the tagging required for the resource CAPI is creating, but deletion has not been a focus, given time constraints.

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,326 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+			subnetMap = s.NetworkStatus().ControlPlaneSubnets
+		} else if !isControlPlane && s.NetworkStatus().WorkerSubnets != nil {
+			subnetMap = s.NetworkStatus().WorkerSubnets
+		}
+		// Based on Network Status, setup either the name or ID for lookup of the subnet's current status.
+		if subnet.Name != nil {
+			if _, ok := subnetMap[*subnet.Name]; ok {
+				subnetName = subnet.Name
+			}
+		} else if subnetID != nil && subnet.ID != nil {
```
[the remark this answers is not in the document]
cjschaef: Yes, I'll clean that up.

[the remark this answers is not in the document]
cjschaef: Done

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,326 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+	resourceStatus := &infrav1beta2.ResourceStatus{
+		ID:    *subnetDetails.ID,
+		Name:  subnetDetails.Name,
+		Ready: false,
+	}
+	if isControlPlane {
+		s.SetResourceStatus(infrav1beta2.ResourceTypeControlPlaneSubnet, resourceStatus)
+	} else {
+		s.SetResourceStatus(infrav1beta2.ResourceTypeWorkerSubnet, resourceStatus)
+	}
```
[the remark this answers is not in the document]
cjschaef: And to be honest, it doesn't matter anymore, we will have a followup loop after subnet(s) creation, so we will callback to `updateSubnetStatus` on the next iteration.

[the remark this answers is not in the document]
dharaneeshvrd: No, what I meant is that I don't see a code difference here from line 986 to 1000 comparing to `updateSubnetStatus()` func.
So why can't we just call that func instead of these dup lines?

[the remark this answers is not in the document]
cjschaef: The code diff is `Ready: false` is hardcoded here, compared with the function lookup.

Karthik-K-N (commented):
Overall LGTM

cloud/scope/vpc_cluster.go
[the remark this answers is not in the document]
dharaneeshvrd: Won't that func sets the ready as false if you pass the subnet details from here? Since it won't be available immediately after creation.
IMHO we can just use that func here instead of building and setting the status from here. It looks duplicate to me.
@Karthik-K-N can you please see whether my opinion making sense here?

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,326 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+		if requiresRequeue, err := s.reconcileSubnet(subnet, true); err != nil {
+			return false, fmt.Errorf("error failed reconciling control plane subnet: %w", err)
+		} else if requiresRequeue {
+			// If the reconcile of the subnet requires further reconciliation, plan to requeue entire ReconcileSubnets call, but attempt to further reconcile additional Subnets (attempt all subnet reconciliation).
+			requeue = true
+		}
+	}
+
+	// If no Worker subnets were supplied, attempt to create one in each zone.
+	if s.IBMVPCCluster.Spec.Network.WorkerSubnets == nil || len(s.IBMVPCCluster.Spec.Network.WorkerSubnets) == 0 {
```
mkumatag, line 773: same as above, remove the nil check

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,326 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+			return false, fmt.Errorf("error retrieving subnet by name %s: %w", *subnet.Name, err)
+		} else if subnetDetails != nil {
+			// Update status if subnet was found.
+			return s.updateSubnetStatus(subnetDetails, isControlPlane)
+		}
+		// If subnet was not found, expect that it needs to be created.
+	} else {
+		// If no ID or name was provided, that is an error to be raised. One or the other must be specified when subnets are supplied.
+		return false, fmt.Errorf("error subnet has no defined id or name, one is required")
+	}
```
mkumatag, line 874: this can be checked in the beginning and return early failure

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,326 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+		}
+		return s.updateSubnetStatus(subnetDetails, isControlPlane)
+	} else if subnet.Name != nil {
+		// Attempt to check if a subnet exists with the name and update status as necessary.
+		subnetDetails, err := s.VPCClient.GetVPCSubnetByName(*subnet.Name)
+		if err != nil {
+			return false, fmt.Errorf("error retrieving subnet by name %s: %w", *subnet.Name, err)
+		} else if subnetDetails != nil {
+			// Update status if subnet was found.
+			return s.updateSubnetStatus(subnetDetails, isControlPlane)
```
mkumatag, line 868: `updateSubnetStatus` called multiple times, lets optimise later.

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,326 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+		return fmt.Errorf("error retrieving resource group id for resource group %s", s.IBMVPCCluster.Spec.ResourceGroup)
+	}
+
+	vpcID, err := s.GetVPCID()
+	if err != nil {
+		return fmt.Errorf("error retrieving vpc id for subnet creation: %w", err)
+	}
+
+	// TODO(cjschaef): Move to webhook validation.
+	if subnet.Zone == nil {
```
mkumatag, line 945: check and return early

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,326 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
+		return fmt.Errorf("error subnet zone must be defined for subnet %s", *subnet.Name)
+	}
+
+	// NOTE(cjschaef): We likely will want to add support to use custom Address Prefixes
+	// For now, we rely on the API to assign us prefixes, as we request via IP count
+	var ipCount int64 = 256
+	// We currnetly only support IPv4
+	ipVersion := "ipv4"
```
mkumatag, line 953: can this be moved to constant section in the debugging of the file?

mkumatag (changes requested):

cloud/scope/vpc_cluster.go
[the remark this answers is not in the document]
Karthik-K-N:
```suggestion
	if subnetDetails == nil || subnetDetails.ID == nil || subnetDetails.CRN == nil {
		return fmt.Errorf("error failed creating subnet: %s", *subnet.Name)
	}

	return s.updateSubnetStatus(subnetDetails, isControlPlane)
```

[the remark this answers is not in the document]
Karthik-K-N: I am not sure whats will be the value for these `subnetDetails`
```
 subnetDetails.Status != nil && *subnetDetails.Status == string(vpcv1.SubnetStatusAvailableConst) 
```
for the newly created subnet, but  if either  subnetDetails.Status is nil or subnetDetails.Status not vpcv1.SubnetStatusAvailableConst then anyhow we are setting ready as false. so we can reuse I believe.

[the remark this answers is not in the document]
dharaneeshvrd: If create succeeds, it would populate and give subnetDetails properly or else it will throw an error.
Not sure in what case it would return err as nil and subnetDetails also as nil.

cloud/scope/vpc_cluster.go
```diff
@@ -307,12 +307,44 @@ func (s *VPCClusterScope) GetServiceName(resourceType infrav1beta2.ResourceType)
...
+			}
+		}
+	}
+	// Otherwise, if no Status, or not found, attempt to look it up via IBM Cloud API.
+	subnet, err := s.VPCClient.GetVPCSubnetByName(name)
+	if err != nil {
+		return nil, err
+	}
+	if subnet == nil {
+		return nil, nil
```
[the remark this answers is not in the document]
cjschaef: updated

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,326 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
...
 	return ptr.To(href), nil
 }
+
+// ReconcileSubnets reconciles the VPC Subnet(s).
+// For Subnets, we collect all of the required subnets, for each Plane, and reconcile them individually. Requeing if one is missing or just created. Reconciliation is attempted on all subnets each loop, to prevent single subnet creation per reconciliation loop.
+func (s *VPCClusterScope) ReconcileSubnets() (bool, error) {
+	var subnets []infrav1beta2.Subnet
+	var err error
+	// If no ControlPlane Subnets were supplied, we default to create one in each availability zone of the region.
+	if s.IBMVPCCluster.Spec.Network.ControlPlaneSubnets == nil || len(s.IBMVPCCluster.Spec.Network.ControlPlaneSubnets) == 0 {
```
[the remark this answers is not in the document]
cjschaef: updated

cloud/scope/vpc_cluster.go
```diff
@@ -686,3 +742,326 @@ func (s *VPCClusterScope) buildCOSObjectHRef() (*string, error) {
 	s.V(3).Info("building image ref", "href", href)
 	return ptr.To(href), nil
 }
+
+// ReconcileSubnets reconciles the VPC Subnet(s).
+// For Subnets, we collect all of the required subnets, for each Plane, and reconcile them individually. Requeing if one is missing or just created. Reconciliation is attempted on all subnets each loop, to prevent single subnet creation per reconciliation loop.
+func (s *VPCClusterScope) ReconcileSubnets() (bool, error) {
+	var subnets []infrav1beta2.Subnet
+	var err error
```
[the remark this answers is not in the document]
cjschaef: I believe I need this declared here, so that when assigning `subnets` below, it is not redefined. The linting at least requires the declarations first, before assigning.

cloud/scope/vpc_cluster.go
cjschaef: updated

cloud/scope/vpc_cluster.go
cjschaef: Okay, I think that is a safe expectation. updated

cloud/scope/vpc_cluster.go
cjschaef: updated

cloud/scope/vpc_cluster.go
cjschaef: Sure, I moved to consts.

mkumatag (approved):
/lgtm

PR #1972, opened 2024-09-19
ServiceEndpoint overrides saved and used correctly for ignition URL

miyamotoh:
Tested COS service endpoint override, found some places that need additional code, and this PR ensures the overridden COS endpoint is indeed used for bootstrap ignition. Verified the bootstrap VM boots up with the ignition on the custom endpoint specified in `install-config.yaml` like below;
```
serviceEndpoints:
    - name: cos
      url: https://s3.direct.jp-osa.cloud-object-storage.appdomain.cloud
```
(Note above `.direct` in URL is specifically for a Disconnected scenario, and a VPE allowing the traffic needs to be in place beforehand.)

### DIFF
--- a/cloud/scope/powervs_machine.go
+++ b/cloud/scope/powervs_machine.go
@@ -242,6 +242,7 @@ func NewPowerVSMachineScope(params PowerVSMachineScopeParams) (scope *PowerVSMac
 
 	scope.IBMVPCClient = vpcClient
 	scope.ResourceClient = rc
+	scope.ServiceEndpoint = params.ServiceEndpoint
 	return scope, nil
 }
 
@@ -426,13 +427,19 @@ func (m *PowerVSMachineScope) createIgnitionData(data []byte) (string, error) {
 	cosServiceEndpoint := endpoints.FetchEndpoints(string(endpoints.COS), m.ServiceEndpoint)
 	if cosServiceEndpoint != "" {
 		m.Logger.V(3).Info("Overriding the default COS endpoint in ignition URL", "cosEndpoint", cosServiceEndpoint)
-		objHost = fmt.Sprintf("%s.%s", bucket, cosServiceEndpoint)
+		cosURL, _ := url.Parse(cosServiceEndpoint)
+		if cosURL.Scheme != "" {
+			objHost = fmt.Sprintf("%s.%s", bucket, cosURL.Host)
+		} else {
+			objHost = fmt.Sprintf("%s.%s", bucket, cosServiceEndpoint)
+		}
 	}
 	objectURL := &url.URL{
 		Scheme: "https",
 		Host:   objHost,
 		Path:   key,
 	}
+	m.Logger.V(3).Info("Generated Ignition URL", "objectURL", objectURL.String())
 
 	return objectURL.String(), nil
 }


cloud/scope/powervs_machine.go
```diff
@@ -424,15 +425,20 @@ func (m *PowerVSMachineScope) createIgnitionData(data []byte) (string, error) {
...
+	var objectURL *url.URL
 	if cosServiceEndpoint != "" {
 		m.Logger.V(3).Info("Overriding the default COS endpoint in ignition URL", "cosEndpoint", cosServiceEndpoint)
-		objHost = fmt.Sprintf("%s.%s", bucket, cosServiceEndpoint)
-	}
-	objectURL := &url.URL{
-		Scheme: "https",
-		Host:   objHost,
-		Path:   key,
+		objectURL, _ = url.Parse(cosServiceEndpoint)
```
mkumatag, line 431: This code doesn't work if user refers the https://cloud.ibm.com/docs/cloud-object-storage?topic=cloud-object-storage-endpoints link for the endpoints and mentioned the URL

e.g: 

```golang
	cosServiceEndpoint := "s3.direct.jp-osa.cloud-object-storage.appdomain.cloud"
	u, err := url.Parse(cosServiceEndpoint)
	if err != nil {
		fmt.Printf("%+v", err)
	}
	u.Host = fmt.Sprintf("%s.%s", "bucket", u.Host)
	fmt.Printf("host: %s\n", u.Host)       // host: bucket.
	fmt.Printf("path: %s\n", u.Path)       // path: s3.direct.jp-osa.cloud-object-storage.appdomain.cloud
	fmt.Printf("string: %s\n", u.String()) // string: //bucket./s3.direct.jp-osa.cloud-object-storage.appdomain.cloud
```

mkumatag (changes requested):

cloud/scope/powervs_machine.go
miyamotoh: If the endpoint value doesn't start with some kind of protocol (ie. `https://`), it'd be rejected much earlier, and wouldn't reach this logic. Isn't that good enough?

mkumatag: but both cosURLDomain and above URL contains the URL without https, so did you test it with without https in the installer checked if it is working or not?

miyamotoh: I started without `https://`, and it never got to this code. Installer was failing sooner, before reaching this ignition part, I think it was in `Cluster` scope, instead of `Machine` scope like above. If you can think of a way to specify this override without `https://` and still reach this ignition URL logic, please let me know.

mkumatag: @dharaneeshvrd @Karthik-K-N ptal at the issue what @miyamotoh is mentioning..

Karthik-K-N: In openshift case we have a validation added for ServiceEndpoints to check for protocal for serviceednpoints for reference https://github.com/Karthik-K-N/api/blob/master/config/v1/types_infrastructure.go#L1118-L1121.

But in normal cluster-api workflow we dont have any such restrictions, so our goal should be to support both type of URLs

Karthik-K-N: How about checking whether the cosServiceEndpoint has scheme set or not and parse the URL only when scheme is set?

```
if cosServiceEndpoint != "" {
	m.Logger.V(3).Info("Overriding the default COS endpoint in ignition URL", "cosEndpoint", cosServiceEndpoint)
	cosURL, _ := url.Parse(cosServiceEndpoint)
	if cosURL.Scheme != "" {
		objHost = fmt.Sprintf("%s.%s", "bucket", cosURL.Host)
	} else {
		objHost = fmt.Sprintf("%s.%s", "bucket", cosServiceEndpoint)
	}
	}


```

mkumatag: @miyamotoh can you please incorporate this ^^^

miyamotoh: Done, but full disclosure, I have not been able to test this code in the installer at this time, due to other issues. Will report back when it's verified working.

mkumatag (approved):
/lgtm
/approve
/ok-to-test
/hold

remove the hold when you finish your testing

Karthik-K-N (commented):
I think we need to update here as well, Please take a look: https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/blob/af008bdb71008976f73d16146b52bfd037f4d166/cloud/scope/powervs_machine.go#L606

miyamotoh:
Finally I was able to verify a private cluster can be deployed successfully in PowerVS with the proposed changes. Proceeding...

/unhold

Issue #1977, opened 2024-09-25
Extend VPC (x86_64) Machine reconciliation support


cjschaef:
/kind feature
/area provider/ibmcloud

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
The current implementation for reconciling VPC (x86_64) Machine support lacks full feature/resource support for creating VPC VSI's (Instances/Machines) to be used for clusters. Expending the support for VPC to properly reconcile instances and in more complex configurations can help with a more complete set of Machines for clusters.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
As the current implementation provides some feature support, we expect to try to add additional configuration options and management in optional fields and code paths to prevent affecting current implementation. Due to this complexity, it is likely going to take a few rounds of refactoring and patches to resolve all the requirements we expect.

---

## Related changes
- [x] [Machine API updates](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1978)
- [x] [Machine Reconciliation updates](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2042)
- [x] [Fix Machine Load Balancer Pool Member deletion](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2051)

cjschaef:
The primary functionality for the extended VPC Machine reconciliation has been completed.

Closing this as done.

PR #1978, opened 2024-09-27
VPC: Extend VPC Machine API's

cjschaef:
Extend the VPC Machine API's to include additional fields and types for more VPC Machine configuration support.

Related: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/1977

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Extend VPCMachine APIs to allow for more configuration and supported features. These are intended to be used with the extended VPC Cluster Infrastructure support.
```


api/v1beta2/ibmvpcmachine_types.go
```diff
@@ -36,10 +39,23 @@ type IBMVPCMachineSpec struct {
 	// Name of the instance.
 	Name string `json:"name,omitempty"`
 
+	// CatalogOffering is the Catalog Offering OS image which would be installed on the instance.
+	// An OfferingCRN or VersionCRN is required, the PlanCRN is optional.
+	// +optional
+	CatalogOffering *IBMCloudCatalogOffering `json:"catalogOffering,omitempty"`
```
Karthik-K-N, line 45: For my understanding, how would this respective image would be imported to VPC or the global catalog images are generally available to all the VPC?

Karthik-K-N: Alos since this catalogOffering is specific to os image should we rename the member  to indicate that?

api/v1beta2/ibmvpcmachine_types.go
```diff
@@ -36,10 +39,23 @@ type IBMVPCMachineSpec struct {
...
+	// +optional
+	DedicatedHost *VPCResource `json:"dedicatedHost,omitempty"`
+
 	// Image is the OS image which would be install on the instance.
 	// ID will take higher precedence over Name if both specified.
 	Image *IBMVPCResourceReference `json:"image"`
 
+	// LoadBalancerPoolMembers is the set of IBM Cloud VPC Load Balancer Backend Pools the machine should be added to as a member.
+	// +optional
+	LoadBalancerPoolMembers []VPCLoadBalancerBackendPoolMember `json:"loadBalancerPoolMembers,omitempty"`
```
Karthik-K-N, line 57: We used to attach the machine to all the loadbalancers defined in cluster.Loadbalancer, Do you plan to avoid that and register to only loadbalancers set here?

What  will happen when this is omitted?

api/v1beta2/ibmvpcmachine_types.go
```diff
@@ -36,10 +39,23 @@ type IBMVPCMachineSpec struct {
...
 	Name string `json:"name,omitempty"`
 
+	// CatalogOffering is the Catalog Offering OS image which would be installed on the instance.
+	// An OfferingCRN or VersionCRN is required, the PlanCRN is optional.
+	// +optional
+	CatalogOffering *IBMCloudCatalogOffering `json:"catalogOffering,omitempty"`
+
+	// DedicatedHost is the name of the underlying provisioned host in the VPC on which the instance will be created.
+	// +optional
+	DedicatedHost *VPCResource `json:"dedicatedHost,omitempty"`
```
Karthik-K-N, line 49: For optional field for our better understanding lets mention what will happen when the field is omited.

api/v1beta2/ibmvpcmachine_types.go
```diff
@@ -126,18 +142,44 @@ type IBMVPCMachineStatus struct {
...
+	FailureMessage *string `json:"failureMessage,omitempty"`
+
+	// InstanceStatus is the status of the IBM Cloud instance for this machine.
 	// +optional
 	InstanceStatus string `json:"instanceState,omitempty"`
+
+	// LoadBalancerPoolMembers is the status of IBM Cloud VPC Load Balancer Backend Pools the machine is a member.
+	// +optional
+	// +k8s:conversion-gen=false
+	LoadBalancerPoolMembers []VPCLoadBalancerBackendPoolMember `json:"loadBalancerPoolMembers,omitempty"`
```
Karthik-K-N, line 182: Is it required here? Is it just to verify that the machine is already registered to LB?

api/v1beta2/types.go
```diff
@@ -356,12 +356,53 @@ type IBMCloudResourceReference struct {
...
+	// +required
+	LoadBalancer VPCResource `json:"loadBalancer"`
+
+	// Pool defines the Load Balancer Pool the Pool Member should be in.
+	// +required
+	Pool VPCResource `json:"pool"`
+
+	// Port defines the Port the Load Balancer Pool Member listens for traffic.
+	// +required
+	Port int64 `json:"port"`
```
Karthik-K-N, line 399: Any reasons for defining Port as int64 and wieght as pointer?

api/v1beta2/ibmvpcmachine_types.go
cjschaef: Sure, I can rename it.
In the cases I have used, this documentation provides details on creating a Private Catalog:  https://cloud.ibm.com/docs/account?topic=account-restrict-by-user&interface=ui
And details on creating a sharable VPC Custom Image:  https://cloud.ibm.com/docs/vpc?topic=vpc-custom-image-cloud-private-catalog&interface=ui

api/v1beta2/ibmvpcmachine_types.go
cjschaef: The current code has some hardcoded values and expectations on attaching Machines to LB's.
I am trying to avoid doing that, especially since it is currently hardcoded to one LB.
I would expect that all machines CAPI creates then would be added into LB pools then, if we rely on Cluster definitions, which I do not expect is valid. Is that a fair assumption, if we rely on the Cluster resource?

api/v1beta2/ibmvpcmachine_types.go
cjschaef: Yes, that is the intention, to track and make sure the Machine is in the required pools.

api/v1beta2/types.go
cjschaef: Yes, Port is required for LB member creation, for which we do not expect to always use `6443`.
And weight is not required, hence it is made optional.

mkumatag (changes requested):

api/v1beta2/ibmvpcmachine_types.go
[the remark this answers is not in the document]
cjschaef: Yes, that is the expected field. We will need to keep parity for allowing `DedicateHostIdentities` to be supported for Machines (VSI's).

api/v1beta2/ibmvpcmachine_types.go
```diff
@@ -36,10 +39,23 @@ type IBMVPCMachineSpec struct {
 	// Name of the instance.
 	Name string `json:"name,omitempty"`
 
+	// CatalogOffering is the Catalog Offering OS image which would be installed on the instance.
+	// An OfferingCRN or VersionCRN is required, the PlanCRN is optional.
```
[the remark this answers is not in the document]
cjschaef: I provided what details are available (not many) in regard to using an offering in place of a Custom Image
This is what I had to piece together to create VSI's (instances) with an offering instead
https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1978#discussion_r1782981167

api/v1beta2/ibmvpcmachine_types.go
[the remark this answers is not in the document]
mkumatag: well, this is not limited to Host right, I see following combinations - 
- dedicated host
- dedicated host group
- placement group

So instead of making this field just to host, lets make it more generic so that user can consume for any of the above options..

Correct me if I'm missing something here!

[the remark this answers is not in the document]
cjschaef: I refactored to support the three options for placement target now.

mkumatag (changes requested):

api/v1beta2/types.go
```diff
@@ -356,12 +356,68 @@ type IBMCloudResourceReference struct {
...
+	// +optional
+	VersionCRN *string `json:"versionCRN,omitempty"`
+}
+
 // NetworkInterface holds the network interface information like subnet id.
 type NetworkInterface struct {
+	// SecurityGroups defines a set of IBM Cloud VPC Security Groups to attach to the network interface.
+	// +optional
+	// +k8s:conversion-gen=false
+	SecurityGroups []VPCResource `json:"securityGroups,omitempty"`
```
mkumatag, line 381: Adding his field throwing error in the apidiff CI job - https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-ibmcloud/1978/pull-cluster-api-provider-ibmcloud-apidiff/1842074223984513024

api/v1beta2/ibmvpcmachine_types.go
```diff
@@ -30,16 +33,29 @@ const (
...
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of machine.
 	// Important: Run "make" to regenerate code after modifying this file
 
 	// Name of the instance.
 	Name string `json:"name,omitempty"`
 
+	// CatalogOfferingImage is the Catalog Offering OS image which would be installed on the instance.
+	// An OfferingCRN or VersionCRN is required, the PlanCRN is optional.
+	// +optional
+	CatalogOfferingImage *IBMCloudCatalogOffering `json:"catalogOfferingImage,omitempty"`
```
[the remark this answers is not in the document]
cjschaef: I have remove Image from the field.

api/v1beta2/types.go
```diff
@@ -356,12 +356,68 @@ type IBMCloudResourceReference struct {
...
+	// +optional
+	Weight *int64 `json:"weight,omitempty"`
+}
+
+// VPCMachinePlacementTarget represents a VPC Machine's placement restrictions.
+// +kubebuilder:validation:XValidation:rule="(has(self.dedicatedHost) && !has(self.dedicatedHostGroup) && !has(self.placementGroup)) || (!has(self.dedicatedHost) && has(self.dedicatedHostGroup) && !has(self.placementGroup)) || (!has(self.dedicatedHost) && !has(self.dedicatedHostGroup) && has(self.placementGroup))",message="only one of dedicatedHost, dedicatedHostGroup, or placementGroup must be defined for machine placement"
+type VPCMachinePlacementTarget struct {
+	// DedicatedHost defines the Dedicated Host to place a VPC Machine (Instance) on.
+	// +optional
+	DedicatedHost *VPCResource `json:"dedicatedHost,omitempty"`
```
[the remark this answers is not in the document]
cjschaef: The expectation is that the user would provide an ID or a name of the existing resource, and a lookup would be performed for those resources (to retrieve the ID if not supplied) and use that in the API calls, for whichever of the 3 was provided. Each of these three options has a Name or ID field, which the user could provide, simplifying the user experience.

api/v1beta2/ibmvpcmachine_types.go
```diff
@@ -126,18 +142,44 @@ type IBMVPCMachineStatus struct {
...
 	Ready bool `json:"ready"`
 
-	// Addresses contains the GCP instance associated addresses.
+	// Addresses contains the IBM Cloud instance associated addresses.
 	Addresses []corev1.NodeAddress `json:"addresses,omitempty"`
 
-	// InstanceStatus is the status of the GCP instance for this machine.
+	// Conditions deefines current service state of the IBMVPCMachine.
+	// +optional
+	// +k8s:conversion-gen=false
```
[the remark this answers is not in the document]
cjschaef: The generator could not perform conversion on this field. I added this field instead.

However, I retried the manual conversion this type and it appears it made linting and everything happy this time.
Hopefully this actually works in the CI checks.

api/v1beta2/types.go
cjschaef: Per [comment above](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/1978#discussion_r1787841017), I retried the manual conversion, so hopefully that resolves this failure now.

mkumatag (approved):
/lgtm

/hold

@Karthik-K-N unhold when you are okay

api/v1beta2/ibmvpcmachine_types.go
Karthik-K-N: We can do this way as well, but my point is eventually the lb referred by VPCLoadBalancerBackendPoolMember.Loadbalancer is created by cluster reconciler, So why can't we directly use it.

Karthik-K-N (commented):
/lgtm

/hold cancel

PR #2023, opened 2024-10-22
Vpc reconcile lbs

cjschaef:
**What this PR does / why we need it**:  Add support to reconcile extended VPC Load Balancers

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Add reconciliation support for extended VPC Load Balancer definitions
```

### DIFF
--- a/cloud/scope/vpc_cluster.go
+++ b/cloud/scope/vpc_cluster.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"reflect"
 
 	"github.com/go-logr/logr"
@@ -54,6 +55,11 @@ const (
 
 	// vpcSubnetIPVersion4 defines the IP v4 string used for VPC Subnet generation.
 	vpcSubnetIPVersion4 = "ipv4"
+
+	// privateLBSuffix is used to tag a default Load Balancer name as private.
+	privateLBSuffix = "private"
+	// publicLBSuffix is used to tag a default Load Balancer name as public.
+	publicLBSuffix = "public"
 )
 
 // VPCClusterScopeParams defines the input parameters used to create a new VPCClusterScope.
@@ -221,6 +227,93 @@ func (s *VPCClusterScope) CheckTagExists(tagName string) (bool, error) {
 	return exists != nil, nil
 }
 
+// GetAPIServerPort will return the API Server's port.
+func (s *VPCClusterScope) GetAPIServerPort() int32 {
+	// TODO(cjschaef): Add logic to handle cases not default.
+	return infrav1beta2.DefaultAPIServerPort
+}
+
+// GetControlPlaneSubnetIDs returns all of the Control Plane subnet Id's.
+func (s *VPCClusterScope) GetControlPlaneSubnetIDs() ([]string, error) {
+	subnets := make([]string, 0)
+	// Retrieve the subnet Id's from Status.
+	if s.NetworkStatus() != nil && s.NetworkStatus().ControlPlaneSubnets != nil {
+		for _, subnet := range s.NetworkStatus().ControlPlaneSubnets {
+			subnets = append(subnets, subnet.ID)
+		}
+		// NOTE(cjschaef): We assume all Subnets are in Status at this point, we could perhaps reconcile Status with any defined in Spec (preventing duplicates) to be safe.
+		return subnets, nil
+	}
+
+	// NOTE(cjschaef): If Status was not set or ControlPlaneSubnets was empty, the Control Plane subnet ID's could be retrieved from Spec. However, for now consider this an error, since Subnet reconciliation should have run prior and no tracked Control Plane subnets would be a major issue.
+	return subnets, fmt.Errorf("error no control plane subnets available in status")
+}
+
+// GetLoadBalancerHostName will return the hostname of the cluster's public Load Balancer, assuming only one public Load Balancer was provided. Or, the hostname of the single private Load Balancer (assuming the cluster has no public access and only one private Load Balancer was provided).
+// This function has a very hard assumption that all Load Balancers have been reconciled within Status (and not just some).
+// NOTE(cjschaef): A webhook validation check could help ensure this.
+func (s *VPCClusterScope) GetLoadBalancerHostName() (*string, error) {
+	// If no Status or Load Balancer Status is populated, assume the Load Balancer's are not ready (have not been reconciled), so no hostname will be available.
+	if s.NetworkStatus() == nil || s.NetworkStatus().LoadBalancers == nil || len(s.NetworkStatus().LoadBalancers) == 0 {
+		return nil, nil
+	}
+
+	// If there is only one Load Balancer in Status, return the hostname.
+	// This heavily assumes all Load Balancers have been reconciled and are in Status.
+	if len(s.NetworkStatus().LoadBalancers) == 1 {
+		for _, lb := range s.NetworkStatus().LoadBalancers {
+			// There should only be one key-value pair in the map.
+			return lb.Hostname, nil
+		}
+	}
+
+	// If no Load Balancer's were defined, return an error, as a Load Balancer must be defined (no default Load Balancer is supported currently).
+	if len(s.NetworkSpec().LoadBalancers) == 0 {
+		return nil, fmt.Errorf("error no load balancers defined for cluster")
+	}
+
+	// Otherwise, if more than one Load Balancer was provided, attempt to use the public Load Balancer's hostname.
+	// TODO(cjschaef): A webhook valiation check could guarantee only one public Load Balancer gets defined, as this will simply return the first public Load Balancer (currently only support one public Load Balancer being defined).
+	for _, loadBalancer := range s.NetworkSpec().LoadBalancers {
+		// Check if the Load Balancer is not public (by default it is, when Public is not defined).
+		// This heavily assumes there is only be one public Load Balancer.
+		if loadBalancer.Public != nil && !*loadBalancer.Public {
+			continue
+		}
+
+		// If an ID was provided in Spec, try to find that within Status.
+		if loadBalancer.ID != nil {
+			if lb, ok := s.NetworkStatus().LoadBalancers[*loadBalancer.ID]; ok {
+				return lb.Hostname, nil
+			}
+			return nil, fmt.Errorf("error defined load balancer not found in status: %s", *loadBalancer.ID)
+		}
+
+		// If the defined Load Balancer name was not supplied (empty), assume one was created using the default service name format (with type suffix).
+		// This heavily assumes only two Load Balancers maximum can be supplied (one public and one private) at this time.
+		name := loadBalancer.Name
+		if name == "" {
+			lbSuffix := publicLBSuffix
+			if loadBalancer.Public != nil && !*loadBalancer.Public {
+				lbSuffix = privateLBSuffix
+			}
+			name = fmt.Sprintf("%s-%s", *s.GetServiceName(infrav1beta2.ResourceTypeLoadBalancer), lbSuffix)
+		}
+
+		// Retrieve the Load Balancer hostname from API.
+		lbDetails, err := s.VPCClient.GetLoadBalancerByName(name)
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving load balancer hostname for %s: %w", name, err)
+		} else if lbDetails == nil {
+			return nil, fmt.Errorf("error retrieving load balancer hostname, %s load balancer not found", name)
+		}
+		return lbDetails.Hostname, nil
+	}
+
+	// If no public Load Balancer or more than one private Load Balancer was found in Spec, expect that a proper Load Balancer was not specified (a default public Load Balancer isn't supported), or cannot be determined.
+	return nil, fmt.Errorf("error no valid load balancer found to retrieve hostname")
+}
+
 // GetNetworkResourceGroupID returns the Resource Group ID for the Network Resources if it is present. Otherwise, it defaults to the cluster's Resource Group ID.
 func (s *VPCClusterScope) GetNetworkResourceGroupID() (string, error) {
 	// Check if the ID is available from Status first.
@@ -300,6 +393,25 @@ func (s *VPCClusterScope) GetResourceGroupID() (string, error) {
 	return *resourceGroup.ID, nil
 }
 
+// GetSecurityGroupID returns the ID of a security group, provided the name.
+// This will first check Status for the Security Group (by name), but as the Security Group may not be tracked by CAPI, a lookup of the Security Group by name is made via the VPC API.
+func (s *VPCClusterScope) GetSecurityGroupID(name string) (*string, error) {
+	// Check Status first.
+	if id := s.getSecurityGroupIDFromStatus(name); id != nil {
+		return id, nil
+	}
+
+	// Otherwise, if no Status, or not found, attempt to look it up via VPC API.
+	securityGroup, err := s.VPCClient.GetSecurityGroupByName(name)
+	if err != nil {
+		return nil, err
+	}
+	if securityGroup == nil {
+		return nil, nil
+	}
+	return securityGroup.ID, nil
+}
+
 func (s *VPCClusterScope) getSecurityGroupIDFromStatus(name string) *string {
 	if s.NetworkStatus() != nil && s.NetworkStatus().SecurityGroups != nil {
 		if sg, ok := s.NetworkStatus().SecurityGroups[name]; ok {
@@ -328,6 +440,12 @@ func (s *VPCClusterScope) GetServiceName(resourceType infrav1beta2.ResourceType)
 	case infrav1beta2.ResourceTypePublicGateway:
 		// Generate a generic public gateway name based off the cluster name, which can be extedned as necessary (for Zone).
 		return ptr.To(fmt.Sprintf("%s-pgateway", s.IBMVPCCluster.Name))
+	case infrav1beta2.ResourceTypeLoadBalancer:
+		// Generate a generic load balancer name based off the cluster name, which can be extended as necessary (for public vs private).
+		return ptr.To(fmt.Sprintf("%s-lb", s.IBMVPCCluster.Name))
+	case infrav1beta2.ResourceTypeLoadBalancerPool:
+		// Generate a generic load balancer pool name based off the cluster name, which can be extended as necessary (for LB).
+		return ptr.To(fmt.Sprintf("%s-lbpool", s.IBMVPCCluster.Name))
 	default:
 		s.V(3).Info("unsupported resource type", "resourceType", resourceType)
 	}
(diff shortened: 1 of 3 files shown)


controllers/ibmvpccluster_controller.go
```diff
@@ -287,10 +287,22 @@ func (r *IBMVPCClusterReconciler) reconcileCluster(clusterScope *scope.VPCCluste
...
+	} else if requeue {
+		clusterScope.Info("Load Balancers creation is pending, requeueing")
+		return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
+	}
+	clusterScope.Info("Reconciliation of Load Balancers complete")
+	conditions.MarkTrue(clusterScope.IBMVPCCluster, infrav1beta2.LoadBalancerReadyCondition)
 
 	// Mark cluster as ready.
 	clusterScope.IBMVPCCluster.Status.Ready = true
+	clusterScope.Info("cluster infrastructure is now ready for cluster", "clusterName", clusterScope.IBMVPCCluster.Name)
```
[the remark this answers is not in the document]
cjschaef: Okay, I will add the necessary logic to retrieve those values to set.

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1711,340 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
 
 	return remotePrototype, nil
 }
+
+// ReconcileLoadBalancers reconciles Load Balancers.
+func (s *VPCClusterScope) ReconcileLoadBalancers() (bool, error) {
+	// TODO(cjschaef): Determine if we want to use default LB configuration or require at least one is defined in Cluster spec.
```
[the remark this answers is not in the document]
cjschaef: Sure, I'll add another comment.

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1711,340 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+		}
+		// Assume a new Load Balancer will not be ready immediately, due to the complexity and time it takes.
+		requeue = true
+	}
+	return requeue, nil
+}
+
+// isLoadBalancerReady checks the state of a Load Balancer.
+// If state is active, true is returned (no requeue required).
+// In all other cases, it returns false, indicating a requeue for reconciliation.
```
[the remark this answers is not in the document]
cjschaef: Okay, I'll only mention return values.

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1711,340 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+		lbStatus, err := s.getLoadBalancer(loadBalancer)
+		if err != nil {
+			return false, fmt.Errorf("error retrieving load balancer: %w", err)
+		}
+
+		// If the Load Balancer was found, update Status and move on.
+		if lbStatus != nil {
+			s.SetLoadBalancerStatus(lbStatus)
+			// If the Load Balancer status isn't ready, flag for requeue and continue to next Load Balancer.
+			if isReady := s.isLoadBalancerReady(string(lbStatus.State)); !isReady {
```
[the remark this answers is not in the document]
cjschaef: I can move things around.

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1711,340 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+		var detailedResponse *core.DetailedResponse
+		getLBOptions := &vpcv1.GetLoadBalancerOptions{
+			ID: lb.ID,
+		}
+		loadBalancer, detailedResponse, err = s.VPCClient.GetLoadBalancer(getLBOptions)
+		if detailedResponse != nil && detailedResponse.StatusCode == http.StatusNotFound {
+			return nil, nil
+		}
+	} else {
+		loadBalancer, err = s.VPCClient.GetLoadBalancerByName(lb.Name)
```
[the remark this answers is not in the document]
cjschaef: No, name cannot be `nil` for a LB based on the spec.

So, my assumption based on that, was if ID is not set, we expect the Name to be set.
If there is a use case to have `ID == nil && Name == ""`, I can another check to try to use a default LB name instead.

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1711,340 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+			s.V(3).Info("addd listener to load balancer", "loadBalancerName", loadBalancer.Name, "listenerPort", listener.Port)
+			listeners = append(listeners, listener)
+		}
+	} else {
+		s.V(3).Info("using default listeners for load balancer", "loadBalancerName", loadBalancer.Name)
+		listeners = append(listeners, s.getDefaultLoadBalancerListeners(loadBalancer.BackendPools == nil)...)
+	}
+	options.SetListeners(listeners)
+
+	// Create the load balancer.
```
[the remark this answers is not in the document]
cjschaef: I can add a log call with v(5) with the options object prior to creation....unless you would rather have each field logged out separately in the log call?

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1711,340 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+	// Only apply a name if one was provided (otherwise rely on generated name from VPC service).
+	if pool.Name != nil {
+		backendPool.Name = pool.Name
+	}
+
+	return backendPool
+}
+
+// getDefaultBalancerBackendPools returns a list of default Load Balancer Backend Pools for a Load Balancer.
+func (s *VPCClusterScope) getDefaultLoadBalancerBackendPools() []vpcv1.LoadBalancerPoolPrototype {
```
dharaneeshvrd, line 1992: Any reason for creating a slice and appending one item and in caller again unpacking the slice?
Can't we return the item directly?

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1711,340 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+		listener.DefaultPool = &vpcv1.LoadBalancerPoolIdentityByName{
+			Name: additionalListener.DefaultPoolName,
+		}
+	}
+
+	return listener
+}
+
+// getDefaultLoadBalancerListeners returns a list of default Load Balancer Listeners for a Load Balancer.
+func (s *VPCClusterScope) getDefaultLoadBalancerListeners(defaultBackendPool bool) []vpcv1.LoadBalancerListenerPrototypeLoadBalancerContext {
```
dharaneeshvrd, line 2035: Same comment as `getDefaultLoadBalancerBackendPools()`'s comment.

cloud/scope/vpc_cluster.go
```diff
@@ -221,6 +222,24 @@ func (s *VPCClusterScope) CheckTagExists(tagName string) (bool, error) {
 	return exists != nil, nil
 }
 
+// GetControlPlaneSubnetIDs returns all of the Control Plane subnet Id's.
+func (s *VPCClusterScope) GetControlPlaneSubnetIDs() ([]string, error) {
+	subnets := make([]string, 0)
+	// Retrieve the subnet Id's from Status.
+	if s.NetworkStatus() != nil {
+		if s.NetworkStatus().ControlPlaneSubnets != nil {
```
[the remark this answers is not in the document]
cjschaef: Will consolidate.

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1711,340 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+
+// getLoadBalancerSubnetIDs builds the set of subnet ID's for a load balancer, or defaults to the Control Plane subnet ID's if no subnets were provided. This will attempt to transform subnet names into their respective ID's.
+func (s *VPCClusterScope) getLoadBalancerSubnetIDs(loadBalancer infrav1beta2.VPCLoadBalancerSpec) ([]string, error) {
+	subnetIDs := make([]string, 0)
+	// If Subnets were provided for the load balancer, find ID's, if necessary, and use them.
+	// Otherwise, default to trying to use the Control Plane subnets.
+	if loadBalancer.Subnets != nil {
+		for _, subnet := range loadBalancer.Subnets {
+			if subnet.ID != nil {
+				subnetIDs = append(subnetIDs, *subnet.ID)
```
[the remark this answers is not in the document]
cjschaef: Sure, I could perform lookups to check if they exist.

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1711,340 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+}
+
+// getLoadBalancerSecurityGroupIDs will collect the ID's of the desired Security Groups for a Load Balancer.
+func (s *VPCClusterScope) getLoadBalancerSecurityGroupIDs(loadBalancer infrav1beta2.VPCLoadBalancerSpec) ([]string, error) {
+	securityGroupIDs := make([]string, 0)
+	// If SecurityGroups were provided for the load balancer, find ID's, if necessary, and use them.
+	if loadBalancer.SecurityGroups != nil {
+		for _, securityGroup := range loadBalancer.SecurityGroups {
+			if securityGroup.ID != nil {
+				securityGroupIDs = append(securityGroupIDs, *securityGroup.ID)
```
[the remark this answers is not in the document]
cjschaef: Same, I could check if they exist.

cloud/scope/vpc_cluster.go
cjschaef: I did this if any additional pools were necessary to add to the defaults.

cloud/scope/vpc_cluster.go
cjschaef: Same as the pools, regarding additional default listeners.

cloud/scope/vpc_cluster.go
```diff
@@ -221,6 +222,89 @@ func (s *VPCClusterScope) CheckTagExists(tagName string) (bool, error) {
...
+
+	// NOTE(cjschaef): If Status was not set or ControlPlaneSubnets was empty, the Control Plane subnet ID's could be retrieved from Spec. However, for now consider this an error, since Subnet reconciliation should have run prior and no tracked Control Plane subnets would be a major issue.
+	return subnets, fmt.Errorf("error no control plane subnets available in status")
+}
+
+// GetLoadBalancerHostName will return the hostname of the cluster's public Load Balancer, assuming only one public Load Balancer was provided. Or, the hostname of the single private Load Balancer (assuming the cluster has no public access and only one private Load Balancer was provided).
+// NOTE(cjschaef): A webhook validation check could help ensure this.
+func (s *VPCClusterScope) GetLoadBalancerHostName() (*string, error) {
+	// If no Status or Load Balancer Status is populated, assume the Load Balancer's are not ready (have not been reconciled), so no hostname will be available.
+	if s.NetworkStatus() == nil || s.NetworkStatus().LoadBalancers == nil || len(s.NetworkStatus().LoadBalancers) == 0 {
```
Karthik-K-N, line 251: we are returning nil here for `len(s.NetworkStatus().LoadBalancers)` and same codition returns error next step, May be we can simplify or remove if not required.

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1774,366 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+func (s *VPCClusterScope) getLoadBalancer(lb infrav1beta2.VPCLoadBalancerSpec) (*infrav1beta2.VPCLoadBalancerStatus, error) {
+	var loadBalancer *vpcv1.LoadBalancer
+	var err error
+	if lb.ID != nil {
+		var detailedResponse *core.DetailedResponse
+		getLBOptions := &vpcv1.GetLoadBalancerOptions{
+			ID: lb.ID,
+		}
+		loadBalancer, detailedResponse, err = s.VPCClient.GetLoadBalancer(getLBOptions)
+		if detailedResponse != nil && detailedResponse.StatusCode == http.StatusNotFound {
```
Karthik-K-N, line 1843: I think we can log these cases with higher level V values for easy debugging in future.

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1774,366 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+	}
+
+	isPublic := true
+	// Load Balancer is private if defined that way (defaults to Public)
+	if loadBalancer.Public != nil && !*loadBalancer.Public {
+		isPublic = false
+	}
+
+	options.SetIsPublic(isPublic)
+	options.SetName(loadBalancer.Name)
```
Karthik-K-N, line 1884: What if loadbalancer.Name is empty?
In getLoadBalancer() function we have this condition
```
name := lb.Name
if name == "" {
	name = *s.GetServiceName(infrav1beta2.ResourceTypeLoadBalancer)
}
```
Should we do something similar here?

cloud/scope/vpc_cluster.go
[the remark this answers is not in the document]
Karthik-K-N: I think logging options should be sufficient.

cloud/scope/vpc_cluster.go
```diff
@@ -221,6 +222,89 @@ func (s *VPCClusterScope) CheckTagExists(tagName string) (bool, error) {
...
+		} else if loadBalancerStatus == nil {
+			// If no Load Balancer Status is found, return an error.
+			return nil, fmt.Errorf("error no load balancer status available for %s", loadBalancer.Name)
+		}
+		return loadBalancerStatus.Hostname, nil
+	}
+
+	// If only one Load Balancer was provided, attempt to use that Load Balancer's hostname.
+	if len(s.NetworkSpec().LoadBalancers) == 1 {
+		return getHostName(s.NetworkSpec().LoadBalancers[0])
```
Karthik-K-N, line 289: Any reason for handling this special case, In case of 1 lb defined it can be private as well is this the reason?

cloud/scope/vpc_cluster.go
```diff
@@ -221,6 +222,89 @@ func (s *VPCClusterScope) CheckTagExists(tagName string) (bool, error) {
...
+			if lb, ok := s.NetworkStatus().LoadBalancers[*loadBalancer.ID]; ok {
+				return lb.Hostname, nil
+			}
+			// If the Load Balancer isn't in Status, assume it is not ready/reconciled.
+			return nil, nil
+		}
+
+		// NOTE(cjschaef): As a default Load Balancer isn't supported (when none are provided in Spec), this should result in an error.
+		// TODO(cjschaef): Likely would be beneficial to have webhook validation for this case.
+		if loadBalancer.Name == "" {
```
Karthik-K-N, line 273: Not sure this check is needed here? in `getLoadBalancer()` we get the default name incase of name is empty.

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1774,366 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+			ID: lb.ID,
+		}
+		loadBalancer, detailedResponse, err = s.VPCClient.GetLoadBalancer(getLBOptions)
+		if detailedResponse != nil && detailedResponse.StatusCode == http.StatusNotFound {
+			return nil, nil
+		}
+	} else {
+		name := lb.Name
+		if name == "" {
+			name = *s.GetServiceName(infrav1beta2.ResourceTypeLoadBalancer)
```
Karthik-K-N, line 1849: In case of multiple loadbalancers configured without name in spec, This may not work as expected as we end up getting same name for all.

cjschaef: That is why we do not create an LB with the default Name (let API generate one). But that will generate its own problems.

I will need to rethink how the default service name is being used, as I don't think it makes sense to use it as requested (perhaps only in the rare case that one LB was defined in Spec). Or perhaps I can try refactoring to append `public`/`private` and hope that doesn't cause overlap (although I think that case is just as susceptible to this same situation).
Perhaps I need to add webhook validation to limit the number of LB's defined in Spec to only be 2, in additional to the refactoring. But that can be very limiting to the design and logic as well.

cloud/scope/vpc_cluster.go
cjschaef: I'll be refactoring, will see what makes sense.

cloud/scope/vpc_cluster.go
cjschaef: Yes, if a Public LB was not found, and there was only one LB, we expect it must be a Private LB, and Public traffic isn't expected to be enabled. This isn't the normal path, but still need to cover this design.

cloud/scope/vpc_cluster.go
cjschaef: I'm hesitant to create LB's with a default service name, especially if multiple LB's get defined with empty names.
Given the structure of the VPCLoadBalancerStatus, we rely on ID as a priority field. We can let VPC API generate a Name, and rely on the ID from Status.

The default service name was added by request, but I don't expect to use it for most use cases. This is likely to be refactored, likely not relying on a default name, if possible.

cloud/scope/vpc_cluster.go
cjschaef: Okay, I bumped to v5.

cloud/scope/vpc_cluster.go
cjschaef: So, the first set of checks looks at Status. Assuming if there is no Status or no LB status, the situation needs to wait for Status to exist or be updated.

The second check is against Spec. If Status was set, but Spec was not, that is an error. Checks are made immediately in the function to short circuit.

Karthik-K-N: apologies, Its correct, I just overlooked spec and status.

cloud/scope/vpc_cluster.go
Karthik-K-N: In Power VS we do something like this , append the index https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/blob/987191d98e364d4bbfc39423d433b791dbb6b102/cloud/scope/powervs_cluster.go#L1963-L1965

cloud/scope/vpc_cluster.go
```diff
@@ -1646,3 +1774,366 @@ func (s *VPCClusterScope) createSecurityGroupRuleRemote(remote infrav1beta2.VPCS
...
+	var loadBalancer *vpcv1.LoadBalancer
+	var err error
+	if lb.ID != nil {
+		var detailedResponse *core.DetailedResponse
+		getLBOptions := &vpcv1.GetLoadBalancerOptions{
+			ID: lb.ID,
+		}
+		loadBalancer, detailedResponse, err = s.VPCClient.GetLoadBalancer(getLBOptions)
+		if detailedResponse != nil && detailedResponse.StatusCode == http.StatusNotFound {
+			return nil, nil
```
dharaneeshvrd, line 1844: Generally if user provides an id, we expect the resource to be exist with the id, if not we would throw an error and not proceed with the reconciliation. It's ok in case of if name is provided and if resource not exist, we can create it. 
Also If you return nil here and create a new load balancer, then the status and spec will have different ids.

cjschaef: Okay,  I'll return an error.

Karthik-K-N:
@cjschaef Is this PR updated or tested?

Karthik-K-N (commented):
/lgtm
/assign @dharaneeshvrd

mkumatag (approved):

PR #2029, opened 2024-10-28
Fix PowerVS cluster's load balancer status check

### LINKED ISSUE #2026: Cluster incorrectly shown as ready when both LoadBalancers are not ready
/kind bug
/area provider/ibmcloud

Deploy a 4.18 cluster on a PowerVS zone where LoadBalancers are slow to create.
We are called with `InfraReady`.  We then create DNS records for the LBs.  However, only the public LB exists.  So the cluster fails to deploy.

We should wait for all specified LBs to become ready.

**Environment:**

- Cluster-api version:
- Minikube/KIND version:

dharaneeshvrd:
Need to depend on load balancer's state to set the cluster's ready status instead of using requeue logic which is designed to block the reconciliation.

Fixes #2026 

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/cloud/scope/powervs_cluster.go
+++ b/cloud/scope/powervs_cluster.go
@@ -1955,6 +1955,8 @@ func (s *PowerVSClusterScope) ReconcileLoadBalancers() (bool, error) {
 		loadBalancers = append(loadBalancers, s.IBMPowerVSCluster.Spec.LoadBalancers...)
 	}
 
+	isAnyLoadBalancerNotReady := false
+
 	for index, loadBalancer := range loadBalancers {
 		var loadBalancerID *string
 		if loadBalancer.ID != nil {
@@ -1974,8 +1976,9 @@ func (s *PowerVSClusterScope) ReconcileLoadBalancers() (bool, error) {
 				return false, err
 			}
 
-			if requeue := s.checkLoadBalancerStatus(*loadBalancer); requeue {
-				return requeue, nil
+			if isReady := s.checkLoadBalancerStatus(*loadBalancer); !isReady {
+				s.V(3).Info("LoadBalancer is still not Active", "name", *loadBalancer.Name, "state", *loadBalancer.ProvisioningStatus)
+				isAnyLoadBalancerNotReady = true
 			}
 
 			loadBalancerStatus := infrav1beta2.VPCLoadBalancerStatus{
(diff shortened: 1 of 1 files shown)


cloud/scope/powervs_cluster.go
```diff
@@ -1974,8 +1974,9 @@ func (s *PowerVSClusterScope) ReconcileLoadBalancers() (bool, error) {
 				return false, err
 			}
 
-			if requeue := s.checkLoadBalancerStatus(*loadBalancer); requeue {
-				return requeue, nil
+			if isReady := s.checkLoadBalancerStatus(*loadBalancer); !isReady {
+				s.V(3).Info("LoadBalancer is still not Active", "name", *loadBalancer.Name, "state", *loadBalancer.ProvisioningStatus)
```
Karthik-K-N, line 1978: *loadbalancer.Name can be nil when user sets only id. so better avoid it.

cloud/scope/powervs_cluster.go
```diff
@@ -1987,26 +1988,14 @@ func (s *PowerVSClusterScope) ReconcileLoadBalancers() (bool, error) {
 			continue
 		}
 
-		// check VPC load balancer exist in cloud
-		loadBalancerStatus, err := s.checkLoadBalancer(loadBalancer)
```
Karthik-K-N, line 1991: why is this removed?

cloud/scope/powervs_cluster.go
dharaneeshvrd: It's coming from cloud API resp right, would be able to create a LB without name?

Karthik-K-N: oh yeah, sorry, I overlooked loadbalancer from
```
for index, loadBalancer := range loadBalancers {
```

cloud/scope/powervs_cluster.go
dharaneeshvrd: `checkLoadBalancer` checks load balancer by the user provided name, which we already do in prev block [here](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/6fc82480e3361b2658699075690765dc66966c1f/cloud/scope/powervs_cluster.go#L1963) by recently added code via [this](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/commit/3edcccbf57561ee7b6498b7f56fcf304bb9f0c19).

Karthik-K-N: `s.GetLoadBalancerID(loadBalancer.Name)` just checks only from status not from cloud!

dharaneeshvrd: Sry, did not look properly, reverted the changes.

Karthik-K-N (commented):
/lgtm
/assign @Amulyam24 @mkumatag

Amulyam24 (commented):
/lgtm

Thank you for the quick fix @dharaneeshvrd!

Karthik-K-N:
/hold

@KeerthanaAP is doing some testing, will update here once done.

Karthik-K-N:
Thanks @KeerthanaAP for testing it quickly.
@dharaneeshvrd i think we discussed about this but forgot , we need to make sure all the LBs are active before infra is set to true.

Karthik-K-N:
> @KeerthanaAP Can you please double check this? It is working for me when I tested in local and in the log you have provided for this deployment does not have my changes.

Checking with @ hamzy  about this in slack

> @Karthik-K-N I have considered and did the changes earlier and it was working properly, but it was creating the LBs serially, now I have added a flag to create them without blocking further LBs reconcile. PTAL now.
Got it earlier will work but there was no gurantee that in case of multiple LBs all will be in Active as we used to return once we found one LB in active. But I think with this change we should be good

Karthik-K-N (commented):
xIts been [verified](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/2026#issuecomment-2442438100) and works as expected.
/lgtm

mkumatag (approved):
/unhold

PR #2039, opened 2024-11-05
Explicitly return error incase of not able to set provider id in correct format

Karthik-K-N:

This PR modifies the logic around setting provider id.
There were many instances where provider id was not set properly and the cluster creation was not successfull. Main reason for that is service instance id being set empty.

Made changes to fetch service instance id from name if id not found in spec or status. This case will occur mainly when cluster is reconciled by external controller.
Made changes to return error instead of empty string in case of not able to find service instance id.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/cloud/scope/powervs_machine.go
+++ b/cloud/scope/powervs_machine.go
@@ -929,29 +929,45 @@ func (m *PowerVSMachineScope) GetZone() string {
 }
 
 // GetServiceInstanceID returns the service instance id.
-func (m *PowerVSMachineScope) GetServiceInstanceID() string {
+func (m *PowerVSMachineScope) GetServiceInstanceID() (string, error) {
 	if m.IBMPowerVSCluster.Status.ServiceInstance != nil && m.IBMPowerVSCluster.Status.ServiceInstance.ID != nil {
-		return *m.IBMPowerVSCluster.Status.ServiceInstance.ID
+		return *m.IBMPowerVSCluster.Status.ServiceInstance.ID, nil
 	}
 	if m.IBMPowerVSCluster.Spec.ServiceInstanceID != "" {
-		return m.IBMPowerVSCluster.Spec.ServiceInstanceID
+		return m.IBMPowerVSCluster.Spec.ServiceInstanceID, nil
 	}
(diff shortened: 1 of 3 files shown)


Karthik-K-N:
/hold
Ready for review but hold for some testing.

cloud/scope/powervs_machine.go
```diff
@@ -934,29 +934,45 @@ func (m *PowerVSMachineScope) GetZone() string {
...
+	if m.IBMPowerVSCluster.Spec.ServiceInstance != nil && m.IBMPowerVSCluster.Spec.ServiceInstance.Name == nil {
+		return "", fmt.Errorf("failed to find service instance id as both name and id are not set")
+	}
+	serviceInstance, err := m.ResourceClient.GetServiceInstance("", *m.IBMPowerVSCluster.Spec.ServiceInstance.Name, ptr.To(m.GetZone()))
+	if err != nil {
+		m.Error(err, "failed to get Power VS service instance id", "serviceInstanceName", *m.IBMPowerVSCluster.Spec.ServiceInstance.Name)
+		return "", err
+	}
+	// It's safe to directly dereference GUID as its already done in NewPowerVSMachineScope
+	return *serviceInstance.GUID, nil
```
mkumatag, line 957: but we have removed this code intentionally before right?

mkumatag (changes requested):

cloud/scope/powervs_machine.go
mkumatag: @Amulyam24 ^^

Amulyam24: @mkumatag, this a different function called using PowerVSMachineScope
The function we modified earlier is https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/cloud/scope/powervs_cluster.go#L388 called using PowerVSClusterScope

Karthik-K-N:
/unhold
Verified creating a CAPI cluster and machine provider id was set as expected.

mkumatag (approved):

PR #2040, opened 2024-11-07
Refactor ReconcileNetwork

### LINKED ISSUE #1933: Validate DHCP server status before reusing existing DHCP network

dharaneeshvrd:

- If only IBMPowerVSCluster.Spec.Network is set, network would be validated and if exists already will get used as cluster’s network or a new network will be created via DHCP service.
- If only IBMPowerVSCluster.Spec.DHCPServer is set, DHCP server would be validated and if exists already, will use DHCP server’s network as cluster network. If not a new DHCP service will be created and it’s network will be used.
- If both IBMPowerVSCluster.Spec.Network & IBMPowerVSCluster.Spec.DHCPServer is set, network and DHCP server would be validated and if both exists already then network is belongs to given DHCP server or not would be validated.
- If both IBMPowerVSCluster.Spec.Network & IBMPowerVSCluster.Spec.DHCPServer is not set, by default DHCP service will be created to setup cluster's network.

Fixes #1933 

/area provider/ibmcloud

**Release note**:

```release-note
Refactor ReconcileNetwork
```

### DIFF
--- a/cloud/scope/powervs_cluster.go
+++ b/cloud/scope/powervs_cluster.go
@@ -120,6 +120,10 @@ func getTGPowerVSConnectionName(tgName string) string { return fmt.Sprintf("%s-p
 
 func getTGVPCConnectionName(tgName string) string { return fmt.Sprintf("%s-vpc-con", tgName) }
 
+func dhcpNetworkName(dhcpServerName string) string {
+	return fmt.Sprintf("DHCPSERVER%s_Private", dhcpServerName)
+}
+
 // NewPowerVSClusterScope creates a new PowerVSClusterScope from the supplied parameters.
 func NewPowerVSClusterScope(params PowerVSClusterScopeParams) (*PowerVSClusterScope, error) {
 	if params.Client == nil {
@@ -457,6 +461,14 @@ func (s *PowerVSClusterScope) SetStatus(resourceType infrav1beta2.ResourceType, 
 		}
 		s.IBMPowerVSCluster.Status.ResourceGroup.Set(resource)
 	}
+}
+
+// GetNetworkID returns the Network id from status of IBMPowerVSCluster object. If it doesn't exist, returns nil.
+func (s *PowerVSClusterScope) GetNetworkID() *string {
+	if s.IBMPowerVSCluster.Status.Network != nil {
+		return s.IBMPowerVSCluster.Status.Network.ID
+	}
+	return nil
 }
 
 // Network returns the cluster Network.
@@ -847,18 +859,33 @@ func (s *PowerVSClusterScope) createServiceInstance() (*resourcecontrollerv2.Res
 	return serviceInstance, nil
 }
 
-// ReconcileNetwork reconciles network.
+// ReconcileNetwork reconciles network
+// If only IBMPowerVSCluster.Spec.Network is set, network would be validated and if exists already will get used as cluster’s network or a new network will be created via DHCP service.
+// If only IBMPowerVSCluster.Spec.DHCPServer is set, DHCP server would be validated and if exists already, will use DHCP server’s network as cluster network. If not a new DHCP service will be created and it’s network will be used.
+// If both IBMPowerVSCluster.Spec.Network & IBMPowerVSCluster.Spec.DHCPServer is set, network and DHCP server would be validated and if both exists already then network is belongs to given DHCP server or not would be validated.
+// If both IBMPowerVSCluster.Spec.Network & IBMPowerVSCluster.Spec.DHCPServer is not set, by default DHCP service will be created to setup cluster's network.
 func (s *PowerVSClusterScope) ReconcileNetwork() (bool, error) {
-	if s.GetDHCPServerID() != nil {
+	if s.GetNetworkID() != nil {
+		// Check the network exists
+		if _, err := s.IBMPowerVSClient.GetNetworkByID(*s.GetNetworkID()); err != nil {
+			return false, err
+		}
+
+		if s.GetDHCPServerID() == nil {
+			// If only network is set, return once network is validated to be ok
+			return true, nil
+		}
+
 		s.V(3).Info("DHCP server ID is set, fetching details", "dhcpServerID", s.GetDHCPServerID())
 		active, err := s.isDHCPServerActive()
 		if err != nil {
 			return false, err
 		}
-		// if dhcp server exist and in active state, its assumed that dhcp network exist
-		// TODO(Phase 2): Verify that dhcp network is exist.
-		return active, nil
-		//	TODO(karthik-k-n): If needed set dhcp status here
+		// DHCP server still not active, skip checking network for now
+		if !active {
+			return false, nil
+		}
+		return true, nil
 	}
 	// check network exist in cloud
 	networkID, err := s.checkNetwork()
@@ -868,10 +895,20 @@ func (s *PowerVSClusterScope) ReconcileNetwork() (bool, error) {
 	if networkID != nil {
 		s.V(3).Info("Found PowerVS network in IBM Cloud", "networkID", networkID)
 		s.SetStatus(infrav1beta2.ResourceTypeNetwork, infrav1beta2.ResourceReference{ID: networkID, ControllerCreated: ptr.To(false)})
+	}
+	dhcpServerID, err := s.checkDHCPServer()
+	if err != nil {
+		return false, err
+	}
+	if dhcpServerID != nil {
+		s.V(3).Info("Found DHCP server in IBM Cloud", "dhcpServerID", dhcpServerID)
+		s.SetStatus(infrav1beta2.ResourceTypeDHCPServer, infrav1beta2.ResourceReference{ID: dhcpServerID, ControllerCreated: ptr.To(false)})
+	}
+	if s.GetNetworkID() != nil {
 		return true, nil
 	}
 
-	dhcpServerID, err := s.createDHCPServer()
+	dhcpServerID, err = s.createDHCPServer()
 	if err != nil {
 		s.Error(err, "Error creating DHCP server")
 		return false, err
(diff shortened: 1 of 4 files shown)


cloud/scope/powervs_cluster.go
```diff
@@ -882,37 +900,110 @@ func (s *PowerVSClusterScope) ReconcileNetwork() (bool, error) {
...
+
+// isNetworkAvailable verifies the user provided network ID is available and it is managed by DHCP.
+func (s *PowerVSClusterScope) isNetworkAvailable() (bool, error) {
+	nw, err := s.IBMPowerVSClient.GetNetworkByID(*s.GetNetworkID())
+	if err != nil {
+		return false, err
 	}
-	return network.NetworkID, nil
+
+	if nw != nil && !nw.DhcpManaged {
```
Karthik-K-N, line 1002: Do you think nw.DhcpManaged is required? We are getting network id from status and we set network id to status only when we create DHCP server.

cloud/scope/powervs_cluster.go
```diff
@@ -882,37 +900,110 @@ func (s *PowerVSClusterScope) ReconcileNetwork() (bool, error) {
...
+	}
+
 	if s.IBMPowerVSCluster.Spec.Network.ID != nil {
-		s.V(3).Info("Checking if PowerVS network exists in IBM Cloud with ID", "networkID", *s.IBMPowerVSCluster.Spec.Network.ID)
-		network, err := s.IBMPowerVSClient.GetNetworkByID(*s.IBMPowerVSCluster.Spec.Network.ID)
+		networkID, err := checkNetworkByID(*s.IBMPowerVSCluster.Spec.Network.ID)
 		if err != nil {
 			return nil, err
 		}
+		if err := s.verifyDHCPServerRespectToNetwork(*networkID); err != nil {
```
Karthik-K-N, line 920: I think this forces the user to set the dhcp network only. so we should update the api spec to mention it.
Also never tried what happens if we pass public network id here, Not relaying on DHCP.

cloud/scope/powervs_cluster.go
dharaneeshvrd: Ok I can remove this check.

cloud/scope/powervs_cluster.go
dharaneeshvrd: Yes. I think we should use DHCP server only, during boot, CoreOS expects the ip from a DHCP server. I hope the case is same for vanilla capi cluster as well.
Shall I modify the API spec as part of this PR?

cloud/scope/powervs_cluster.go
```diff
@@ -882,37 +901,98 @@ func (s *PowerVSClusterScope) ReconcileNetwork() (bool, error) {
...
+	}
+
 	// if the user has provided the already existing dhcp server name then there might exist network name
 	// with format DHCPSERVER<DHCPServer.Name>_Private , try fetching that
 	var networkName string
 	if s.DHCPServer() != nil && s.DHCPServer().Name != nil {
-		networkName = fmt.Sprintf("DHCPSERVER%s_Private", *s.DHCPServer().Name)
+		networkName = dhcpNetworkName(*s.DHCPServer().Name)
 	} else {
 		networkName = *s.GetServiceName(infrav1beta2.ResourceTypeNetwork)
```
Amulyam24, line 971: is this required? as we are already checking this condition above

cloud/scope/powervs_cluster.go
```diff
@@ -882,37 +901,98 @@ func (s *PowerVSClusterScope) ReconcileNetwork() (bool, error) {
 	return false, nil
 }
 
-// checkNetwork checks if network exists in cloud with given name or ID mentioned in spec.
-func (s *PowerVSClusterScope) checkNetwork() (*string, error) {
-	if s.IBMPowerVSCluster.Spec.Network.ID != nil {
-		s.V(3).Info("Checking if PowerVS network exists in IBM Cloud with ID", "networkID", *s.IBMPowerVSCluster.Spec.Network.ID)
-		network, err := s.IBMPowerVSClient.GetNetworkByID(*s.IBMPowerVSCluster.Spec.Network.ID)
+// checkNetwork checks if network exists in cloud with given network's ID or name or dhcp server's ID or name mentioned in spec.
+// Priority order: Network ID - Network name - DHCP server ID - DHCP server name - Network with cluster name.
```
Amulyam24, line 905:
```suggestion
// Priority order: Network ID > Network name > DHCP server ID > DHCP server name > Network with cluster name.
```

cloud/scope/powervs_cluster.go
dharaneeshvrd: Can you please give me the ref? I think this is the only place where I am checking with cluster name.

Amulyam24: Thanks for clarifying, I understand it is required to preserve the order.

dharaneeshvrd: Sure, addressed your other comments. ptal.

cloud/scope/powervs_cluster.go
Karthik-K-N: We use Centos images in normal capi flow as well. 
@mkumatag whats your thought. Should we make this restriction or do you foresee any other use cases.

mkumatag: nope, we should allow both dhcp based and regular public networks.

dharaneeshvrd: Ok, current code designed to handle only DHCP network. @Karthik-K-N We can create a story for this to handle later I believe. Wdyt?

Karthik-K-N: Yeah, That can be done.

dharaneeshvrd:
@Karthik-K-N @Amulyam24 I have refactored this PR to handle plain network and DHCP network separately. PTAL.

cloud/scope/powervs_cluster.go
```diff
@@ -882,37 +915,80 @@ func (s *PowerVSClusterScope) ReconcileNetwork() (bool, error) {
...
 		if err != nil {
 			return nil, err
 		}
-		return network.NetworkID, nil
+		if s.GetNetworkID() == nil {
+			if _, err := s.IBMPowerVSClient.GetNetworkByID(*dhcpServer.Network.ID); err != nil {
+				return nil, err
+			}
+			s.SetStatus(infrav1beta2.ResourceTypeNetwork, infrav1beta2.ResourceReference{ID: dhcpServer.Network.ID, ControllerCreated: ptr.To(false)})
+		} else if *dhcpServer.Network.ID != *s.GetNetworkID() {
```
Amulyam24, line 932: can there occur a scenario where `dhcpServer.Network` is nil?

Amulyam24 (commented):
Overall flow looks neat and good, couple of questions

1. Should we update this var name [dhcpServerActive](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/30ee9688b71c520ba8e3a8a35f75a501431bc91c/controllers/ibmpowervscluster_controller.go#L163C5-L163C21) since we check for both network and dhcp?
2. In a case where only network name or ID is set in spec, we do not check for associated dhcp servers right, is this to accommodate public network type?

cloud/scope/powervs_cluster.go
dharaneeshvrd: I don't think so, but have added a nil check for `dhcpServer.Network` anyway, ptal.

dharaneeshvrd:
> Should we update this var name [dhcpServerActive](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/30ee9688b71c520ba8e3a8a35f75a501431bc91c/controllers/ibmpowervscluster_controller.go#L163C5-L163C21) since we check for both network and dhcp?
Agree, We mainly check network's active status, so kept `networkActive`

Amulyam24 (commented):
overall LGTM.

one minor suggestion, in UT, can we also verify the `ControllerCreated` value wherever applicable?

/cc @Karthik-K-N

cloud/scope/powervs_cluster.go
```diff
@@ -868,10 +891,20 @@ func (s *PowerVSClusterScope) ReconcileNetwork() (bool, error) {
 	if networkID != nil {
 		s.V(3).Info("Found PowerVS network in IBM Cloud", "networkID", networkID)
 		s.SetStatus(infrav1beta2.ResourceTypeNetwork, infrav1beta2.ResourceReference{ID: networkID, ControllerCreated: ptr.To(false)})
+	}
+	dhcpServerID, err := s.checkDHCPServer()
+	if err != nil {
+		return false, err
+	}
+	if dhcpServerID != nil {
+		s.V(3).Info("Found DHCP server in IBM Cloud", "DHCP Server ID", dhcpServerID)
```
Karthik-K-N, line 900: To keep uniform and improve our [logging](https://cluster-api.sigs.k8s.io/developer/logging#keyvalue-pairs), lets use , `dhcpServerID` instead of `"DHCP Server ID"`

cloud/scope/powervs_cluster.go
```diff
@@ -882,37 +915,80 @@ func (s *PowerVSClusterScope) ReconcileNetwork() (bool, error) {
...
+// If exists and s.IBMPowerVSCluster.Status.Network is populated already will validate the DHCP server's network and cluster networks are matching, if not will throw an error.
+func (s *PowerVSClusterScope) checkDHCPServer() (*string, error) {
+	if s.DHCPServer() != nil && s.DHCPServer().ID != nil {
+		dhcpServer, err := s.IBMPowerVSClient.GetDHCPServer(*s.DHCPServer().ID)
 		if err != nil {
 			return nil, err
 		}
-		return network.NetworkID, nil
+		if s.GetNetworkID() == nil {
+			if _, err := s.IBMPowerVSClient.GetNetworkByID(*dhcpServer.Network.ID); err != nil {
```
Karthik-K-N, line 928: Is it safe to say that dhcpServer will be active by this time? If not the network may be nil.
I am just concerned about `*dhcpServer.Network.ID`

cloud/scope/powervs_cluster.go
```diff
@@ -882,37 +915,80 @@ func (s *PowerVSClusterScope) ReconcileNetwork() (bool, error) {
...
 	var networkName string
 	if s.DHCPServer() != nil && s.DHCPServer().Name != nil {
-		networkName = fmt.Sprintf("DHCPSERVER%s_Private", *s.DHCPServer().Name)
+		networkName = dhcpNetworkName(*s.DHCPServer().Name)
 	} else {
-		networkName = *s.GetServiceName(infrav1beta2.ResourceTypeNetwork)
+		networkName = dhcpNetworkName(s.InfraCluster())
 	}
 
 	s.V(3).Info("Checking if PowerVS network exists in IBM Cloud with name", "name", networkName)
```
Karthik-K-N, line 946: This log line should be updated to state that, we are trying to look for network from DHCPServer's network list

cloud/scope/powervs_cluster.go
```diff
@@ -882,37 +915,80 @@ func (s *PowerVSClusterScope) ReconcileNetwork() (bool, error) {
...
+		s.V(3).Info("Checking if PowerVS network exists in IBM Cloud with ID", "networkID", *s.Network().ID)
+		network, err := s.IBMPowerVSClient.GetNetworkByID(*s.Network().ID)
+		if err != nil {
+			return nil, err
+		}
+		return network.NetworkID, nil
+	}
+
+	if s.Network().Name != nil {
+		s.V(3).Info("Checking if PowerVS network exists in IBM Cloud with name")
```
Karthik-K-N, line 980: Lets log network name also.

Karthik-K-N (commented):
ReconcileNetwork is very compilcated logic, Thank you for taking this up.

Few comments otherwise LGTM. Can we also mention in high level as comments over ReconcileNetwork() function what we are doing.

cloud/scope/powervs_cluster.go
dharaneeshvrd: Here we are validating existing DHCP server, new DHCP server creation won't reach this block, but anyway I have handled this via nil check, PTAL.

dharaneeshvrd:
@Karthik-K-N @Amulyam24 I have tried to address your comments, PTAL.

cloud/scope/powervs_cluster.go
```diff
@@ -847,18 +859,32 @@ func (s *PowerVSClusterScope) createServiceInstance() (*resourcecontrollerv2.Res
 	return serviceInstance, nil
 }
 
-// ReconcileNetwork reconciles network.
+// ReconcileNetwork reconciles network
+// If only IBMPowerVSCluster.Spec.Network is set, network would be validated and if exists already will get used as cluster’s network or a new network will be created via DHCP service.
```
Karthik-K-N, line 863: Can you also mention what happens when both network and dhcpserver is not set, Currently that configuration is used the most.

Karthik-K-N (commented):
LGTM lets fix the failing test. Thank you.

Karthik-K-N (commented):
/lgtm

Amulyam24 (approved):

Prajyot-Parab (approved):
/lgtm
Thanks @dharaneeshvrd

PR #2051, opened 2024-11-13
VPC: Enhance the Machine LB Pool Member delete logic

cjschaef:
Enhance the deletion logic for extended Machine LB Pool Member support.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Enhance the deletion logic for extended Machine LB Pool Member support.
```


cjschaef:
This is requires #2042 before I rebase and open.

cloud/scope/machine.go
```diff
@@ -767,6 +772,75 @@ func (m *MachineScope) DeleteVPCLoadBalancerPoolMember() error {
...
+		}
+
+		m.Logger.V(5).Info("list load balancer pool members options", "machineName", m.IBMVPCMachine.Name, "options", *listMembersOptions)
+		poolMembers, _, err := m.IBMVPCClient.ListLoadBalancerPoolMembers(listMembersOptions)
+		if err != nil {
+			return fmt.Errorf("error retrieving load balancer pool members for load balancer pool member deletion for machine %s: %w", m.IBMVPCMachine.Name, err)
+		}
+
+		for _, poolMember := range poolMembers.Members {
+			if poolMemberTarget, ok := poolMember.Target.(*vpcv1.LoadBalancerPoolMemberTarget); ok {
```
Karthik-K-N, line 819: May be if we negate the condition it looks cleaner.
```
if poolMemberTarget, ok := poolMember.Target.(*vpcv1.LoadBalancerPoolMemberTarget); !ok {
    continue
}
```

Karthik-K-N (commented):
one small change, otherwise LGTM

another question
We can only perform action on the LB Pool member when the LB in ACTIVE state right, if its not in ACTIVE state wont it through error? 
If I remember correctly when I was trying to add another pool member I used to hit this error.

cjschaef:
Rebasing now that #2042 has merged.

cloud/scope/machine.go
cjschaef: I can add a continue, but I cannot make the change you are asking for, as `poolMemberTarget` will only be scoped to the if block, which will now only contain `continue`.

cjschaef: Sure, I'll change the API call for the LB to adjust how the error gets reported for the LB status.

cloud/scope/machine.go
```diff
@@ -766,6 +771,99 @@ func (m *MachineScope) DeleteVPCLoadBalancerPoolMember() error {
...
+		// Retrieve the Load Balancer.
+		var loadBalancerDetails *vpcv1.LoadBalancer
+		if member.LoadBalancer.ID != nil {
+			loadBalancerDetails, _, err = m.IBMVPCClient.GetLoadBalancer(&vpcv1.GetLoadBalancerOptions{
+				ID: member.LoadBalancer.ID,
+			})
+		} else if member.LoadBalancer.Name != nil {
+			loadBalancerDetails, err = m.IBMVPCClient.GetLoadBalancerByName(*member.LoadBalancer.Name)
+		} else {
+			return fmt.Errorf("error load balancer has no id or name for load balancer pool member deleteion")
```
Prajyot-Parab, line 802: wondering if this scenario is valid, will the flow reach here anytime?

cloud/scope/machine.go
```diff
@@ -766,6 +771,99 @@ func (m *MachineScope) DeleteVPCLoadBalancerPoolMember() error {
...
+		if err != nil {
+			return fmt.Errorf("error retrieving load balancer pool members for load balancer pool member deletion for machine %s: %w", m.IBMVPCMachine.Name, err)
+		}
+
+		for _, poolMember := range poolMembers.Members {
+			poolMemberTarget, ok := poolMember.Target.(*vpcv1.LoadBalancerPoolMemberTarget)
+			if !ok {
+				continue
+			}
+			if poolMemberTarget.Address != nil && *poolMemberTarget.Address == *instanceDetails.PrimaryNetworkInterface.PrimaryIP.Address {
```
Karthik-K-N, line 832:
```suggestion
			if poolMemberTarget.Address != nil && *poolMemberTarget.Address != *instanceDetails.PrimaryNetworkInterface.PrimaryIP.Address {
			continue
			}
```

cloud/scope/machine.go
```diff
@@ -766,6 +771,99 @@ func (m *MachineScope) DeleteVPCLoadBalancerPoolMember() error {
...
+				m.Logger.V(3).Info("found load balancer pool member to delete", "machineName", m.IBMVPCMachine.Name, "poolMemberID", *poolMember.ID)
+				// Make LB status check now that it has been determined a change is required.
+				if *loadBalancerDetails.ProvisioningStatus != string(infrav1beta2.VPCLoadBalancerStateActive) {
+					m.Logger.V(5).Info("load balancer not in active status prior to load balancer pool member deletion", "machineName", m.IBMVPCMachine.Name, "loadBalancerID", *loadBalancerDetails.ID, "loadBalancerProvisioningStatus", *loadBalancerDetails.ProvisioningStatus)
+					// Set flag that some cleanup was not completed, and break out of member target loop, to try next member from Machine Status.
+					cleanupIncomplete = true
```
Karthik-K-N, line 838: I think we can directly return with error from here right? Anyhow we are not doing anything outside for loop apart from checking `cleanupIncomplete`

Prajyot-Parab (changes requested):

cloud/scope/machine.go
Prajyot-Parab: also name is not being set here - https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/cloud/scope/machine.go#L642-L650
do you plan to handle this at a later stage.

cjschaef: Ideally not, but seems like a good idea to catch this edge case than continue

cloud/scope/machine.go
cjschaef: It is more ideal to attempt to reconcile additional LB members if possible, not block on a single LB at a time.

Correct, once going through all supplied LB Pool Member "requests", if any failed, return the failure. If none failed or were necessary, return nil.

cloud/scope/machine.go
cjschaef: Ah, will fix

cjschaef: I've refactored and combined with previous if block to `continue`.

cloud/scope/machine.go
Karthik-K-N: In that case of iterating over other pool members, we should use continue instead of break right?

cjschaef: No, if we found there is a Pool Member that matches the Machine's IP, that should be removed, but the LB is not active (cannot be updated), we don't need to check any additional Pool Members. So, break from the inner for loop, and continue with the outer for loop, checking the defined Pool Members in Machine Status (if it has more than one Pool Member defined for the Machine).
(Continue would simply continue iterating through the inner for loop, which shouldn't find any additional IP Address matches, since we already found the one matching the Machine IP Address). Shortcut those remaining non-matching members in the inner loop by `break`.

If we do not find a matching Member inside the inner for loop, we don't care what state the LB is in (active or not), so it should simply iterate through the inner loop, finding no matches, and naturally continue the outer for loop, if necessary.

cloud/scope/machine.go
cjschaef: The hope is the ID gets set during Machine creation, but if Name is specified instead, outside the standard deployment process, it can still be used in place of ID.

cloud/scope/machine.go
Karthik-K-N: Got the point thanks for the explaination, Another question is we are fetching the lb details in the outerloop can't we check teh lb status there instead of checking it inside the second loop?

cloud/scope/machine.go
```diff
@@ -766,6 +771,99 @@ func (m *MachineScope) DeleteVPCLoadBalancerPoolMember() error {
...
+		}
+		instanceDetails, _, err := m.IBMVPCClient.GetInstance(instanceOptions)
+		if err != nil {
+			return fmt.Errorf("error retrieving instance for load balancer pool member deletion for machine %s: %w", m.IBMVPCMachine.Name, err)
+		}
+		// Verify the instance has a primary network interface IP address needed to delete the correct Load Balancer Pool Member.
+		if instanceDetails.PrimaryNetworkInterface == nil || instanceDetails.PrimaryNetworkInterface.PrimaryIP == nil || instanceDetails.PrimaryNetworkInterface.PrimaryIP.Address == nil {
+			return fmt.Errorf("error instance is missing the primary network interface IP address for load balancer pool member deletion for machine: %s", m.IBMVPCMachine.Name)
+		}
+		m.Logger.V(5).Info("collected instance details for load balancer pool member deletion", "machienName", m.IBMVPCMachine.Name, "instanceID", *instanceDetails.ID, "instanceIP", *instanceDetails.PrimaryNetworkInterface.PrimaryIP.Address)
```
Karthik-K-N, line 791: Can thins also moved out of for loop? We are fetching machine details and it remains same?

cjschaef: Yep, will move.

cloud/scope/machine.go
cjschaef: I could, but if we find out the LB Pool Member is already gone (it does not show up in the inner loop), we don't need to worry about the LB status.
Thus, we don't need to wait for every LB to be active, just to check if it has a matching Pool member.
So, if another machine is requesting removal of a member in the same LB, we aren't blocked waiting, just to find out that LB no longer have the member anyway.

cloud/scope/machine.go
cjschaef: done

Karthik-K-N (commented):
Thank you
/lgtm

Prajyot-Parab (approved):
/lgtm

PR #2059, opened 2024-11-21
Support bootstrap only additional listener

hamzy:
Add an option to the AdditionalListenerSpec to signify that this pool should only contain the boostrap node.  An example of this would be for port 22 since every master node also has ssh but we need the public load balancer to only connect to the bootstrap ssh server and not randomly to any one of the four nodes.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/api/v1beta2/ibmvpccluster_types.go
+++ b/api/v1beta2/ibmvpccluster_types.go
@@ -127,6 +127,11 @@ type AdditionalListenerSpec struct {
 	// Will default to TCP protocol if not specified.
 	// +optional
 	Protocol *VPCLoadBalancerListenerProtocol `json:"protocol,omitempty"`
+
+	// Add defines which VMs are added to this pool.  If empty, it will default to all VMs.
+	// +kubebuilder:validation:Enum=all;bootstrap_only;masters_only
+	// +optional
+	Add VPCLoadBalancerPoolAdd `json:"add,omitempty"`
 }
 
 // VPCLoadBalancerBackendPoolSpec defines the desired configuration of a VPC Load Balancer Backend Pool.
--- a/cloud/scope/powervs_machine.go
+++ b/cloud/scope/powervs_machine.go
@@ -1060,6 +1060,35 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember() (*vpcv1.LoadBala
 					}
 				}
 			}
+
+			for _, al := range lb.AdditionalListeners {
+				var shouldAdd bool
+
+				machineType, ok := m.IBMPowerVSMachine.ObjectMeta.Labels["powervs.cluster.x-k8s.io/node-type"]
+				m.V(3).Info("Testing additional listener", "Add", al.Add, "Port", al.Port, "machineType", machineType, "ok", ok)
+
+				if al.Add == "" || !ok {
+					continue
+				}
+
+				switch al.Add {
+				case infrav1beta2.VPCLoadBalancerPoolAddAll:
+					shouldAdd = true
+				case infrav1beta2.VPCLoadBalancerPoolAddBootstrapOnly:
+					shouldAdd = machineType == "bootstrap"
+				case infrav1beta2.VPCLoadBalancerPoolAddMastersOnly:
+					shouldAdd = machineType == "master"
+				default:
+					return nil, fmt.Errorf("unknown VPCLoadBalancerPoolAdd type: %s", al.Add)
+				}
+				m.V(3).Info("Testing additional listener", "targetPort", targetPort, "shouldAdd", shouldAdd)
+				if al.Port == targetPort && !shouldAdd {
+					m.V(3).Info("Skipping PoolMember because of Add value of", "Add", al.Add, "name", m.IBMPowerVSMachine.Name)
+					alreadyRegistered = true
+					continue
+				}
+			}
+
 			if alreadyRegistered {
 				m.V(3).Info("PoolMember already exist", "pool", *pool.Name, "targetip", internalIP, "port", targetPort)
 				continue
@@ -1089,7 +1118,7 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember() (*vpcv1.LoadBala
 			if err != nil {
 				return nil, fmt.Errorf("failed to create VPC load balancer %s pool member %v", *loadBalancer.Name, err)
 			}
-			m.Info("Created VPC load balancer pool member", "id", *loadBalancerPoolMember.ID)
+			m.Info("Created VPC load balancer pool member", "id", *loadBalancerPoolMember.ID, "pool", *pool.Name)
 			return loadBalancerPoolMember, nil
 		}
 	}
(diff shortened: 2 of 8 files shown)


Karthik-K-N:
Overall I like the idea just thinking, Can we generalise this rather than sticking only to bootstrap configuration?
Say tomorrow we want to configure something only on the master nodes then we need to make the changes again.
I would like to know the opinions
@dharaneeshvrd @Amulyam24 @Prajyot-Parab

dharaneeshvrd:
> Overall I like the idea just thinking, Can we generalise this rather than sticking only to bootstrap configuration?
Say tomorrow we want to configure something only on the master nodes then we need to make the changes again.

+1

hamzy:
@Karthik-K-N @dharaneeshvrd 
What do you think now?

cloud/scope/powervs_machine.go
```diff
@@ -1060,6 +1060,25 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember() (*vpcv1.LoadBala
...
+			for _, al := range lb.AdditionalListeners {
+				var shouldAdd bool
+
+				switch al.Enabled {
+				case infrav1beta2.AdditionalListenerAll:
+					shouldAdd = true
+				case infrav1beta2.AdditionalListenerBootstrapOnly:
+					shouldAdd = strings.Contains(m.IBMPowerVSMachine.Name, "-bootstrap")
+				case infrav1beta2.AdditionalListenerMastersOnly:
+					shouldAdd = strings.Contains(m.IBMPowerVSMachine.Name, "-master")
```
Karthik-K-N, line 1073: I am not sure need to check but I think rather than name we can use labels/anotations which identify that a machine is control-plan or data?

Karthik-K-N (commented):
This approach looks good to me in general but would like to know @dharaneesh openion as well.

Karthik-K-N:
Tagging @cjschaef as he might thought about it or might have some other approach.

hamzy:
> Tagging @cjschaef as he might thought about it or might have some other approach.

Are we bike shedding?

hamzy:
Can we progress on this?  It's been two weeks and this is an important issue.

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -108,6 +108,20 @@ type VPCLoadBalancerSpec struct {
 	Subnets []VPCResource `json:"subnets,omitempty"`
 }
 
+// AdditionalListenerEnabled defines what category of VMs are added to the listener pool.
+type AdditionalListenerEnabled int
+
+const (
+	// AdditionalListenerAll add every VM to the listener pool.
```
dharaneeshvrd, line 115: Using 0, 1 & 2 to identify the nodes does not look good to me. Can we define the possible values using enum like [this](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/api/v1beta2/types.go#L138) and add the necessary kubebuilder validation?

cloud/scope/powervs_machine.go
dharaneeshvrd: I too feel using static strings to identify the nodes is not a good idea. If there is no other way we can go with this approach.

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -108,6 +108,20 @@ type VPCLoadBalancerSpec struct {
...
+	AdditionalListenerAll AdditionalListenerEnabled = iota
+
+	// AdditionalListenerBootstrapOnly add only the bootstrap VM to the listener pool.
+	AdditionalListenerBootstrapOnly
```
mkumatag, line 119: there is nothing like bootstrap term in the capi project, we need to find some generic way to deal with this problem.

mkumatag (changes requested):

mkumatag (changes requested):
doesn't matter, can we have something like - add the machines based on the condition set in the `AdditionalListenerSpec` spec e.g: some label/annotation so that we can make this generic.

e.g:

```
selectors:
      - "powervs.cluster.x-k8s.io/node-type": "bootstrap"
```

and bootstrap machine can have this annotation to get added into that additional listener.

default: we can keep the current behaviour as is to add all the machines if no selector specified.
>The bootstrap VM is not technically a node.

hamzy:
Do you have a code example for querying a selector in Go?

mkumatag:
> Do you have a code example for querying a selector in Go?

I don't have code handy

cloud/scope/powervs_machine.go
```diff
@@ -1060,6 +1060,35 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember() (*vpcv1.LoadBala
...
+				}
+
+				switch al.Add {
+				case infrav1beta2.VPCLoadBalancerPoolAddAll:
+					shouldAdd = true
+				case infrav1beta2.VPCLoadBalancerPoolAddBootstrapOnly:
+					shouldAdd = machineType == "bootstrap"
+				case infrav1beta2.VPCLoadBalancerPoolAddMastersOnly:
+					shouldAdd = machineType == "master"
+				default:
```
dharaneeshvrd, line 1081: Don't we need to set `shouldAdd = true` here also, when it's not set? Also can you please add UT covering this new logic?

cloud/scope/powervs_machine.go
```diff
@@ -1060,6 +1060,35 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember() (*vpcv1.LoadBala
 					}
 				}
 			}
+
+			for _, al := range lb.AdditionalListeners {
+				var shouldAdd bool
+
+				machineType, ok := m.IBMPowerVSMachine.ObjectMeta.Labels["powervs.cluster.x-k8s.io/node-type"]
```
Karthik-K-N, line 1067: Should we use  label or annotation for selecting the machines to configure, Usually CAPI uses annotation to select the machines.
https://github.com/Karthik-K-N/cluster-api/blob/f3d39741b47ae458625a28e6b37e3adeb18dfeb1/api/v1beta1/common_types.go#L98-L103

cloud/scope/powervs_machine.go
Karthik-K-N: Yes, The default case workflow should be like the existing behavior, It should not break the existing flow.

api/v1beta2/ibmvpccluster_types.go
```diff
@@ -127,6 +127,11 @@ type AdditionalListenerSpec struct {
 	// Will default to TCP protocol if not specified.
 	// +optional
 	Protocol *VPCLoadBalancerListenerProtocol `json:"protocol,omitempty"`
+
+	// Add defines which VMs are added to this pool.  If empty, it will default to all VMs.
+	// +kubebuilder:validation:Enum=all;bootstrap_only;masters_only
+	// +optional
+	Add VPCLoadBalancerPoolAdd `json:"add,omitempty"`
```
Karthik-K-N, line 134: Can we rename it with better name, Something like MachineSelector or something?

cloud/scope/powervs_machine.go
hamzy: The default case throws an error so it does not need shouldAdd to be set.
What is UT?  Do you have an example of how to implement it?

mkumatag: first lets focus on finalising the API spec, see my comments in the api/

PR #2095, opened 2024-12-09
Bumped go.mod version in tools

anshuman-agarwala:
**What this PR does / why we need it**: Fixes the lint failure here: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2071
This is happening because dependabot does not update the go.mod version automatically and our CI checks that running `go mod tidy` does not make any further changes to the go.mod file. The version change is happening because go-lint depends on version 1.22.1 which is due to this dependency in go-lint: honnef.co/go/tools

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/hack/tools/go.mod
+++ b/hack/tools/go.mod
@@ -2,8 +2,6 @@ module sigs.k8s.io/cluster-api-provider-ibmcloud/hack/tools
 
 go 1.22.9
 
-toolchain go1.22.9
-
 replace sigs.k8s.io/cluster-api => sigs.k8s.io/cluster-api v1.9.0-rc.0
 
 require (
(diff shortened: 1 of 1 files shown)


hack/tools/go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/cluster-api-provider-ibmcloud/hack/tools
 
-go 1.22.0
+go 1.22.10
```
mkumatag, line 3: right now we are using `1.22.9` version, we need to limit to that or bump the version for the project before this change

cc @Amulyam24 @Prajyot-Parab

mkumatag (changes requested):

hack/tools/go.mod
anshuman-agarwala: Changed version from `1.22.10` to `1.22.9` to be in line with our go version.

mkumatag (approved):
/lgtm

PR #2145, opened 2025-01-27
Network regex validation for powervs cluster

### LINKED ISSUE #1994: Support fetching Power VS network with regular expression
/kind feature
/area provider/ibmcloud

A Power VS network can be identified by [id, name or regular expression](https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/blob/632a722671b28ad67730b94e244c9e479ead999d/api/v1beta2/ibmpowervscluster_types.go#L47-L46), Currently we support both id and name. I would like to have support for fetching network by regular expression.

carmal891:

Fixes #1994 

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/api/v1beta2/common.go
+++ b/api/v1beta2/common.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1beta2
 
 import (
+	"regexp"
 	"strconv"
 
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -42,14 +43,22 @@ func validateIBMPowerVSResourceReference(res IBMPowerVSResourceReference, resTyp
 	if res.ID != nil && res.Name != nil {
 		return false, field.Invalid(field.NewPath("spec", resType), res, "Only one of "+resType+" - ID or Name may be specified")
 	}
+
 	return true, nil
 }
 
 func validateIBMPowerVSNetworkReference(res IBMPowerVSResourceReference) (bool, *field.Error) {
+	// Ensure only one of ID, Name, or RegEx is specified
 	if (res.ID != nil && res.Name != nil) || (res.ID != nil && res.RegEx != nil) || (res.Name != nil && res.RegEx != nil) {
 		return false, field.Invalid(field.NewPath("spec", "Network"), res, "Only one of Network - ID, Name or RegEx can be specified")
 	}
 	return true, nil
+}
+
+// regexMatches validates if a given regex matches the target string.
+func regexMatches(pattern, target string) bool {
+	matched, err := regexp.MatchString(pattern, target)
+	return err == nil && matched
 }
 
 func validateIBMPowerVSMemoryValues(resValue int32) bool {
--- a/api/v1beta2/ibmpowervscluster_webhook.go
+++ b/api/v1beta2/ibmpowervscluster_webhook.go
@@ -95,8 +95,46 @@ func (r *IBMPowerVSCluster) validateIBMPowerVSCluster() (admission.Warnings, err
 		r.Name, allErrs)
 }
 
+func (r *IBMPowerVSCluster) validateNetworkRegex() (bool, *field.Error) {
+	if r.Spec.Network.RegEx == nil {
+		return true, nil
+	}
+	var targetName string
+	var validationMessage string
+
+	// If only spec.Network.RegEx is set and no other network resources are specified,
+	// the controller will create a DHCP server and the corresponding network in the format:
+	// DHCPSERVER<dhcp_server_name>_Private.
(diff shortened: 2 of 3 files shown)


carmal891:
@Karthik-K-N As discussed, set the network regex validation only for powervs cluster for now. We can further triage the any additional changes required

Amulyam24 (commented):
@Karthik-K-N, I'm not quite sure if this validation makes sense in all scenarios.

IMO using regex is ideal when a resource exists but how will this validation help when DHCP server name is not set and it will end up creating a new DHCP server with name in format `DHCPSERVER<dhcp_server_name>_Private` later in the flow. Isn't the main aim to check if the network regex matches the format?

cc @dharaneeshvrd

Karthik-K-N:
> @Karthik-K-N, I'm not quite sure if this validation makes sense in all scenarios.
> 
> IMO using regex is ideal when a resource exists but how will this validation help when DHCP server name is not set and it will end up creating a new DHCP server with name in format `DHCPSERVER<dhcp_server_name>_Private` later in the flow. Isn't the main aim to check if the network regex matches the format?
> 
> cc @dharaneeshvrd

Correct, But currently there is no way to identify whether the network exist or not only from spec.

What we are trying to solve here is not allowing user to configure invalid network regex, 

For example:

Say cluster name is `powervs-cluster`, spec.DHCPSserver not set , the network is set to RegEx say `*.abc.$ `and network does not exist, 
The controller will create a network with `DHCPSERVERpowervs-cluster_Private` and set the network id in status, The next reconcilation continues by fetching network id from status.
The problem here is we are not at all caring what set in spec.Regex and if someone sees spec and actual resource created both are not at all related.

I think RegEx was introduced when we didn't what will be the DHCP created subnet name format. But now we have clear idea.

Should we support RegEx? or mark it as not supported?
@dharaneeshvrd @Prajyot-Parab  whats your thought.

dharaneeshvrd:
The same case will apply to network Name as well, when network name is given and network is not present, DHCP server that get's created will have either DHCP server name or cluster name not the network name. 
I think we have to decide on the priority. We have already discussed about this issue [here](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/1933#issuecomment-2510515529). Once we decide on the priority and what will happen in each case, we can allow regex too. And we can document it in API description I believe.
Since we are going to allow non DHCP network also, we have to consider that situation too before removing the support for regex IMO.

Prajyot-Parab:
IMO
1. we should create new resources only if name is set
  - If present fetch and use it
  - If not present, create resource using that name
2. we should avoid creating new resources when either of ID or regex is set (Their sole purpose should be to retrieve existing  resources if present and not create anything), just throw error if not present.
3. If all three are not set -> name, id, regex, then use default value set for name of that particular resource and use that for creation.

Karthik-K-N:
> IMO
> 
>     1. we should create new resources only if name is set
> 
> 
>     * If present fetch and use it
> 
>     * If not present, create resource using that name
> 
> 
>     2. we should avoid creating new resources when either of ID or regex is set (Their sole purpose should be to retrieve existing  resources if present and not create anything), just throw error if not present.
> 
>     3. If all three are not set -> name, id, regex, then use default value set for name of that particular resource and use that for creation.

Yeah, I think this should be the behavior, We created https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/1726 to handle these scenarios.
@Amulyam24  whats your thought.

Amulyam24:
> > IMO
> > ```
> > 1. we should create new resources only if name is set
> > 
> > 
> > * If present fetch and use it
> > 
> > * If not present, create resource using that name
> > 
> > 
> > 2. we should avoid creating new resources when either of ID or regex is set (Their sole purpose should be to retrieve existing  resources if present and not create anything), just throw error if not present.
> > 
> > 3. If all three are not set -> name, id, regex, then use default value set for name of that particular resource and use that for creation.
> > ```
> 
> Yeah, I think this should be the behavior, We created #1726 to handle these scenarios. @Amulyam24 whats your thought.

I agree. This makes sense and seems like the best way to handle new and existing resources. 

Regex should only be considered when the resource already exists.

Karthik-K-N:
Discussed this in CAPIBM meeting today and decide to take this up once we fix https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/2159

carmal891:
@Karthik-K-N @Amulyam24 @Prajyot-Parab @dharaneeshvrd Thank you for the clarifications. I will close this for now and pick up after the proposed changes in #2159

PR #2122, opened 2025-01-08
remove deprecated v1 provider id format

Prajyot-Parab:
**What this PR does / why we need it**:
- remove deprecated v1 provider id format

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1917 

**Special notes for your reviewer**:

/area provider/ibmcloud

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note
remove deprecated v1 provider id format
```

PR #2168, opened 2025-02-04
make powervs cluster setup with external cloud provider as default

Prajyot-Parab:
- make powervs cluster setup with external cloud provider as default
- Follow up on (https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2122#pullrequestreview-2577518690)
Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
make powervs cluster setup with external cloud provider as default
```


docs/book/src/developer/tilt.md
```diff
@@ -100,7 +101,7 @@ extra_args:
 
 ### 1.  Configuration to deploy workload cluster with external cloud controller manager
```
Amulyam24, line 102: This section seems repetitive, I think it can be removed as we made the default template addition in L87 to use CCM.

docs/book/src/topics/powervs/creating-a-cluster.md
```diff
@@ -131,31 +135,11 @@ following the steps below.
...
-  --worker-machine-count=1 \
-  --flavor=powervs | kubectl apply -f -
-  ```
-
 ### Deploy a PowerVS cluster with infrastructure creation
 
 #### Prerequisites: 
-- Set `EXP_CLUSTER_RESOURCE_SET` to true as the cluster will be deployed with external cloud provider which will create the resources to run the cloud controller manager.
-- Set the `provider-id-fmt` [flag](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/5e7f80878f2252c6ab13c16102de90c784a2624d/main.go#L168-L173) to `v2` via `PROVIDER_ID_FORMAT` environment variable.
+- Set `EXP_CLUSTER_RESOURCE_SET` to `true` as the cluster will be deployed with external cloud provider which will create the resources to run the cloud controller manager.
```
Amulyam24, line 141: Since this is added commonly at L42, can it be omitted in this and below sections to avoid repetition?

docs/book/src/developer/tilt.md
Prajyot-Parab: Fixed.

docs/book/src/topics/powervs/creating-a-cluster.md
Prajyot-Parab: Fixed.

docs/book/src/developer/tilt.md
Amulyam24: I think you removed it in the other file and not here

docs/book/src/topics/powervs/creating-a-cluster.md
```diff
@@ -176,37 +158,10 @@ following the steps below.
     --flavor=powervs-create-infra | kubectl apply -f -
   ```
 
-### Deploy a PowerVS cluster with external cloud provider
```
Amulyam24, line 179: do you think we need to keep a section here with mentioning for default template or the point in L39 will suffice?

Prajyot-Parab: point at L39 should be enough as per me, no need to explicitly mention it again to avoid repetition.

templates/README.md
```diff
@@ -20,15 +20,22 @@ clusterctl generate cluster ibm-vpc-1 --kubernetes-version v1.14.3 \
 
 ## Power VS
 
+> **Note:**
```
Amulyam24, line 23: would be better to add this in the start as it is common for both VPC and PowerVS

Amulyam24 (approved):
/lgtm

PR #2269, opened 2025-03-24
Fix priority between network and DHCP server

### LINKED ISSUE #2159: Priority to network name or DHCP server name while creating the DHCP server
/kind bug
/area provider/ibmcloud

[A clear and concise description of what the bug is.]

Address the issue described here https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/1933#issuecomment-2510515529

**Environment:**

- Cluster-api version:
- Minikube/KIND version:

dharaneeshvrd:
Fix priority between network and DHCP server
Do not allow to set both network and DHCP server details which will create conflict while creating new DHCP server on whether to use network name or DHCP server name

Fixes #2159 

/area provider/ibmcloud

**Release note**:

```release-note
Fix priority between network and DHCP server
```

### DIFF
--- a/api/v1beta2/ibmpowervscluster_types.go
+++ b/api/v1beta2/ibmpowervscluster_types.go
@@ -42,7 +42,7 @@ type IBMPowerVSClusterSpec struct {
 	// 1. in the case of DHCPServer.Name is not set the name will be DHCPSERVER<CLUSTER_NAME>_Private.
 	// 2. if DHCPServer.Name is set the name will be DHCPSERVER<DHCPServer.Name>_Private.
 	// when Network.ID is set, its expected that there exist a network in PowerVS workspace with id or else system will give error.
-	// when Network.Name is set, system will first check for network with Name in PowerVS workspace, if not exist network will be created by DHCP service.
+	// when Network.Name is set, system will first check for network with Name in PowerVS workspace, if not exist system will check DHCP network with given Network.name, if that also not exist, it will create a new DHCP service and name will be DHCPSERVER<Network.Name>_Private.
 	// Network.RegEx is not yet supported and system will ignore the value.
 	Network IBMPowerVSResourceReference `json:"network"`
 
--- a/cloud/scope/powervs_cluster.go
+++ b/cloud/scope/powervs_cluster.go
@@ -872,10 +872,11 @@ func (s *PowerVSClusterScope) createServiceInstance(ctx context.Context) (*resou
 }
 
 // ReconcileNetwork reconciles network
-// If only IBMPowerVSCluster.Spec.Network is set, network would be validated and if exists already will get used as cluster’s network or a new network will be created via DHCP service.
+// If only IBMPowerVSCluster.Spec.Network is set, network would be validated and if exists already will get used as cluster’s network or DHCP network would be validated with this name if not exits then a new network will be created via DHCP service.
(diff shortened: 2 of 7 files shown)


api/v1beta2/ibmpowervscluster_webhook_test.go
```diff
@@ -68,6 +68,36 @@ func TestIBMPowerVSCluster_create(t *testing.T) {
...
+			name: "Should error if both Network id and DHCP are set",
+			powervsCluster: &IBMPowerVSCluster{
+				Spec: IBMPowerVSClusterSpec{
+					ServiceInstanceID: "capi-si-id",
+					Network: IBMPowerVSResourceReference{
+						ID: ptr.To("capi-net-id"),
+					},
+					DHCPServer: &DHCPServer{
+						Name: ptr.To("capi-dhcp"),
+					},
```
Karthik-K-N, line 96: May lets also add test with network name - dhcp id , network id -dhcp id combination?

Karthik-K-N (commented):
https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/blob/c4171f21acc320fdfe17b812c98be908653134f7/cloud/scope/powervs_cluster.go#L870, This comment needs an update as now we are not allowing to set both network and DHCP details in spec.

dharaneeshvrd:
Addressed the comments, ptal @Karthik-K-N

Karthik-K-N (commented):
Do you think is there a way to handle/inform about this scenario, Correct me if my understanding is wrong.

What if user sets network.Name="abc" the network "abc" does not exist in cloud. I think our controller will create a DHCP network `DHCPSERVERabc_Private`. Now user has set "abc" as network in spec but actually its `DHCPSERVERabc_Private`

Karthik-K-N:
True, Thats the only thing we can do, along with that should we update spec.Network.Name with the newly created DHCP network name? will that be allowed?

dharaneeshvrd:
> should we update spec.Network.Name with the newly created DHCP network name? will that be allowed?

Can you please elaborate this? Do you mean to update the spec once we created the DHCP network? Just wondering since we won't follow that generally right!

Karthik-K-N:
> > should we update spec.Network.Name with the newly created DHCP network name? will that be allowed?
> 
> Can you please elaborate this? Do you mean to update the spec once we created the DHCP network? Just wondering since we won't follow that generally right!

True we wont follow it, Just thinking.

Karthik-K-N:
Or else how about strict validation like ID, when network.Name is set it should exist or else return error?

Most of the time we ommit network field and controller will autocreate the resource so it wont be a breaking change.

dharaneeshvrd:
Yeah that might be an option if we consider that field only for non DHCP networks. Anyway we are validating with that intention only. But only concern is that it would deviate from the other places on how we treat the name field of a resource.

Karthik-K-N:
> Yeah that might be an option if we consider that field only for non DHCP networks. Anyway we are validating with that intention only. But only concern is that it would deviate from the other places on how we treat the name field of a resource.

True that, Though the use case is very minimal, I am not so sure about showing wrong info about network name in spec. 

May be lets discuss this in the community meeting and we can either go with that or keep as it is and add note like now.

Karthik-K-N:
Discussed about it in the CAPIBM community meeting, We will keep the flow as is and not add a validation for network.Name like network.ID.

Need to setup another meeting to decide on how to allow  users to create DHCP or static network.

Karthik-K-N (commented):
/lgtm

Prajyot-Parab (approved):
/lgtm

PR #2310, opened 2025-04-15
Fix trivy scan flow

### LINKED ISSUE #2328: Trivy build failure
/kind bug
/area provider/ibmcloud

Trivy scans failing, as we are unable to build trivy. This is because the docker deps got bumped in project as part of some other package requirement, and the new docker version isn't compatible with trivy version we build.

Error
```
/home/runner/go/pkg/mod/github.com/aquasecurity/trivy@v0.58.2/pkg/fanal/image/daemon/docker.go:66:58: cannot use c.ImageSave (value of type func(ctx context.Context, imageIDs []string, saveOpts ..."github.com/docker/docker/client".ImageSaveOption) (io.ReadCloser, error)) as imageSave value in argument to imageOpener
```

- Trivy binary to build

We cannot use the latest version of trivy (with fix for above issue) as they have bumped go to 1.24 and our project is still at 1.23

Prajyot-Parab:
- Fix trivy scan flow

avoid build error -
```
Error: /home/runner/go/pkg/mod/github.com/aquasecurity/trivy@v0.58.2/pkg/fanal/image/daemon/docker.go:66:58: cannot use c.ImageSave (value of type func(ctx context.Context, imageIDs []string, saveOpts ..."github.com/docker/docker/client".ImageSaveOption) (io.ReadCloser, error)) as imageSave value in argument to imageOpener
```

Fixes #2328 

/area provider/ibmcloud

**Release note**:

```release-note
Fix trivy scan flow
```


Prajyot-Parab:
> @Prajyot-Parab, any reason for adding a new script for installing trivy?
> 
> currently we are updating the [version](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/79795d03b1c9f045d4ae5b7b450f8544ae92a103/hack/tools/go.mod#L8) and installing via make.

The present approach to build binary fails with following error - 
```
Error: /home/runner/go/pkg/mod/github.com/aquasecurity/trivy@v0.58.2/pkg/fanal/image/daemon/docker.go:66:58: cannot use c.ImageSave (value of type func(ctx context.Context, imageIDs []string, saveOpts ..."github.com/docker/docker/client".ImageSaveOption) (io.ReadCloser, error)) as imageSave value in argument to imageOpener
```

This is because the docker package getting installed is not compatible with the one trivy needs, (as some other package needs latest docker), so now I am switching to directly download the binary rather than building it, this is inline with CAPI approach.

Amulyam24:
Can we remove the duplicate code which will not be used?
https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/hack/tools/Makefile#L129-L131

mkumatag:
The challenge with this approach is that you will not be able to run this on ppc64le environment.

Prajyot-Parab:
> The challenge with this approach is that you will not be able to run this on ppc64le environment.

agreed. However this seems like the only way right now to get `trivy` scans to work for now (once the build issue is fixed in future release of trivy we can revert back to that approach)

Prajyot-Parab:
> Can we remove the duplicate code which will not be used? https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/hack/tools/Makefile#L129-L131

As per my prior comment, we will switch back to that approach once we have present build issue fixed in future `trivy` release, hence I was of opinion to keep the code for now.

Amulyam24 (approved):
/lgtm

Prajyot-Parab:
> Can we remove the duplicate code which will not be used? https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/hack/tools/Makefile#L129-L131

Yes removed, trivy is available for ppc64le as well so this approach should hold good in future too.

Amulyam24 (approved):
/lgtm

PR #2313, opened 2025-04-17
Initialise IBMPowerVSImage only when ImageRef is set in PowerVS machine controller

Amulyam24:
With the recent change, cluster installation was requeuing continuously with
```
IBMPowerVSImage is not ready yet, skipping reconciliation
```
since IBMPowerVSImage will not be nil and it expects [scope.IBMPowerVSImage.Status.Ready](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/controllers/ibmpowervsmachine_controller.go#L276-L277) to be true. It should not block and proceed to fetch the image details from `spec.IBMPowerVSImage`.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Initialise IBMPowerVSImage only when ImageRef is set in PowerVS machine controller
```

### DIFF
--- a/controllers/ibmpowervsmachine_controller.go
+++ b/controllers/ibmpowervsmachine_controller.go
@@ -144,8 +144,9 @@ func (r *IBMPowerVSMachineReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	ctx = ctrl.LoggerInto(ctx, log)
 
 	// Fetch the IBMPowerVSImage.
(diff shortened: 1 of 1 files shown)


Karthik-K-N (commented):
Thank you for the fix

/lgtm

mkumatag (approved):
/lgtm

I'm curious to know why didn't we hit this issue all these days? please document more information change triggered this flow now and causing the issue.

PR #2338, opened 2025-05-05
drop older k8s images

Prajyot-Parab:
- drop older k8s images

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
drop older k8s images
```


Karthik-K-N:
I think we dont have 1.30 images built due to known issue, I think we should document it somewhere so we know why its missing from here

Also did you clean up the COS bucket?

Prajyot-Parab:
> I think we dont have 1.30 images built due to known issue, I think we should document it somewhere so we know why its missing from here
> 
> Also did you clean up the COS bucket?

not yet, will do post this PR gets merged.

Amulyam24 (approved):
/lgtm

PR #2358, opened 2025-05-14
[release-0.10] bump go to 1.23.8 and add CVEs to trivyignore

Prajyot-Parab:
- bump go to 1.23.8 and add CVEs to trivyignore

The [weekly security scan task](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/actions/workflows/weekly-security-scan.yaml) has found two modules that have CVEs in our codebase.
- golang.org/x/crypto
- golang.org/x/net

To bump these to the fixed version (0.35 for crypto and 0.38 for net), we would need to bump the project to use 1.23. We don't generally do this as it'd force anyone consuming the module downstream to upgrade their go as well and that might not work for their projects or timelines. (Hence ignoring those explicitly)

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
bump go to 1.23.8 and add CVEs to trivyignore
```

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -1,2 +1,4 @@
 # According to govulncheck we are not using code that is affected by CVEs
+CVE-2025-22869
 CVE-2025-22870
+CVE-2025-22872
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ RELEASE_DIR := out
 OUTPUT_TYPE ?= type=registry
 
 # Go
-GO_VERSION ?=1.22.12
+GO_VERSION ?=1.23.8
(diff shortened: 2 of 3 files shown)


.trivyignore
```diff
@@ -1,2 +1,4 @@
 # According to govulncheck we are not using code that is affected by CVEs
+CVE-2025-22869
```
Amulyam24, line 2: @Prajyot-Parab, for future reference, can you please add description to PR on why these CVEs are ignored?

Prajyot-Parab: Done

Amulyam24 (approved):
/lgtm

PR #2364, opened 2025-05-20
Add tags to PowerVS Cluster resources and delete resources based on tags

arshadd-b:
Proposal for adding the tags to PowerVS Cluster resources and performing delete of resources on the bases of tags
Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Add tags to PowerVS Cluster resources and delete resources based on tags

```


Karthik-K-N (commented):
Thanks for the PR, Most of the things are good,lets update it bit more to make it better.

Amulyam24 (commented):
> Hi @Amulyam24 , It is the same functionality that we do from UI, adding tags to IBM Cloud resources if user wants to tag resources. It can help to user in resource management.

Got it, Thanks for the clarification.

A couple of suggestions have been added.

In the create flow diagram, how about we enhance the condition such as `cluster.spec.UserTags > 0` , then proceed to attach user provided tags.

docs/proposal/controller-and-user-tags.md
```diff
@@ -0,0 +1,70 @@
...
+PowerVS cluster reconciler sets controllercreated field whenever resource is created by controller, which was initially introduced to allow proper cleanup of newly created resource vs the use of existing resources.
+
+Though its working as expected and fulfills the purpose, we see some drawbacks.
+1. The field is initially set to true during the first reconciliation cycle when the resource is being created. In subsequent reconciliation loops, the field is not updated because the resource already exists in the cloud. This behavior introduces non-idempotency in the controller logic. As a result, if the initial reconciliation event is missed, the controller exhibits inconsistent behavior. Its against k8s principle of reconcilation of having level trigger rather than edge triggered.
+2. The status is expected to be created from spec, considering the scenario of backup and recover. If we move the spec to fresh management cluster which is setting the status, the controller created will be set as false as the resource already exists in cloud but it was created during its previous concilation.
+
+## Goal
```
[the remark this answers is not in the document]
arshadd-b: done

docs/proposal/controller-and-user-tags.md
```diff
@@ -0,0 +1,70 @@
...
+#### Proposed API changes
+UserTags field will contain list of tags that will be attached to resources.
+
+```shell
+
+// IBMPowerVSClusterSpec defines the desired state of IBMPowerVSCluster.
+type IBMPowerVSClusterSpec struct {
+
+	// UserTags contains list of tags needs to be attached to resources
+	UserTags []string `json:"tags,omitempty"`
```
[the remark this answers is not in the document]
arshadd-b: Have added new struct `Tag` having key value pair

Amulyam24 (commented):
minor nits, LGTM!

Amulyam24 (approved):
/lgtm

@arshadd-b, please squash the commits.

dharaneeshvrd (commented):
How do we handle existing resource which is created by controller in earlier runs which will have controllercreated tag?
E.g.
Due to some reason, resources created during earlier runs did nt get cleaned up properly, so when we process it, it would have the tag but it was nt created as part of current cluster's reconciliation. It would cause collision during destroy flow. We need to find a way to handle this situation too.

docs/proposal/controller-and-user-tags.md
```diff
@@ -0,0 +1,84 @@
...
+1. Deprecate and remove controllerCreated flag.
+2. Tag user provided resources.
+
+## Proposal
+This proposal presents adding two kinds of tags to the resources created by controller
+1. Controller tag
+2. User tags
+
+
+### Controller tag
```
dharaneeshvrd, line 26: How are we going to differentiate the vpc subnet or DHCP network tag for itself and TG connection when only TG connection is created by controller to use them during destroy flow?

arshadd-b: vpc subnet and DHCP network supports tagging,  So when these resources are newly created and will be tagged.
For  TG connection  we will check either vpc subnet or DHCP network since TG connnection doesn't support tagging.

dharaneeshvrd: Yes, my question is what are the labels you would use when TG connection is created by controller when vpc subnet and DHCP network also created by controller? Dont you need to add two label to the vpc subnet? one is to denote its own itself and another one for vpc TG connection. In that case dont you need to use a different tag format? I feel that also need to be documented properly here.

docs/proposal/controller-and-user-tags.md
```diff
@@ -0,0 +1,70 @@
...
+2. Provide user ability to set custom tags to cloud resources.
+
+## Proposal
+This proposal presents adding two kinds of tags to the resources created by controller
+1. Controller tag
+2. User tags
+
+
+### Controller tag
+A tag of format`powervs.cluster.x-k8s.io-resource-owner:<cluster_name>` will be added by the controller to newly created cloud resources marking the resource as created by controller. During deletion phase the system will look for the presence of the tag inorder to proceed with deletion or to keep as it is.
```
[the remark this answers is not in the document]
Karthik-K-N: cool thanks, I think having owned helps as later we can add more keywords if required, Then can we stick to this format
```
sigs.k8s.io/cluster-api-provider-ibmcloud/cluster/<cluster-name>: owned
```

docs/proposal/controller-and-user-tags.md
Karthik-K-N: I understand it like this
1. Transit Gateway supports tagging but its connections does not support tagging

Deletion flow

Case 1: All TG and its connections created by controller

If TG has tag attached then it means TG and its connections are created by controller  then we can directly delete all 3 resources(TG, PowerVS Connection, VPC Connections)

Case 2: Pre-existing TG but new connections

1. TG tag does not match the cluster in deletion's tag means we are not suppose to delete it
2. Now to delete PowerVS Connection, Get the network name of connection and check whether the network has the cluster in deletion's tag, If present delete 
3. Same for VPC.

Is my understanding correct or do you think is there any gap?

Karthik-K-N:
> How do we handle existing resource which is created by controller in earlier runs which will have controllercreated tag? E.g. Due to some reason, resources created during earlier runs did nt get cleaned up properly, so when we process it, it would have the tag but it was nt created as part of current cluster's reconciliation. It would cause collision during destroy flow. We need to find a way to handle this situation too.

Good point, I think as a transition plan I think we should check whether the resource has controllerCreated is set to true, In that case we should go ahead and create the tag. In this case we can completely depend on the tag to delete.

docs/proposal/controller-and-user-tags.md
arshadd-b: There is one more scenario 

case 3: What if only connections are created but TG and networks(vpc subnet and DHCP network) pre-existing.
In this case we have to delete only connections, to handle this case we need to add one tag to TG with below format
```sigs.k8s.io/cluster-api-provider-ibmcloud/cluster/<cluster-name>/TG: connection1, connection2```

Then before deleting we will check for this tag,
will be adding this in proposal

**Comment by dharaneeshvrd** [comment]
> Good point, I think as a transition plan I think we should check whether the resource has controllerCreated is set to true, In that case we should go ahead and create the tag. In this case we can completely depend on the tag to delete.

With this design it feels like we would need the controllercreated status as a dependency for attaching tags. But I feel we should have some idea on how can we move away from controllercreated status in future. Because after implementing this, if we always dependant on controllercreated status to handle certain cases, then what's the point of implementing tags?

File: docs/proposal/controller-and-user-tags.md (hunk 3 above)
Comment (arshadd-b) [reply docs/proposal/controller-and-user-tags.md:26]: have updated the  proposal and flow diagram according to this.

**Comment by dharaneeshvrd** [comment]
Please see will this work.
During tag attachment only we should ensure that the resource is tied to a particular IBMPowerVSCluster. Since we cant be sure whether it would get removed properly or not during deletion phase, we should ensure the uniqueness of the resource during attachment phase itself by adding UID of the k8s object in the tag. 
When a resource is tagged with a UID of a particular cluster it would always belong to that cluster until its properly cleaned up by the controller. In case of resource not being cleaned up properly due to any reason, when a new deployment tries to use the same cluster we can match that UID is not belong to the current cluster and we should not process it, since its not cleaned up properly during earlier deployment and user is responsible for cleaning such stale resources.

**Comment by Karthik-K-N** [comment]
> Please see will this work. During tag attachment only we should ensure that the resource is tied to a particular IBMPowerVSCluster. Since we cant be sure whether it would get removed properly or not during deletion phase, we should ensure the uniqueness of the resource during attachment phase itself by adding UID of the k8s object in the tag. When a resource is tagged with a UID of a particular cluster it would always belong to that cluster until its properly cleaned up by the controller. In case of resource not being cleaned up properly due to any reason, when a new deployment tries to use the same cluster we can match that UID is not belong to the current cluster and we should not process it, since its not cleaned up properly during earlier deployment and user is responsible for cleaning such stale resources.

Thanks @dharaneeshvrd for your inputs ,
 Attaching the UUID is a good idea but I fear attaching UUID to existing tag can make it very lengthy as UUID may be of 36 characters.
So should we attach two tags then? 
```
sigs.k8s.io/cluster-api-provider-ibmcloud/cluster/<cluster-name>: owned
cluster-uuid: UUID
```
We can make use of UUID only when necessary?

**Comment by Karthik-K-N** [comment]
> Will be able to attach multiple tags via single POST call?

We will be able to attach multiple tags for both PowerVS and VPC in single call from go client this is what I could see

VPC global tag client allows us to add multiple tags via tagNames field: https://github.com/IBM/platform-services-go-sdk/blob/25a5c32de880d93b94a0bb50ded94596015dc354/globaltaggingv1/global_tagging_v1.go#L742

PowerVS allows tagging while creating services, even it can be multiple tags: https://github.com/IBM/platform-services-go-sdk/blob/25a5c32de880d93b94a0bb50ded94596015dc354/resourcecontrollerv2/resource_controller_v2.go#L2540

> Can you please explain what are the situations which might not need UID?

When we are truly reusing the resources(when the resource already exist in the cloud), then we wont tag the resource at that time we need not attach either the UUID tag nor the cluster tag.

If the owned tag exist then only we compare with the UUID tag to make sure that the resource actually belongs to the deleting cluster.

I see the workflow as below

1. Creating a new resources with clusterName "powervs-cluster"

Following tags will be created
```
sigs.k8s.io/cluster-api-provider-ibmcloud/cluster/powervs-cluster: owned
powervs.cluster.x-k8s.io/cluster-uuid: 12134-5678
```

2. Creating a new cluster with already existing resources with clusterName "powervs-cluster-1"

for this use case for now , we will not attach any tags

3. Creating a cluster with already existing resources with same cluster name clusterName "powervs-cluster"

Even here we dont attach any new tags while creating
During deletion flow , since tag `sigs.k8s.io/cluster-api-provider-ibmcloud/cluster/powervs-cluster: owned` matches we will compare the uuid-tag `powervs.cluster.x-k8s.io/cluster-uuid: 12134-5678`

dharaneeshvrd:
> Creating a cluster with already existing resources with same cluster name clusterName "powervs-cluster"
Even here we dont attach any new tags while creating

Is it ok to use a stale resource for the new cluster? There might be chances for the cluster to be created by different user and can use same name.
How about during creation if we find the tag with uuid, intimate the user about this and not proceed with using the that resource for the cluster?

Remaining two cases sounds good to me.

Karthik-K-N:
> > Creating a cluster with already existing resources with same cluster name clusterName "powervs-cluster"
> > Even here we dont attach any new tags while creating
> 
> Is it ok to use a stale resource for the new cluster? There might be chances for the cluster to be created by different user and can use same name. How about during creation if we find the tag with uuid, intimate the user about this and not proceed with using the that resource for the cluster?
> 
> Remaining two cases sounds good to me.

Do you suggest to intimate user only if the resource is not created by them?

dharaneeshvrd (commented):
- In delete-tag-workflow.png, for load balancer deletion, need to fix the typo on resource name.
- Please mention what will happen in case of user passing certain tags and the resource is already exists. Do we need to attach the user provided tags? If yes, we have to clean the tags alone during deletion.

Amulyam24 (commented):
IMO, tagging a resource with UUID alone should suffice right?
As in either case we check with it at the end and decide the flow based on that. wdyt?

arshadd-b:
> IMO, tagging a resource with UUID alone should suffice right? As in either case we check with it at the end and decide the flow based on that. wdyt?

I think it should be okay to use only UUID alone 

So before creating we check if resource exist with same UUID, then error out to user. If it doesn't exist then create.
what do you think ?  @dharaneeshvrd  @Karthik-K-N

Karthik-K-N:
> > IMO, tagging a resource with UUID alone should suffice right? As in either case we check with it at the end and decide the flow based on that. wdyt?
> 
> I think it should be okay to use only UUID alone
> 
> So before creating we check if resource exist with same UUID, then error out to user. If it doesn't exist then create. what do you think ? @dharaneeshvrd @Karthik-K-N

I think that should be perfect, I can't recall why we didn't drop using name when we discoverd UUID is necessary

PR #2415, opened 2025-07-08
🌱  Remove external cloud-provider flag from apiserver as its removed in k8s v1.33

Karthik-K-N:

Since the flag is not supported, setting this flag in v1.33 leads to apiserver not starting as unknown flag error.

K8s reference: https://github.com/kubernetes/kubernetes/pull/130162

Fixes # part of #2416

/area provider/ibmcloud

**Release note**:

```release-note
Remove apiserver cloud-provider: external flag as its removed in k8s v1.33
```

### DIFF
--- a/templates/cluster-template-powervs-clusterclass.yaml
+++ b/templates/cluster-template-powervs-clusterclass.yaml
@@ -88,9 +88,6 @@ spec:
             certSANs:
             - ${IBMPOWERVS_VIP}
             - ${IBMPOWERVS_VIP_EXTERNAL}
-            extraArgs:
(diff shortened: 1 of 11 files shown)


Karthik-K-N:
> > /hold for pending tasks in parent issue
> 
> @Karthik-K-N, this can be merged right and the remaining tasks can be taken up in a new PR?
> 
> or would this be a breaking change with v1.32 which are the current CI images?

I think once the 1.33 image is imported into CI, we can merge this. What do you think?

Prajyot-Parab (approved):
/lgtm

PR #2449, opened 2025-07-29
Bumped ccm go version to be in line with main repo

anshuman-agarwala:
Bumped ccm go version to be in line with main repo

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note

```

### DIFF
--- a/hack/ccm/Makefile
+++ b/hack/ccm/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Go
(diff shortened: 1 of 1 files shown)


anshuman-agarwala:
I raised this because there was a warning when I ran `make verify`.
If the ccm go version is dependent on a different repo then we might want to remove the following check:
https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/Makefile#L577-L583

Issue #2574, opened 2025-12-17
Migrate to latest Kubebuilder version


Karthik-K-N:
Inorder to adhere to the CAPI contract of v1beta2, In CAPIBM we need to create v1beta3 api version, Inorder to do that we need to make use of kubebuilder but currently we are using v2 version of kubebuilder and its out of support the latest version is v4, We need migrate our project to use v4.

Issue #2401, opened 2025-06-24
Tracking issue for PowerVS v1Beta3


Karthik-K-N:
## Introduce the new API types

- [x] Update Kubebuilder verison to create new API version: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2576
- [x] Kubebuilder version upgrade follow up to change the api path: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2592
- [x] Refactor to keep vpc and powervs in separate packages: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2585

## PowerVS resources tasks

- [x] Create v1beta3 types & use them in controllers/webhooks

    - [x] Add PowerVS v1beta3 types: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2602
    - [x] Drop PowerVS v1beta1 types: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2608
    - [x] Promote CAPI v1beta2 conditions in CAPIBM v1beta3 api's: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2623
    - [x] Mark v1beta2 API's as deprecated and rename v1beta2 condition constants name: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2649
    - [x] Stop using v1beta2 conditions: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2688

 - [x] Implement v1beta2 [contract](https://cluster-api.sigs.k8s.io/developer/providers/migrations/v1.10-to-v1.11#cluster-api-contract-changes)

    - [x] Implement v1beta2 contract for PowerVS resources: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2652

### API Improvements

- [x] Enhancement for v1beta3 :  https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2743
- [x] Enahcnemet update to remove use of annoations: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2761 
- [x] Replace ServiceInstance with workspace: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2771
[...]

PR #2576, opened 2025-12-19
✨ Update kubebuilder version from v2 to v3

Karthik-K-N:

This PR updates the Kubebuilder version which is used to bootstrap the API from v2 to v3

In an efforts towards introducing v1beta3 API to adhere to the new CAPI contracts, the prerequisite is to update kubebuilder version, because the v2 version is no more supported and we can't make use of the kuebbuilder to create v1beta3 API.

In an efforts towards https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/2401
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/2574  

Kubebuilder v3 changes
move main.go from root to cmd/main.go
move controllers to internal/controller
Update PROJECT file with new v3 format.

/area provider/ibmcloud

**Release note**:

```release-note
Update project Kubebuilder version to v3 from v2
```


Karthik-K-N:
/hold cancel

Ready for review.

Dockerfile
```diff
@@ -34,9 +34,10 @@ RUN  --mount=type=cache,target=/root/.local/share/golang \
      go mod download
 
 # Copy the go source
-COPY main.go main.go
+COPY cmd/main.go cmd/main.go
 COPY api/ api/
-COPY controllers/ controllers/
+copy controllers/ controllers/
```
Amulyam24: typo? PTAL

Karthik-K-N: Yes, Fixed it thanks.

Dockerfile
```diff
@@ -34,9 +34,10 @@ RUN  --mount=type=cache,target=/root/.local/share/golang \
      go mod download
 
 # Copy the go source
-COPY main.go main.go
+COPY cmd/main.go cmd/main.go
 COPY api/ api/
 COPY controllers/ controllers/
+COPY internal/controllers/ internal/controllers/
```
Prajyot-Parab: on line 43 we `COPY internal/` do we need this as well?

controllers/alias.go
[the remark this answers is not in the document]
Karthik-K-N: Yes, It does not affect the CAPIBM repo  and need not do that but we are anticipating some major refactoring during the API bump and separate out powervs and vpc separately so thought anyhow we are refactoring import path and we can do this as well.
Also just in case we want to expose some internal methods to external package we can use this alias file to do so without causing any disruption.
It also stays in standard like how we handle webhooks currently by having alias.

Amulyam24 (approved):
/lgtm

Issue #2579, opened 2026-01-05
API changes required to bump github.com/IBM/vpc-go-sdk to 0.78.1


Amulyam24:
Build failure due to modified APIs
```
# sigs.k8s.io/cluster-api-provider-ibmcloud/api/v1beta2
api/v1beta2/types.go:326:71: undefined: vpcv1.NetworkACLRuleProtocolAllConst
make: *** [Makefile:169: vet] Error 1
```

Job link - https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-ibmcloud/2577/pull-cluster-api-provider-ibmcloud-build/2006005845594214400
PR ref - https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2577

Karthik-K-N:
I think its fixed via PR https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2894

Thank you
/close

PR #2583, opened 2026-01-06
⚠️ Update vpc-go-sdk module and replace SecurityGroupProtocolRule from all to icmp_tcp_udp

Karthik-K-N:

This PR updates the vpc-go-sdk version

The latest release of vpc-go-sdk contains some breaking changes for SecurityGroupRuleProtocal, Its changed from all to icmp_tcp_udp and we are no more able to use 'all' protocal.

From sdk release notes we can see
```
icmp_tcp_udp for ICMP, TCP, and UDP traffic (for API version 2025-12-08 or earlier, the value all is used instead of icmp_tcp_udp)
```

This PR changes the protocol usage from 'all' to 'icmp_tcp_udp'.

The release notes: https://cloud.ibm.com/docs/vpc?topic=vpc-api-change-log#9-december-2025

Fixes: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/2579

/area provider/ibmcloud

**Release note**:

```release-note
VPC Security Group Rule Protocol 'all' is replaced with 'icmp_tcp_udp'
```

### DIFF
--- a/api/v1beta2/types.go
+++ b/api/v1beta2/types.go
@@ -285,8 +285,8 @@ var (
 )
 
 const (
-	// VPCSecurityGroupRuleProtocolAllType is a string representation of the 'SecurityGroupRuleSecurityGroupRuleProtocolAll' type.
-	VPCSecurityGroupRuleProtocolAllType = "*vpcv1.SecurityGroupRuleSecurityGroupRuleProtocolAll"
+	// VPCSecurityGroupRuleProtocolIcmptcpudpType is a string representation of the 'SecurityGroupRuleProtocolIcmptcpudp' type.
+	VPCSecurityGroupRuleProtocolIcmptcpudpType = "*vpcv1.SecurityGroupRuleProtocolIcmptcpudp"
 
 	// VPCSecurityGroupRuleProtocolIcmpType is a string representation of the 'SecurityGroupRuleSecurityGroupRuleProtocolIcmp' type.
 	VPCSecurityGroupRuleProtocolIcmpType = "*vpcv1.SecurityGroupRuleSecurityGroupRuleProtocolIcmp"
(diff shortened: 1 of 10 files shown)


Karthik-K-N:
@dharaneeshvrd Please take a look when time permits, Thanks

api/v1beta2/types.go
```diff
@@ -519,7 +519,6 @@ type VPCSecurityGroupRuleRemote struct {
 
 // VPCSecurityGroupRulePrototype defines a VPC Security Group Rule's traffic specifics for a series of remotes (destinations or sources).
 // +kubebuilder:validation:XValidation:rule="self.protocol != 'icmp' ? (!has(self.icmpCode) && !has(self.icmpType)) : true",message="icmpCode and icmpType are only supported for VPCSecurityGroupRuleProtocolIcmp protocol"
-// +kubebuilder:validation:XValidation:rule="self.protocol == 'all' ? !has(self.portRange) : true",message="portRange is not valid for VPCSecurityGroupRuleProtocolAll protocol"
```
dharaneeshvrd: Respective change is not required?

Karthik-K-N:
> Overall changes looks good to me, Since this is a breaking change, hope we have an action plan to change its consumers as well, not sure is some component really consuming this, but just in case please make sure those gets updated as well.

Yes, we wont merge it , I have also requested the VPC team for review. We will merge post this.

Karthik-K-N:
> @Karthik-K-N The vpc-go-sdk update also introduced two new security group rule types:
> 
>     * `SecurityGroupRuleProtocolAny`
> 
>     * `SecurityGroupRuleProtocolIndividual`
> 
> 
> If someone defines a security group using these types, I believe it will cause errors. If you bump the vpc-go-sdk, I think you’ll also need to add support for these.

Thanks for the review, Can we take those feature addition in separate PR, For now  can we continue support the existing rules only?

Pacho20:
> Thanks for the review, Can we take those feature addition in separate PR, For now can we continue support the existing rules only?

It can be a separate PR but I think it needs to be added in the next couple of weeks.

Karthik-K-N:
> > Thanks for the review, Can we take those feature addition in separate PR, For now can we continue support the existing rules only?
> 
> It can be a separate PR but I think it needs to be added in the next couple of weeks.

Sure, Thanks, If you have bandwidth and can take that up work, that will be really helfull.

Pacho20:
> > > Thanks for the review, Can we take those feature addition in separate PR, For now can we continue support the existing rules only?
> > 
> > 
> > It can be a separate PR but I think it needs to be added in the next couple of weeks.
> 
> Sure, Thanks, If you have bandwidth and can take that up work, that will be really helfull.

Sure, I can do that.

PR #2610, opened 2026-01-28
:sparkles: Implement CRD migration

Karthik-K-N:

Recently we have added v1beta3 apiVersion for PowerVS resources and eventually we need to remove v1beta2 apiVersion as documented here https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/2611.

There are few important things to consider while removing apiVersion for a CRD, On a high level
1. Ensure all custom resources can still be read from etcd after the apiVersion is removed ("Storage version migration")
2. Remove the old apiVersion from managedFields of all custom resources ("ManagedField cleanup")

More info about it can be found here
1. https://github.com/kubernetes-sigs/cluster-api/issues/11894

CAPI already has a controller to handle these scenarios, In this PR we are configuring that controller to use it for IBMPowerVS resources.

More info can be found here: https://github.com/kubernetes-sigs/cluster-api/pull/11889

Fixes # 

/area provider/ibmcloud

**Release note**:

```release-note
Implement CRD migration
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -28,6 +28,7 @@ import (
 	// +kubebuilder:scaffold:imports
 	"github.com/spf13/pflag"
 
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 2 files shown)


internal/controllers/powervs/ibmpowervscluster_controller.go
```diff
@@ -79,7 +79,7 @@ type reconcileResult struct {
 }
 
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=ibmpowervsclusters,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=ibmpowervsclusters/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=ibmpowervsclusters/status,verbs=get;list;watch;update;patch
```
Amulyam24: Any reason for adding list & watch for the status for all resources?

internal/controllers/powervs/ibmpowervsmachine_controller.go
```diff
@@ -74,7 +74,7 @@ func init() {
 }
 
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=ibmpowervsmachines,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=ibmpowervsmachines/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=ibmpowervsmachines/status,verbs=get;patch;update
```
Prajyot-Parab: the original line was the same, does changing order matter?

PR #2705, opened 2026-04-09
🐛VPC: fix break logic in findOrCreateSecurityGroupRule

Pacho20:

The break statements inside the switch cases of the findOrCreateSecurityGroupRule method were intended to exit the inner loop. However, they only exited the switch block, causing the loop to continue unexpectedly.

A loop label has been introduced so the break statements correctly exit the intended loop.

**Release note**:

```release-note
fix break logic in findOrCreateSecurityGroupRule
```


Karthik-K-N:
@Pacho20 would like this PR to be kept open or should we close this?

Pacho20:
Please keep it open. Hopefully the PR will be ready this week.

PR #2788, opened 2026-06-02
Improved additional volumes update logic

anshuman-agarwala:
Currently some dangling volumes are created when a new cluster is created with additional volumes. 
This is happening due to a delay in when the additional volume status is updated. There is a timing issue where if two reconciles are called for the same CR, both the reconciles end up creating volumes because they both check for existing volumes in the status and find nothing.

With these code changes, if the scenario above occurs, one of the reconcile calls is guaranteed to fail because it will try to update an earlier version of the CR. When this happens the volumes deletion is initiated, thus avoiding the dangling volume issue.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Improved additional volumes update logic
```

### DIFF
--- a/internal/controllers/vpc/ibmvpcmachine_controller.go
+++ b/internal/controllers/vpc/ibmvpcmachine_controller.go
@@ -426,6 +426,7 @@ func patchIBMVPCMachine(ctx context.Context, patchHelper *v1beta1patch.Helper, i
 		clusterv1beta1.PausedV1Beta2Condition,
 	}})
 }
+
 func (r *IBMVPCMachineReconciler) reconcileAdditionalVolumes(ctx context.Context, machineScope *vpc.MachineScope) (ctrl.Result, error) {
 	log := ctrl.LoggerFrom(ctx)
 	// Return immediately if no additional volumes exist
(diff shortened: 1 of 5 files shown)


internal/controllers/vpc/ibmvpcmachine_controller.go
```diff
@@ -483,12 +484,25 @@ func (r *IBMVPCMachineReconciler) reconcileAdditionalVolumes(ctx context.Context
...
-				log.Error(err, "Could not update Machine status. Created Volume needs to be cleaned up manually", "VolumeID", volumeID)
+				log.Error(err, "Could not create new Volume")
 				errList = append(errList, err)
+			} else {
+				machineScope.IBMVPCMachine.Status.V1Beta2.AdditionalVolumeIDs[v] = volumeID
+				err = r.Status().Update(ctx, machineScope.IBMVPCMachine)
+				if err != nil {
+					log.Error(err, "Could not update Machine status. trying to clean up volume manually", "VolumeID", volumeID)
+					errList = append(errList, err)
+					if deleteError := machineScope.DeleteVolume(volumeID); deleteError != nil {
```
mkumatag, line 496: how this case is handles for the other providers?

internal/controllers/vpc/ibmvpcmachine_controller.go
```diff
@@ -483,12 +484,25 @@ func (r *IBMVPCMachineReconciler) reconcileAdditionalVolumes(ctx context.Context
 		} else {
 			// volume does not exist, create it and requeue so that it becomes available
 			volumeID, err := machineScope.CreateVolume(machineVolumes[v])
```
Karthik-K-N, line 486: Usually status update takes time to reflect unless we use direct API client. Instead of doing status update and delete on failure, how about just querying cloud and check whether the volume exist? if exist just adopt?

anshuman-agarwala: I think you've marked the wrong lines here. I'm assuming you mean https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2788/changes/BASE..efaea83fb6a293d210c8e7dcd0c8cc7415246418#diff-4d3d7bc2bb86603c23811772156d8fdbf1b1a037e94fd934f751872a3a0cb231R491-R493
I'm a little confused with what you're referring to as the direct API client. 
I've used the API client here to update the status. My understanding is that this will bypass the cache and call update against the API server directly.
Querying IBMCloud and checking for free volumes does not seem like a good idea because there volumes could be present there which are not attached but still used by the consumer for a different purpose. This is just one edge case I can think of, I'm pretty sure there are others as well.

Karthik-K-N: Got your point but I think it may not be the permanent fix, based on the severity of bug i leave this upto to @Prajyot-Parab or @Amulyam24 to take decision.
My point is even though you update the status object, There is no guarantee that it appears when you fetch the object again as we use [cached client](https://github.com/kubernetes-sigs/controller-runtime/blob/2bed742f15b43e30f1b5b16af7dd6c249fdff0c2/pkg/cluster/cluster.go#L58-L62). 

We have made a mistake of relaying too much on status before and we are eventually getting rid of it. 
Ideally we should use the volume tags and query the cloud and filter by tag and take the decision. ( this may be too much for this bug but for future)

aman4433:
@anshuman-agarwala Just checking, is there anything pending from your side on this PR?

anshuman-agarwala:
> @anshuman-agarwala Just checking, is there anything pending from your side on this PR?

Yes, there were some offline discussions we had regarding this that I should have updated here in the PR. 
We realized that even after adding this logic the overall flow is still brittle and timing issues can still come up.
The solution suggested by Karthik was to use Volume tags. I'm working on this approach right now and aim to have a PR up next week.

aman4433:
Hi @anshuman-agarwala, just wanted to check whether there's any update on the volume tags approach.

Karthik-K-N:
Created the issue for tracking https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/2968

closing this now due to inactivity, Feel free to open up when you resume work on this.

PR #2811, opened 2026-06-18
Extend IBMVPCMachineSpec to allow gen2 storage

Pacho20:
This PR extends IBMVPCMachineSpec to support a new sdp profile, which is a second-generation storage solution for VSIs offering significantly improved performance. Users can now define clusters that leverage this new storage. The changes are backward-compatible, adding a new profile value and an optional field without breaking the existing API.

**Release note**:

```release-note
Extend IBMVPCMachineSpec to allow gen2 storage
```


aman4433:
I noticed we now validate the profile-specific max size for `sdp` boot volumes, while additional volumes still only check that `sizeGiB` is non-zero. Should we also enforce the `sdp` max size for additional volumes here, or are we intentionally relying on the VPC API for additional volume size limits? Not a blocker if this is existing validation behavior.

api/vpc/v1beta1/ibmvpcmachine_types.go
```diff
@@ -95,16 +95,21 @@ type VPCVolume struct {
 	// Profile is the volume profile for the bootdisk, refer https://cloud.ibm.com/docs/vpc?topic=vpc-block-storage-profiles
 	// for more information.
 	// Default to general-purpose
-	// +kubebuilder:validation:Enum="general-purpose";"5iops-tier";"10iops-tier";"custom"
+	// +kubebuilder:validation:Enum="general-purpose";"5iops-tier";"10iops-tier";"custom";"sdp"
```
Amulyam24: @Pacho20, the API changes should be going into v1beta2 as that is the current version maintained for VPC. v1beta1 can be ignored.

anshuman-agarwala:
LGTM aside from the v1beta1 issue. 
Do you mind attaching screenshots of what ibmvpcmachine resources look like with the new volume types?

Pacho20:
> I noticed we now validate the profile-specific max size for `sdp` boot volumes, while additional volumes still only check that `sizeGiB` is non-zero. Should we also enforce the `sdp` max size for additional volumes here, or are we intentionally relying on the VPC API for additional volume size limits? Not a blocker if this is existing validation behavior.

@aman4433 This is intentional, and the asymmetry comes from the API itself:

- Boot volumes have a fixed platform maximum baked into the VPC API: 250 GiB for storage_generation: 1 and 32,000 GiB for storage_generation: 2 (sdp). Since that ceiling is a constant, we validate it client-side in the webhook.
- Additional (data) volumes have a profile-dependent maximum. The API defines it as the chosen profile's runtime capacity.max, not a fixed number (the 16000 values in the spec are only examples, and the minimum is 1 GiB). Because it varies per profile and can expand over time, there's no stable constant to enforce here, so we rely on the VPC API to reject out-of-range sizes for additional volumes.

api/vpc/v1beta1/ibmvpcmachine_types.go
Pacho20: Dropped the v1beta1 changes

Amulyam24:
> LGTM aside from the v1beta1 issue. Do you mind attaching screenshots of what ibmvpcmachine resources look like with the new volume types?

@Pacho20, can you share results from a test run?

Pacho20:
@anshuman-agarwala  @Amulyam24 Example for the IBMVPCMachineTemplate used in my tests:

```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
kind: IBMVPCMachineTemplate
metadata:
  name: test-capi-md-0
  namespace: default
spec:
  template:
    spec:
      profile: bx2-4x16
      zone: eu-de-1
      image:
        name: capibm-vpc-ubuntu-2404-kube-v1-33-0
      sshKeys:
      - name: test-ipi-vm-key
      primaryNetworkInterface:
        subnet: test-capi-worker-subnet
        securityGroups:
        - name: test-capi-worker-sg
      bootVolume:
        profile: sdp
        sizeGiB: 100
        iops: 4000
        bandwidth: 2000
      additionalVolumes:
      - name: test-capi-md-0-data-sdp
        profile: sdp
        sizeGiB: 500
        iops: 8000
        bandwidth: 3000
        deleteVolumeOnInstanceDelete: true
```
```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
kind: IBMVPCMachineTemplate
metadata:
  name: test-capi-control-plane
  namespace: default
spec:
  template:
    spec:
      bootVolume:
        bandwidth: 2000
        deleteVolumeOnInstanceDelete: true
        iops: 4000
        profile: sdp
        sizeGiB: 100
      image:
        name: capibm-vpc-ubuntu-2404-kube-v1-33-0
      loadBalancerPoolMembers:
      - loadBalancer:
          name: test-capi-load-balancer
        pool:
          name: test-capi-lbpool
        port: 6443
      primaryNetworkInterface:
        securityGroups:
        - name: test-capi-controlplane-sg
        subnet: test-capi-controlplane-subnet
      profile: bx2-4x16
      sshKeys:
      - name: test-ipi-vm-key
      zone: eu-de-1
```
As you can see the created VSIs match the new settings. 

image


image

Amulyam24 (approved):
/lgtm

Thank you @Pacho20!

cloud/scope/vpc/machine.go
```diff
@@ -447,6 +447,10 @@ func (m *MachineScope) volumeToVPCVolumeAttachment(ctx context.Context, volume *
 		bootVolume.Volume.Iops = core.Int64Ptr(volume.Iops)
 	}
 
+	if volume.Bandwidth != 0 {
```
Karthik-K-N, line 450: I hope 0 is not a valid value or it does not have any special meaning for bandwidth.

Karthik-K-N (commented):
LGTM

just  a small query.

cloud/scope/vpc/machine.go
Pacho20: Based on the [VPC API docs](https://cloud.ibm.com/docs/apis/vpc/latest#create-instance) 0 is not a valid value for bandwidth.

PR #2749, opened 2026-04-23
Dynamic sysType support for power VS

taliandre49:
**What this PR does / why we need it**:

Implements dynamic validation for PowerVS systemType field following CAPI architectural patterns. This replaces the previous webhook-based hardcoded validation with runtime validation against the PowerVS API, enabling automatic support for new systemTypes without code changes. Before these additions the design of CAPI required continuous maintenance when a new sysType is made available (i.e s1122).

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

Key Changes:

- Moved systemType validation from webhook (hardcoded list) to controller (dynamic API validation)
- Added validateSystemType() function in [controllers/validation.go](vscode-webview://0ir53id75arbuqj7kd4tipg5tdu9iaq6hhjuhvgs5kridndk0hdj/controllers/validation.go) that queries PowerVS API for supported types
- Updated webhook to only validate format pattern (^[a-z][0-9]+$), delegating reality checks to controller
- Added InvalidConfigurationReason condition for clear error reporting
- Exposed GetPISession() method in PowerVS service interface for API access

Benefits:

- Automatically supports new systemTypes added by PowerVS without requiring code updates
- Follows CAPI best practices: CRDs define shape, webhooks protect invariants, controllers enforce reality
- Provides clear error messages listing all supported systemTypes when validation fails
- Prevents VM creation with invalid configurations

Note: A separate PR is also open for these changes to backport into v0.12.0

/area provider/ibmcloud

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.
- no change to image version

**Release note**:

```release-note
PowerVS systemType validation is now dynamic, automatically supporting new system types added by IBM Cloud PowerVS without requiring provider updates. Validation moved from webhook (hardcoded list) to controller (API-based), following CAPI architectural patterns.
```

PR #2825, opened 2026-07-01
:bug: Fix panic while validating PER availibility for a zone

Karthik-K-N:

[getPowerVSClient](vscode-webview://0ot0tlvd4dqfcskp6il2tm8847un4qpfgr9c9jl69sq77f2rvfjg/cloud/scope/powervs/powervs_cluster.go:253) calls [powervs.NewService](vscode-webview://0ot0tlvd4dqfcskp6il2tm8847un4qpfgr9c9jl69sq77f2rvfjg/pkg/cloud/services/powervs/service.go:52), which only creates the session — it returns a &Service{session: session} with all clients nil. [WithClients](vscode-webview://0ot0tlvd4dqfcskp6il2tm8847un4qpfgr9c9jl69sq77f2rvfjg/pkg/cloud/services/powervs/service.go:79) (which initialises dataCenterClient) is never called in the cluster scope setup path.

So when ValidateZoneSupportsPER calls s.IBMPowerVSClient.GetDatatcenterDetails(zone) → s.dataCenterClient.Get(zone), dataCenterClient is nil → panic.

The dataCenterClient is zone-scoped (doesn't need a CloudInstanceID — the datacenter API is global). WithClients requires a CloudInstanceID which isn't known at cluster scope creation time (the workspace hasn't been provisioned yet). The fix is to initialise dataCenterClient separately, independent of WithClients.

Introduced as part of PR: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2749

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Fix panic while validating PER availibility for a zone
```

### DIFF
--- a/pkg/cloud/services/powervs/service.go
+++ b/pkg/cloud/services/powervs/service.go
@@ -70,8 +70,10 @@ func NewService(options ServiceOptions) (PowerVS, error) {
 		return nil, err
 	}
 
+	ctx := context.Background()
 	return &Service{
-		session: session,
+		session:          session,
+		dataCenterClient: instance.NewIBMPIDatacenterClient(ctx, session, ""),
 	}, nil
 }
 


Karthik-K-N:
@taliandre49 I think this bug is introduced as a part of dynamic system type validation, But you mentioned you have tested the create infra flow of cluster creation as well. Curious to know how this was escaped from testing.
Which method did you follow to create cluster. (just checking to make sure we can be more carefull)

Amulyam24 (approved):
/lgtm

Thank you for the quick fix @Karthik-K-N!

taliandre49:
Hey @Karthik-K-N , correct yes I ran the kubectl workflow you sent me as well as a local installation of a cluster and both passed, i wonder if this was introduced during the final rebase done 🤔. Thanks for the catch

Karthik-K-N:
> Hey @Karthik-K-N , correct yes I ran the kubectl workflow you sent me as well as a local installation of a cluster and both passed, i wonder if this was introduced during the final rebase done 🤔. Thanks for the catch

Yeah, may be. Thank you.

PR #2837, opened 2026-07-09
changes to CAPI needed to support service Endpoint Overrides in the installer

taliandre49:
Cherry-picked from 46e20dee onto upstream/main. Resolved path conflicts from the cloud/scope/powervs/ refactor and api/powervs/v1beta3 migration. These are the CAPI changes needed to support the service endpoint overrides in the installer.

Changes:
- Use FetchEndpoints (ID-only) instead of deprecated FetchPVSEndpoint in powervs_image.go and powervs_machine.go scope constructors to ensure PowerVS service URL override is applied regardless of region string.
- Initialise ResourceControllerV2Options in serviceOption in powervs_image.go and powervs_machine.go so RC endpoint overrides are propagated.
- Wire GetIAMAuthenticator into NewPowerVSImageScope so IAM endpoint (AUTH_URL) overrides are respected by the PowerVS session.
- Pass AUTH_URL into IamAuthenticator.URL in authenticator.go.
- Make COS NewService read AUTH_URL from authenticator properties so the IAM token endpoint can be overridden for staging environments.
- Fix structured log keys in ibmpowervsimage_controller.go (imageID, imageState) to satisfy structured logging linting.

This PR makes the PowerVS machine, image, and cluster scope constructors correctly
honour service endpoint overrides (e.g. staging/test environment URLs) passed in
via `ServiceEndpoint`, and ensures those overrides flow through to both the IAM
authenticator and the COS client.

Four concrete bugs are fixed:

1. **`FetchPVSEndpoint` silently dropped PowerVS endpoint overrides** in
   `NewPowerVSImageScope` and `NewMachineScope`. `FetchPVSEndpoint` (deprecated)
   matches on *both* a region string *and* a service ID. When the region derived
   from `serviceInstance.RegionID` did not round-trip through
   `ConstructRegionFromZone` to exactly the same string stored in the
   `ServiceEndpoint` list, it returned `""`, leaving the session URL as the
   production default. This caused the PowerVS session to mint a `bluemix` CRN
   instead of a `staging` CRN, which then made the PowerVS API validate bearer
   tokens against `iam.cloud.ibm.com/identity/keys` instead of
   `iam.test.cloud.ibm.com/identity/keys`, causing auth failures in the installer
   on staging. Fixed by switching to `FetchEndpoints` (ID-only match) in both
   scopes, consistent with how `ClusterScope.getPowerVSClient` already works.

2. **`ResourceControllerServiceOptions` was zero-initialised** in
   `NewPowerVSImageScope` and `NewMachineScope` (`var serviceOption
   resourcecontroller.ServiceOptions`), leaving `ResourceControllerV2Options` as
   `nil`. This prevented RC endpoint URL overrides from being applied. Fixed by
   eagerly initializing `ResourceControllerV2Options` to match the pattern already
   established in `ClusterScope`.

3. **IAM authenticator did not pick up `AUTH_URL`** from external configuration
   (e.g. `IBMCLOUD_AUTH_URL` env var / credentials file). The `IamAuthenticator`
   was built with only `ApiKey` set, so all token requests always went to the
   production IAM endpoint. Fixed by wiring `props["AUTH_URL"]` into
   `IamAuthenticator.URL`, and by calling `GetIAMAuthenticator()` in
   `NewPowerVSImageScope` to pass the resulting authenticator into the PowerVS
   session options.

4. **COS `NewService` hard-coded the production IAM token URL**. Fixed by reading
   `AUTH_URL` from `authenticator.GetProperties()` at runtime and appending
   `/identity/token`, with a fallback to the original production default when the
   property is absent.

Additionally, two unstructured `log.Info` calls in `ibmpowervsimage_controller.go`
that were missing key-value pairs (`log.Info("ImageID", value)` instead of
`log.Info("ImageID", "imageID", value)`) are corrected to satisfy structured
logging requirements.

### DIFF
--- a/pkg/cloud/services/cos/service.go
+++ b/pkg/cloud/services/cos/service.go
@@ -29,12 +29,14 @@ import (
 	"github.com/IBM/ibm-cos-sdk-go/aws/request"
 	cosSession "github.com/IBM/ibm-cos-sdk-go/aws/session"
 	"github.com/IBM/ibm-cos-sdk-go/service/s3"
+
+	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/cloud/services/authenticator"
 )
 
 // iamEndpoint represent the IAM authorisation URL.
 const (
-	iamEndpoint  = "https://iam.cloud.ibm.com/identity/token"
-	cosURLDomain = "cloud-object-storage.appdomain.cloud"
+	iamEndpointDefault = "https://iam.cloud.ibm.com/identity/token"
+	cosURLDomain       = "cloud-object-storage.appdomain.cloud"
 )
 
 // Service holds the IBM Cloud Resource Controller Service specific information.
@@ -120,6 +122,13 @@ func NewService(options ServiceOptions, apikey, serviceInstance string) (Cos, er
 			TLSHandshakeTimeout:   10 * time.Second,
 			ExpectContinueTimeout: 1 * time.Second,
 		},
+	}
+	// Get IAM endpoint from authenticator properties, fallback to default.
+	iamEndpoint := iamEndpointDefault
+	if props, err := authenticator.GetProperties(); err == nil {
+		if authURL := props["AUTH_URL"]; authURL != "" {
+			iamEndpoint = authURL + "/identity/token"
+		}
 	}
 	options.Config.Credentials = ibmiam.NewStaticCredentials(aws.NewConfig(), iamEndpoint, apikey, serviceInstance)
 
--- a/pkg/cloud/services/resourcecontroller/service.go
+++ b/pkg/cloud/services/resourcecontroller/service.go
@@ -18,6 +18,8 @@ package resourcecontroller
 
 import (
 	"fmt"
+	"os"
+	"strings"
 
 	"github.com/IBM/go-sdk-core/v5/core"
 	"github.com/IBM/platform-services-go-sdk/resourcecontrollerv2"
@@ -44,7 +46,20 @@ const (
 	// CosResourcePlanID is IBM COS plan id, can be retrieved using ibmcloud cli
 	// ibmcloud catalog service cloud-object-storage.
 	CosResourcePlanID = "1e4e33e4-cfa6-4f12-9016-be594a6d5f87"
+	// CosResourceLitePlanID is IBM COS Lite plan id (free, staging/non-paid accounts).
+	// Can be retrieved using: ibmcloud catalog service cloud-object-storage.
+	CosResourceLitePlanID = "2fdf0c08-2d32-4f46-84b5-32e0c92fffd8"
 )
+
+// GetCOSResourcePlanID returns the appropriate COS plan ID based on the IBMCLOUD_STAGING
+// environment variable. Staging (non-paid) accounts require the Lite plan; production uses Standard.
+func GetCOSResourcePlanID() string {
+	v := strings.ToLower(os.Getenv("IBMCLOUD_STAGING"))
+	if v == "true" || v == "1" || v == "yes" || v == "on" {
+		return CosResourceLitePlanID
+	}
+	return CosResourcePlanID
+}
 
 // InstanceFilter defines the filters for searching resource instances.
 type InstanceFilter struct {
@@ -227,7 +242,6 @@ func (s *Service) GetResourceInstanceByFilter(filter InstanceFilter) (*resourcec
 
 		serviceInstances, _, err := s.client.ListResourceInstances(listServiceInstanceOptions)
 		if err != nil {
-			fmt.Printf("error listing instances")
 			return false, "", err
 		}
 
--- a/cloud/scope/powervs/powervs_cluster.go
+++ b/cloud/scope/powervs/powervs_cluster.go
@@ -2363,7 +2363,7 @@ func (s *ClusterScope) reconcileCOSReference(_ context.Context, ref infrav1.Reso
 		filter := resourcecontroller.InstanceFilter{
 			Name:           ref.Name,
 			ResourceID:     resourcecontroller.CosResourceID,
-			ResourcePlanID: resourcecontroller.CosResourcePlanID,
+			ResourcePlanID: resourcecontroller.GetCOSResourcePlanID(),
 		}
 		instance, err := s.ResourceClient.GetResourceInstanceByFilter(filter)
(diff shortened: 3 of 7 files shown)


taliandre49:
@Prajyot-Parab could you please add needed labels

Prajyot-Parab:
@Karthik-K-N can you please review this one.

taliandre49:
@Karthik-K-N could you please take a look and add the necessary labels, thank you!!

taliandre49:
Related PR: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2864 
 PR is needed in order to vendor changes properly into Openshift installer main

Prajyot-Parab:
@taliandre49 please fix the conflicts

mjturek:
@Prajyot-Parab please take another look

pkg/cloud/services/resourcecontroller/service.go
```diff
@@ -44,8 +46,21 @@ const (
...
 
+// GetCOSResourcePlanID returns the appropriate COS plan ID based on the IBMCLOUD_STAGING
+// environment variable. Staging (non-paid) accounts require the Lite plan; production uses Standard.
+func GetCOSResourcePlanID() string {
+	v := strings.ToLower(os.Getenv("IBMCLOUD_STAGING"))
+	if v == "true" || v == "1" || v == "yes" || v == "on" {
+		return CosResourceLitePlanID
+	}
+	return CosResourcePlanID
+}
```
prb112, line 62: Why not just check if it exists?

taliandre49: latest push is updated to simplify this to use an existence check

pkg/cloud/services/cos/service.go
```diff
@@ -178,6 +181,15 @@ func NewService(options ServiceOptions, apikey, serviceInstance string) (Cos, er
 			ExpectContinueTimeout: 1 * time.Second,
 		},
 	}
+	// Get IAM endpoint from authenticator properties, fallback to default.
+	iamEndpoint := iamEndpointDefault
+	if props, err := authenticator.GetProperties(); err == nil {
+		if authURL := props["AUTH_URL"]; authURL != "" {
+			authURL = strings.TrimSuffix(authURL, "/identity/token")
+			authURL = strings.TrimSuffix(authURL, "/")
+			iamEndpoint = authURL + "/identity/token"
```
Karthik-K-N, line 190: Can we please log this iamEndpoint here so it will become easy for debug in future.

Karthik-K-N (changes requested):
Is this change is sufficient, or do we need to add support for any other services?

pkg/cloud/services/cos/service.go
taliandre49: log added!

taliandre49:
@Karthik-K-N These were all the changes identified that were needed for the successful cluster deployment. Addressed feedback

pkg/cloud/services/cos/service.go
```diff
@@ -178,6 +182,16 @@ func NewService(options ServiceOptions, apikey, serviceInstance string) (Cos, er
...
+		}
+	}
+	klog.V(4).Infof("COS IAM endpoint: %s", iamEndpoint)
```
Karthik-K-N: This is not how we log in this repo, we use context logger, Looks like adding log is not straight forward, You can ignore and remove this , fix the failing test
Apart from this LGTM
/assign @Amulyam24

taliandre49: correct, this was the pattern i saw in the envtest,go file, to log items that dont have context.logger. Removing this in latest push, removing this should also pass the test as it seems the test failed due to a format lint error when importing the klog package.

taliandre49: latest push reflects these changes

pkg/cloud/services/resourcecontroller/service.go
```diff
@@ -44,8 +45,20 @@ const (
...
 	CosResourcePlanID = "1e4e33e4-cfa6-4f12-9016-be594a6d5f87"
+	// CosResourceLitePlanID is IBM COS Lite plan id (free, staging/non-paid accounts).
+	// Can be retrieved using: ibmcloud catalog service cloud-object-storage.
+	CosResourceLitePlanID = "2fdf0c08-2d32-4f46-84b5-32e0c92fffd8"
 )
 
+// GetCOSResourcePlanID returns the appropriate COS plan ID based on the IBMCLOUD_STAGING
+// environment variable. Staging (non-paid) accounts require the Lite plan; production uses Standard.
+func GetCOSResourcePlanID() string {
+	if os.Getenv("IBMCLOUD_STAGING") != "" {
```
Amulyam24: This variable should be added to [manager.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/config/manager/manager.yaml#L28-L33) to be consumed by the controller?

Amulyam24 (commented):
@taliandre49, can you please confirm if the changes have been tested?

Amulyam24:
/retitle Support overriding IAM and COS service endpoints

pkg/cloud/services/resourcecontroller/service.go
Karthik-K-N: You are correct I think we should add something like this
```
        env:
        - name: IBMCLOUD_STAGING
          value: "${IBMCLOUD_STAGING}"
```

taliandre49: Can confirm changes have been tested for this. Looking it the variable addition to the yaml!

taliandre49: env:
      - name: IBMCLOUD_STAGING
        value: "${IBMCLOUD_STAGING:-}"

The only concern i have is that the maybe to add the :- to  ensure clusterctl substitutes an empty string instead of failing when unset.

taliandre49: @Amulyam24 @Karthik-K-N  ^^

taliandre49: incorporated in latest push

taliandre49:
addressing failing tests

taliandre49:
@Amulyam24 @Karthik-K-N Feedback Adressed

PR #2791, opened 2026-06-04
Add CentOS Stream 10 Kubernetes v1.34.7 DHCP image for PowerVS

aman4433:
What this PR does / why we need it:
Adds the CentOS Stream 10 Kubernetes v1.34.7 DHCP-enabled PowerVS image 
to the machine image documentation and updates the infrastructure-creation 
cluster example to use the new DHCP image artifact.

The image was built using image-builder with a custom DHCP-enabled base OVA 
and validated using the PowerVS infrastructure-creation workflow.

Which issue(s) this PR fixes:
part of #2764

Special notes for your reviewer:
/area provider/ibmcloud

Please confirm that if this PR changes any image versions, then that's the 
sole change this PR makes.

Validation:
- Built DHCP-enabled CentOS Stream 10 image using image-builder
- Uploaded image to power-oss-bucket as capibm-powervs-centos-streams10-1-34-7-dhcp.ova.gz
- Validated cluster creation using cluster-template-powervs-create-infra.yaml
- VM successfully obtained a DHCP lease and completed Kubernetes provisioning

Release note:
Add CentOS Stream 10 Kubernetes v1.34.7 DHCP image for PowerVS infrastructure-creation deployments

PR #2856, opened 2026-07-17
Use ds.meta_data.name for hostname in PowerVS create-infra template

aman4433:
## What this PR does / why we need it:

`templates/cluster-template-powervs-create-infra.yaml` currently uses `{{ v1.local_hostname }}` to configure the node hostname.

During manual validation of the PowerVS `create-infra` workflow, we observed intermittent hostname mismatches on some PowerVS VMs. In those cases, `v1.local_hostname` resolved to an internal PowerVS-generated hostname instead of the expected CAPI machine name. This prevented the Cloud Controller Manager (CCM) from matching the Kubernetes Node with the corresponding PowerVS instance, leaving the `providerID` unset.

Using `{{ ds.meta_data.name }}` consistently produced the expected CAPI machine name during our validation and avoided the hostname mismatch. On images where `v1.local_hostname` already resolves to the expected value (for example, our CentOS Stream 9 validation), this change is effectively a no-op, as both fields resolve to the same hostname.

This change is intentionally scoped to `templates/cluster-template-powervs-create-infra.yaml` only. The base templates (`templates/bases/powervs/*.yaml`) and other generated templates are left unchanged while the underlying hostname behavior is investigated further.

Ref: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2791#issuecomment-4978351218

## Special notes for your reviewer:

/area provider/ibmcloud

This is a targeted mitigation for the `create-infra` template used for manual validation and the packaged release flavor. It is **not** intended to address the underlying cause of why `v1.local_hostname` and `ds.meta_data.name` can differ on some PowerVS instances. That investigation is ongoing.

## Release note:

```release-note
Use `ds.meta_data.name` instead of `v1.local_hostname` in the PowerVS `create-infra` cluster template to avoid intermittent hostname mismatches observed during cluster creation.
```

### DIFF
--- a/templates/cluster-template-powervs-create-infra.yaml
+++ b/templates/cluster-template-powervs-create-infra.yaml
@@ -91,7 +91,7 @@ spec:
           value: external
         - name: eviction-hard
           value: nodefs.available<0%,nodefs.inodesFree<0%,imagefs.available<0%
-        name: '{{ v1.local_hostname }}'
+        name: '{{ ds.meta_data.name }}'
     joinConfiguration:
       nodeRegistration:
         criSocket: /var/run/containerd/containerd.sock
@@ -100,13 +100,13 @@ spec:
           value: external
         - name: eviction-hard
           value: nodefs.available<0%,nodefs.inodesFree<0%,imagefs.available<0%
-        name: '{{ v1.local_hostname }}'
+        name: '{{ ds.meta_data.name }}'
     preKubeadmCommands:
-    - hostname "{{ v1.local_hostname }}"
+    - hostname "{{ ds.meta_data.name }}"
     - echo "::1         ipv6-localhost ipv6-loopback" >/etc/hosts
     - echo "127.0.0.1   localhost" >>/etc/hosts
-    - echo "127.0.0.1   {{ v1.local_hostname }}" >>/etc/hosts
-    - echo "{{ v1.local_hostname }}" >/etc/hostname
+    - echo "127.0.0.1   {{ ds.meta_data.name }}" >>/etc/hosts
+    - echo "{{ ds.meta_data.name }}" >/etc/hostname
   machineTemplate:
     spec:
       infrastructureRef:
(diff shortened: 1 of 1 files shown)


Prajyot-Parab:
@aman4433 this one is generated template, you might need to change in base yamls from where this is generated.

aman4433:
> @aman4433 this one is generated template, you might need to change in base yamls from where this is generated.

Thanks for pointing that out, @Prajyot-Parab.

I wanted to double-check before changing the PR, so I looked for where `templates/cluster-template-powervs-create-infra.yaml` is generated. I checked the `generate-templates` target, the `hack/ scripts`, CI workflows, and looked for a `clusterctl-settings.json` or a corresponding kustomize source, but I couldn't find a generation path for this template. It also looks like this file has been edited directly in previous PRs.

Could you point me to the base template or generation mechanism you're referring to? If there is a source of truth for this template, I'm happy to update that instead and regenerate the output.

Prajyot-Parab:
@aman4433 Is this change only specific to create-infra template? If yes then the changes are good, however I was wondering if this change could make sense in other templates where it is referred.

aman4433:
Thanks, Prajyot.

At the moment, I intended to scope this change only to the create-infra template.

The hostname mismatch was observed during manual validation of the create-infra workflow, and that's the only path where I've reproduced the issue so far. I haven't yet validated whether the same behavior occurs in the other PowerVS templates, so I didn't want to broaden the change without evidence.

If we confirm that `v1.local_hostname` can resolve differently in the other templates as well, I'd be happy to follow up with a separate PR to update those after validating them.

Amulyam24 (commented):
@aman4433, is there any documentation on the use of `ds.meta_data.name`?

aman4433:
> @aman4433, is there any documentation on the use of `ds.meta_data.name`?

Thanks, Amulya.

I looked for PowerVS-specific documentation describing the ConfigDrive `meta_data.json` schema, but couldn't find any public documentation that defines the semantics of `ds.meta_data.name` for the PowerVS datasource. Cloud-init documents `ds` as datasource-specific metadata and `v1.local_hostname` only as "the internal or local hostname of the system"; it doesn't document `v1.local_hostname` as representing the cloud instance name or VM name.

After your question, I did another round of validation on freshly provisioned PowerVS VMs. On both CentOS Stream 9 and CentOS Stream 10, `v1.local_hostname` and `ds.meta_data.name` resolved to the same value, matching the expected CAPI machine name, so I wasn't able to reproduce the original mismatch.

At this point, it looks like the original behavior is intermittent.

Amulyam24:
Thanks for verifying Aman, is this good to close?

PR #2864, opened 2026-07-28
CAPI changes to support staging Overrides, backport release-0.12

taliandre49:
These are the CAPI changes needed to support the service endpoint overrides in the installer. Backport of [#2837](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2837) Backport is needed to revendor the installer which pins version 12.2

Changes:

- Use FetchEndpoints (ID-only) instead of deprecated FetchPVSEndpoint in powervs_image.go and powervs_machine.go scope constructors to ensure PowerVS service URL override is applied regardless of region string.
- Initialise ResourceControllerV2Options in serviceOption in powervs_image.go and powervs_machine.go so RC endpoint overrides are propagated.
- Wire GetIAMAuthenticator into NewPowerVSImageScope so IAM endpoint (AUTH_URL) overrides are respected by the PowerVS session.
- Pass AUTH_URL into IamAuthenticator.URL in authenticator.go.
- Make COS NewService read AUTH_URL from authenticator properties so the IAM token endpoint can be overridden for staging environments.
- Fix structured log keys in ibmpowervsimage_controller.go (imageID, imageState) to satisfy structured logging linting.

What this PR does / why we need it:

This PR makes the PowerVS machine, image, and cluster scope constructors correctly
honour service endpoint overrides (e.g. staging/test environment URLs) passed in
via ServiceEndpoint, and ensures those overrides flow through to both the IAM
authenticator and the COS client.

Four concrete bugs are fixed:

1. FetchPVSEndpoint silently dropped PowerVS endpoint overrides in
NewPowerVSImageScope and NewMachineScope. FetchPVSEndpoint (deprecated)
matches on both a region string and a service ID. When the region derived
from serviceInstance.RegionID did not round-trip through
ConstructRegionFromZone to exactly the same string stored in the
ServiceEndpoint list, it returned "", leaving the session URL as the
production default. This caused the PowerVS session to mint a bluemix CRN
instead of a staging CRN, which then made the PowerVS API validate bearer
tokens against iam.cloud.ibm.com/identity/keys instead of
iam.test.cloud.ibm.com/identity/keys, causing auth failures in the installer
on staging. Fixed by switching to FetchEndpoints (ID-only match) in both
scopes, consistent with how ClusterScope.getPowerVSClient already works.

2. ResourceControllerServiceOptions was zero-initialised in
NewPowerVSImageScope and NewMachineScope (var serviceOption resourcecontroller.ServiceOptions), leaving ResourceControllerV2Options as
nil. This prevented RC endpoint URL overrides from being applied. Fixed by
eagerly initializing ResourceControllerV2Options to match the pattern already
established in ClusterScope.

3. IAM authenticator did not pick up AUTH_URL from external configuration
(e.g. IBMCLOUD_AUTH_URL env var / credentials file). The IamAuthenticator
was built with only ApiKey set, so all token requests always went to the
production IAM endpoint. Fixed by wiring props["AUTH_URL"] into
IamAuthenticator.URL, and by calling GetIAMAuthenticator() in
NewPowerVSImageScope to pass the resulting authenticator into the PowerVS
session options.

4. COS NewService hard-coded the production IAM token URL. Fixed by reading
AUTH_URL from authenticator.GetProperties() at runtime and appending
/identity/token, with a fallback to the original production default when the
property is absent.

Additionally, two unstructured log.Info calls in ibmpowervsimage_controller.go
that were missing key-value pairs (log.Info("ImageID", value) instead of
log.Info("ImageID", "imageID", value)) are corrected to satisfy structured
logging requirements.

### DIFF
--- a/cloud/scope/powervs_image.go
+++ b/cloud/scope/powervs_image.go
@@ -30,6 +30,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-ibmcloud/api/v1beta2"
+	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/cloud/services/authenticator"
 	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/cloud/services/powervs"
 	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/cloud/services/resourcecontroller"
 	"sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/endpoints"
@@ -78,7 +79,9 @@ func NewPowerVSImageScope(ctx context.Context, params PowerVSImageScopeParams) (
 	scope.IBMPowerVSImage = params.IBMPowerVSImage
 
 	// Create Resource Controller client.
-	var serviceOption resourcecontroller.ServiceOptions
+	serviceOption := resourcecontroller.ServiceOptions{
+		ResourceControllerV2Options: &resourcecontrollerv2.ResourceControllerV2Options{},
+	}
 	// Fetch the resource controller endpoint.
 	rcEndpoint := endpoints.FetchEndpoints(string(endpoints.RC), params.ServiceEndpoint)
 	if rcEndpoint != "" {
@@ -132,11 +135,20 @@ func NewPowerVSImageScope(ctx context.Context, params PowerVSImageScopeParams) (
 		},
 	}
 
-	// Fetch the service endpoint.
-	if svcEndpoint := endpoints.FetchPVSEndpoint(endpoints.ConstructRegionFromZone(*res.RegionID), params.ServiceEndpoint); svcEndpoint != "" {
+	// Use FetchEndpoints (ID-only match) so that PowerVS service URL overrides are applied
+	// regardless of whether res.RegionID round-trips through ConstructRegionFromZone to the
+	// same region string stored in the ServiceEndpoint list.
+	if svcEndpoint := endpoints.FetchEndpoints(string(endpoints.PowerVS), params.ServiceEndpoint); svcEndpoint != "" {
 		options.IBMPIOptions.URL = svcEndpoint
 		log.V(3).Info("Overriding the default PowerVS service endpoint", "serviceEndpoint", svcEndpoint)
 	}
+
+	// Get the authenticator to ensure IAM endpoint overrides are respected.
+	auth, err := authenticator.GetIAMAuthenticator()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create authenticator: %w", err)
+	}
+	options.Authenticator = auth
 
 	c, err := powervs.NewService(options)
 	if err != nil {
--- a/controllers/ibmpowervsimage_controller.go
+++ b/controllers/ibmpowervsimage_controller.go
@@ -232,9 +232,9 @@ func reconcileImage(ctx context.Context, img *models.ImageReference, imageScope 
 		}
 
 		imageScope.SetImageID(image.ImageID)
-		log.Info("ImageID", imageScope.GetImageID())
+		log.Info("Setting image ID", "imageID", imageScope.GetImageID())
 		imageScope.SetImageState(image.State)
-		log.Info("ImageState", image.State)
+		log.Info("Current image state", "imageState", image.State)
 
 		switch imageScope.GetImageState() {
 		case infrav1.PowerVSImageStateQueued:
--- a/pkg/cloud/services/authenticator/authenticator.go
+++ b/pkg/cloud/services/authenticator/authenticator.go
@@ -70,8 +70,11 @@ func GetIAMAuthenticator() (*core.IamAuthenticator, error) {
 		fmt.Printf("ibmcloud api key is not provided, set %s environmental variable", "IBMCLOUD_API_KEY")
 	}
 
+	// URL is the IAM token endpoint base URL (e.g. https://iam.test.cloud.ibm.com).
+	// When empty the IBM SDK defaults to the production endpoint (https://iam.cloud.ibm.com).
 	auth := &core.IamAuthenticator{
 		ApiKey: apiKey,
+		URL:    props["AUTH_URL"],
 	}
 
 	return auth, nil
--- a/pkg/cloud/services/cos/service.go
+++ b/pkg/cloud/services/cos/service.go
@@ -20,6 +20,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"golang.org/x/net/http/httpproxy"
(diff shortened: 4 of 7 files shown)


cloud/scope/powervs_image.go
```diff
@@ -132,12 +135,26 @@ func NewPowerVSImageScope(ctx context.Context, params PowerVSImageScopeParams) (
...
-	// Fetch the service endpoint.
-	if svcEndpoint := endpoints.FetchPVSEndpoint(endpoints.ConstructRegionFromZone(*res.RegionID), params.ServiceEndpoint); svcEndpoint != "" {
+	// Use FetchEndpoints (matches on ID only) rather than the deprecated FetchPVSEndpoint
+	// (which also requires a region match). The cluster scope's getPowerVSClient uses the
+	// same ID-only lookup; using the region-aware variant here caused the URL override to be
+	// silently skipped when res.RegionID did not round-trip through ConstructRegionFromZone
+	// to the same region string that was stored in the ServiceEndpoint list, leaving the
+	// session URL as the production default and producing a "bluemix" CRN instead of
+	// "staging", which then caused the PowerVS API to validate the bearer token against
+	// iam.cloud.ibm.com/identity/keys instead of iam.test.cloud.ibm.com/identity/keys.
```
Karthik-K-N: I think in the comment lets only capture the current state, Previous state might be not needed

pkg/cloud/services/authenticator/authenticator.go
```diff
@@ -72,6 +72,7 @@ func GetIAMAuthenticator() (*core.IamAuthenticator, error) {
 
 	auth := &core.IamAuthenticator{
 		ApiKey: apiKey,
+		URL:    props["AUTH_URL"],
```
Karthik-K-N, line 77: what will happen when the URL is empty? is it by default prod?

pkg/cloud/services/cos/service.go
```diff
@@ -121,6 +122,13 @@ func NewService(options ServiceOptions, apikey, serviceInstance string) (Cos, er
 			ExpectContinueTimeout: 1 * time.Second,
 		},
 	}
+	// Get IAM endpoint from authenticator properties, fallback to default.
+	iamEndpoint := iamEndpointDefault
+	if props, err := authenticator.GetProperties(); err == nil {
+		if authURL := props["AUTH_URL"]; authURL != "" {
```
Karthik-K-N, line 130: dont we need to do any other validation? what if the AUTH_URL itself is set with /identity/token

pkg/cloud/services/resourcecontroller/service.go
```diff
@@ -44,8 +46,22 @@ const (
 	// CosResourcePlanID is IBM COS plan id, can be retrieved using ibmcloud cli
 	// ibmcloud catalog service cloud-object-storage.
 	CosResourcePlanID = "1e4e33e4-cfa6-4f12-9016-be594a6d5f87"
+
+	// CosResourceLitePlanID is IBM COS Lite plan id (free, staging/non-paid accounts).
+	// Can be retrieved using: ibmcloud catalog service cloud-object-storage.
+	CosResourceLitePlanID = "2fdf0c08-2d32-4f46-84b5-32e0c92fffd8"
```
Karthik-K-N, line 52: If you have any doc in handy, you can attach here for reference. Thanks for adding the ibmcloud cli reference.

pkg/cloud/services/authenticator/authenticator.go
taliandre49: correct yes this is the fall back that the authenticator uses for the optional url in the[sdk]( https://github.com/IBM/go-sdk-core/blob/v5.23.2/core/iam_authenticator.go):
```
/ url returns the authenticator's URL property after potentially initializing it.
func (authenticator *IamAuthenticator) url() string {
	authenticator.urlInit.Do(func() {
		if authenticator.URL == "" {
			// If URL was not specified, then use the default IAM endpoint.
			authenticator.URL = defaultIamTokenServerEndpoint
		} else {
			// Canonicalize the URL by removing the operation path if it was specified by the user.
			authenticator.URL = strings.TrimSuffix(authenticator.URL, iamAuthOperationPathGetToken)
		}
	})
	return authenticator.URL
}
```

pkg/cloud/services/cos/service.go
taliandre49: Validation added

pkg/cloud/services/resourcecontroller/service.go
taliandre49: Unfortunately i don't have a doc :( , i found the plan that needed to be targeted through the CLI however can find documentation on COS plans if needed!

controllers/ibmpowervsimage_controller.go
```diff
@@ -232,9 +232,9 @@ func reconcileImage(ctx context.Context, img *models.ImageReference, imageScope
 		}
 
 		imageScope.SetImageID(image.ImageID)
-		log.Info("ImageID", imageScope.GetImageID())
+		log.Info("ImageID", "imageID", imageScope.GetImageID())
```
Karthik-K-N: Setting image ID
```suggestion
		log.Info("Setting image ID", "imageID", imageScope.GetImageID())
```

controllers/ibmpowervsimage_controller.go
```diff
@@ -232,9 +232,9 @@ func reconcileImage(ctx context.Context, img *models.ImageReference, imageScope
...
 		imageScope.SetImageState(image.State)
-		log.Info("ImageState", image.State)
+		log.Info("ImageState", "imageState", image.State)
```
Karthik-K-N:
```suggestion
		log.Info("Current image state", "imageState", image.State)
```

pkg/cloud/services/cos/service.go
```diff
@@ -121,6 +123,15 @@ func NewService(options ServiceOptions, apikey, serviceInstance string) (Cos, er
...
+			authURL = strings.TrimSuffix(authURL, "/identity/token")
+			authURL = strings.TrimSuffix(authURL, "/")
```
Karthik-K-N, line 132: can't this be combined?

Karthik-K-N (commented):
LGTM

I will defer the final review to @Amulyam24 

It would be great if you can update the doc/book with the info on how to use staging if its missing.

pkg/cloud/services/cos/service.go
taliandre49: They definitely can, I've kept these as two separate TrimSuffix calls rather than chaining them since each line removes a distinct thing and reads more clearly that way. The IBM SDK's own iam_authenticator.go uses the same single-call-per-line style for the same normalization step. Thoughts??

pkg/cloud/services/resourcecontroller/service.go
```diff
@@ -44,8 +46,22 @@ const (
...
 )
 
+// GetCOSResourcePlanID returns the appropriate COS plan ID based on the IBMCLOUD_STAGING
+// environment variable. Staging (non-paid) accounts require the Lite plan; production uses Standard.
+func GetCOSResourcePlanID() string {
+	v := strings.ToLower(os.Getenv("IBMCLOUD_STAGING"))
```
Amulyam24, line 58: Where would this env value and the AUTH_URL be set?

Amulyam24 (commented):
Overall LGTM, a minor comment added.

@taliandre49, hope you have verified the changes.

pkg/cloud/services/resourcecontroller/service.go
taliandre49: IBMCLOUD_STAGING is set by the operator in their shell before invoking the installer or by runnig with the staging flag. The installer reads it via IsStagingMode(), derives the correct IAM URL from it, and explicitly forwards both IBMCLOUD_STAGING and IBMCLOUD_AUTH_URL to the CAPI controller subprocess in system.go. By the time GetCOSResourcePlanID() and authenticator.GetProperties() run inside the CAPI process, both env vars are already present in its environment

taliandre49:
@Amulyam24 @Karthik-K-N All tests passing could you please add other required labels  (lgtm, approved) when you get the chance, thank you!

Amulyam24 (commented):
/lgtm
/approve

Thank you @taliandre49 !

PR #2876, opened 2026-08-04
🌱 Consolidate resource name generator at one place

Karthik-K-N:

Fixes # Part of https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/2401

/area provider/ibmcloud

**Release note**:

```release-note
Consolidate resource name generator at one place
```

### DIFF
--- a/pkg/cloud/scope/powervs/resource_name.go
+++ b/pkg/cloud/scope/powervs/resource_name.go
@@ -1,0 +1,95 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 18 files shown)


pkg/cloud/scope/powervs/resource_name.go
```diff
@@ -0,0 +1,89 @@
...
+	suffix := string(rt)
+	if qualifier != "" {
+		suffix = fmt.Sprintf("%s-%s", rt, qualifier)
+	}
+
+	// +1 accounts for the "-" separator between base and suffix.
+	maxBase := resourceNameMaxLen - len(suffix) - 1
+	base := clusterName
+	if len(base) > maxBase {
+		base = base[:maxBase]
```
Amulyam24: This might lead to less readable resource names with abrupt base name or the base might end with a hyphen, combining the full name to, for ex: my-cluster--subnet-1

What about handling it in a more meaningful manner?

Karthik-K-N: I have made changes to remove the extra hyphen at the end if it present, Regarding truncate we can't do anything as its strict limit, But ideally we wont be having more than 63 character name.

Amulyam24 (approved):
/lgtm
/approve

Issue #2839, opened 2026-07-10
[PowerVS] Optimise VPC Load Balancer pool member addition to reduce sequential update_pending cycles during cluster installation


Neha-dot-Yadav:
/kind feature
/area provider/ibmcloud

## User Story
We would like the VPC Load Balancer pool member addition workflow to be optimised so that OpenShift cluster provisioning  on IBM Cloud PowerVS does not spend 15–25 minutes serially waiting for LB state transitions.

## Detailed Feature Description
Currently, the CAPI provider adds pool members one at a time via individual
`CreateLoadBalancerPoolMember` (`POST /load_balancers/{id}/pools/{pool_id}/members`)
calls in `powervs_machine.go`. IBM Cloud VPC LB enforces strictly serial processing,
each call triggers a full `active - update_pending - active` cycle, and no next call
is accepted until the LB returns to `active`.
With 4 machines × 2 pools × 2 LBs, this results in **16 sequential update_pending
cycles**, taking 15–25 minutes in total.
The goal of this issue is to explore possible code changes in the CAPI provider workflow to optimise this, potentially as part of the ongoing API enhancement work in CAPIBM.

## Context & Motivation
- Observed consistently across eu-gb (lon04, lon06) and eu-de (Frankfurt)
- Present since at least January 2026 across versions 4.19–5.0
- 15–25 min delay increases over all deploy time and sometimes causes failures.
- Even when installs succeed, the cluster is not fully stable when tests start.
- VPC team stated the delay is purely structural, neither retry frequency nor faulted members contribute. Each of the 16 additions forces its own full LB config push cycle.

Karthik-K-N:
Completed via https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/pull/2909
Thank you for creating the issue.
/close

PR #2883, opened 2026-08-05
fix(powervs): skip busy load balancers instead of erroring, cut requeue interval to 15s

aman4433:
<h2>What this fixes</h2>
<p>During PowerVS cluster provisioning, control-plane machines register into VPC load balancer pools. Each registration briefly locks the load balancer while it applies the change. The old code treated a locked LB as an error and gave up on the whole reconcile pass — including any <em>other</em> LB that was free the whole time. It also waited a flat 60 seconds before trying again, no matter how quickly the LB actually settled.</p>
<p>Result: with 2 LBs × 2 pools per machine, registering 3 control-plane machines took 15–25 minutes in production, occasionally blowing past the installer's bootstrap timeout.</p>
<h2>What changed</h2>
<ul>
<li>A load balancer in a transient state (<code>update_pending</code>, <code>create_pending</code>) is now skipped, not errored — the other LB still makes progress in the same pass.</li>
<li>A genuinely broken LB (<code>delete_pending</code>, unknown/nil status) still returns an error, so real failures still surface through conditions and events instead of retrying forever.</li>
<li>At most one write per LB per pass. If a machine needs more than one pool on the same LB, the rest are picked up on the next pass rather than attempted against a now-locked LB.</li>
<li>Requeue interval dropped from a fixed 1 minute to 15 seconds.</li>
</ul>
<p><code>CreateVPCLoadBalancerPoolMember</code> now returns <code>(pending bool, err error)</code> instead of <code>(*LoadBalancerPoolMember, error)</code> — the caller only ever needs to know whether more work remains, not the specific member object.</p>
<h2>Testing</h2>
<p>Full unit test coverage for the new behavior: busy-LB skip, parallel progress across two LBs, one-write-per-LB-per-pass deferral, multi-pass completion, nil-status handling, and the busy-vs-broken distinction (including a dedicated <code>delete_pending</code> case that must still error).</p>
<p>Also validated against real IBM Cloud infrastructure — 3 control-plane machines, 2 LBs, 2 pools each, comparing the stock <code>v0.12.2</code> controller against this fix:</p>

  | Before | After
-- | -- | --
Time to register all 3 machines | 15m 31s | 7m 56s
Mean gap between writes | 86s | 45s
Max gap | 120s | 84s

<p>~1.95× faster on identical infrastructure and manifest. Both runs were confirmed to have matching load balancer configuration before timing started and were checked against live IBM Cloud state before teardown, so this is a clean apples-to-apples comparison, not a partial or estimated one.</p>

<h2>Why not something else</h2>
<p>A few other approaches were considered and ruled out:</p>
<ul>
<li><strong>Batching registrations across machines</strong> — not viable in the CAPI model; machine count can scale up/down, and the first control plane needs to be behind the LB immediately for HA.</li>
<li><strong>Replacing the whole pool member list in one call</strong> — the LB lock is appliance-wide, not per-pool, so this doesn't reduce the number of lock cycles, and it risks silently dropping members added concurrently by another machine.</li>
<li><strong>Every other major cloud (AWS, Azure, GCP) avoids this entirely</strong> — none of them lock the LB appliance itself when adding a backend. IBM VPC's load balancer is the outlier here, so there's no existing pattern to borrow; the fix has to live in this provider.</li>
</ul>
<p>Related: kubernetes-sigs/cluster-api-provider-ibmcloud#2839</p></body></html>

### DIFF
--- a/cloud/scope/powervs_machine.go
+++ b/cloud/scope/powervs_machine.go
@@ -955,16 +955,35 @@ func (m *PowerVSMachineScope) GetMachineInternalIP() string {
 	return ""
 }
 
-// CreateVPCLoadBalancerPoolMember creates a member in load balancer pool.
-func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (*vpcv1.LoadBalancerPoolMember, error) { //nolint:gocyclo
+// CreateVPCLoadBalancerPoolMember registers the current machine's IP in every eligible load balancer
+// pool. It is designed for parallel progress across multiple load balancers:
+//
+//   - A transiently-busy load balancer (update_pending, create_pending) is skipped — not an error —
+//     so the other load balancer can still make progress in the same reconcile pass. This avoids the
+//     error-backoff retry storm that occurred when "LB not active" was returned as an error.
+//   - A non-transient non-active state (e.g. delete_pending, or any unknown state) is returned as
+//     an error so that a genuinely broken load balancer surfaces through conditions/events rather
+//     than silently retrying forever.
+//   - At most one CreateLoadBalancerPoolMember (POST) is issued per load balancer per pass. After a
+//     write the load balancer goes update_pending, so subsequent pools on the same load balancer are
+//     checked for membership only; if any are still unregistered the pending flag is set so the
+//     controller requeues.
+//
+// Returns (pendingUpdate, err):
+//   - pendingUpdate: true when any registration is still incomplete (load balancer busy, or more
+//     pools remain after this pass's one write). The controller requeues at
+//     loadBalancerSettleRequeueInterval while this is true.
+//   - err: only set for real API failures or a non-transient load balancer state; a
+//     transiently-busy load balancer is never an error.
+func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (bool, error) { //nolint:gocyclo
 	log := ctrl.LoggerFrom(ctx)
+
 	loadBalancers := make([]infrav1.VPCLoadBalancerSpec, 0)
 	if len(m.IBMPowerVSCluster.Spec.LoadBalancers) == 0 {
-		loadBalancer := infrav1.VPCLoadBalancerSpec{
+		loadBalancers = append(loadBalancers, infrav1.VPCLoadBalancerSpec{
 			Name:   fmt.Sprintf("%s-loadbalancer", m.IBMPowerVSCluster.Name),
 			Public: ptr.To(true),
-		}
-		loadBalancers = append(loadBalancers, loadBalancer)
+		})
 	}
 	for index, loadBalancer := range m.IBMPowerVSCluster.Spec.LoadBalancers {
 		if loadBalancer.Name == "" {
@@ -973,51 +992,70 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Contex
 		loadBalancers = append(loadBalancers, loadBalancer)
 	}
 
-	for _, lb := range loadBalancers {
-		var lbID *string
+	pendingUpdate := false
+
+	for _, loadBalancer := range loadBalancers {
 		if m.IBMPowerVSCluster.Status.LoadBalancers == nil {
-			return nil, fmt.Errorf("failed to find VPC load balancer ID")
+			return false, fmt.Errorf("failed to find VPC load balancer ID")
 		}
-		if val, ok := m.IBMPowerVSCluster.Status.LoadBalancers[lb.Name]; ok {
-			lbID = val.ID
-		} else {
-			return nil, fmt.Errorf("failed to find VPC load balancer ID")
+		loadBalancerStatus, ok := m.IBMPowerVSCluster.Status.LoadBalancers[loadBalancer.Name]
+		if !ok {
+			return false, fmt.Errorf("failed to find VPC load balancer ID")
 		}
-		loadBalancer, _, err := m.IBMVPCClient.GetLoadBalancer(&vpcv1.GetLoadBalancerOptions{
-			ID: lbID,
-		})
+		loadBalancerID := loadBalancerStatus.ID
+
+		loadBalancerDetails, _, err := m.IBMVPCClient.GetLoadBalancer(&vpcv1.GetLoadBalancerOptions{ID: loadBalancerID})
 		if err != nil {
-			return nil, fmt.Errorf("failed to find VPC load balancer details: %w", err)
+			return false, fmt.Errorf("failed to find VPC load balancer details: %w", err)
 		}
-		if *loadBalancer.ProvisioningStatus != string(infrav1.VPCLoadBalancerStateActive) {
-			return nil, fmt.Errorf("VPC load balancer is not in active state, current state %s", *loadBalancer.ProvisioningStatus)
+
+		// Only transient states are safe to skip and retry. A non-transient non-active state
+		// (e.g. delete_pending, or any unknown state) means the load balancer is genuinely broken
+		// and should surface as an error so conditions/events reflect it.
+		// Note: the provisioning-status check must come before the pool-count check because a
+		// create_pending LB may legitimately have zero pools until creation completes.
+		if loadBalancerDetails.ProvisioningStatus == nil {
+			return false, fmt.Errorf("VPC load balancer %s is in non-recoverable state %q", loadBalancer.Name, "<nil>")
 		}
-		if len(loadBalancer.Pools) == 0 {
-			return nil, fmt.Errorf("no pools exist for the VPC load balancer %s", lb.Name)
+
+		loadBalancerState := infrav1.VPCLoadBalancerState(*loadBalancerDetails.ProvisioningStatus)
+
+		switch loadBalancerState {
+		case infrav1.VPCLoadBalancerStateActive:
+			// Load balancer is active and ready; proceed normally.
+		case infrav1.VPCLoadBalancerStateUpdatePending, infrav1.VPCLoadBalancerStateCreatePending:
+			log.V(3).Info("VPC load balancer is transiently busy, skipping this pass",
+				"loadBalancerName", loadBalancer.Name, "loadBalancerState", loadBalancerState)
+			pendingUpdate = true
+			continue
+		default:
+			return false, fmt.Errorf("VPC load balancer %s is in non-recoverable state %q", loadBalancer.Name, loadBalancerState)
+		}
+
+		if len(loadBalancerDetails.Pools) == 0 {
+			return false, fmt.Errorf("no pools exist for the VPC load balancer %s", loadBalancer.Name)
 		}
 
 		internalIP := m.GetMachineInternalIP()
 
-		// lbAdditionalListeners is a mapping of additionalListener's port-protocol to the additionalListener as defined in the specification
-		// It will be used later to get the default pool associated with the listener
+		// lbAdditionalListeners: port-protocol key → spec entry (used to resolve target ports).
 		lbAdditionalListeners := map[string]infrav1.AdditionalListenerSpec{}
-		for _, additionalListener := range lb.AdditionalListeners {
+		for _, additionalListener := range loadBalancer.AdditionalListeners {
 			if additionalListener.Protocol == nil {
 				additionalListener.Protocol = &infrav1.VPCLoadBalancerListenerProtocolTCP
 			}
 			lbAdditionalListeners[fmt.Sprintf("%d-%s", additionalListener.Port, *additionalListener.Protocol)] = additionalListener
 		}
 
-		// loadBalancerListeners is a mapping of the loadBalancer listener's defaultPoolName to the additionalListener
-		// as the default pool name might be empty in spec and should be fetched from the cloud's listener
+		// loadBalancerListeners: pool name → listener spec (populated from the cloud's listener list).
 		loadBalancerListeners := map[string]infrav1.AdditionalListenerSpec{}
-		for _, listener := range loadBalancer.Listeners {
-			listenerOptions := &vpcv1.GetLoadBalancerListenerOptions{}
-			listenerOptions.SetLoadBalancerID(*loadBalancer.ID)
-			listenerOptions.SetID(*listener.ID)
-			loadBalancerListener, _, err := m.IBMVPCClient.GetLoadBalancerListener(listenerOptions)
+		for _, listenerRef := range loadBalancerDetails.Listeners {
+			listenerOpts := &vpcv1.GetLoadBalancerListenerOptions{}
+			listenerOpts.SetLoadBalancerID(*loadBalancerDetails.ID)
+			listenerOpts.SetID(*listenerRef.ID)
+			loadBalancerListener, _, err := m.IBMVPCClient.GetLoadBalancerListener(listenerOpts)
 			if err != nil {
-				return nil, fmt.Errorf("failed to list %s load balancer listener: %w", *listener.ID, err)
+				return false, fmt.Errorf("failed to get load balancer listener %s: %w", *listenerRef.ID, err)
 			}
 			if additionalListener, ok := lbAdditionalListeners[fmt.Sprintf("%d-%s", *loadBalancerListener.Port, *loadBalancerListener.Protocol)]; ok {
 				if loadBalancerListener.DefaultPool != nil {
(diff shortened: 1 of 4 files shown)


controllers/ibmpowervsmachine_controller.go
```diff
@@ -250,14 +250,20 @@ func (r *IBMPowerVSMachineReconciler) reconcileDelete(ctx context.Context, scope
 	return ctrl.Result{}, nil
 }
 
+// lbSettleRequeue is the poll interval while waiting for a VPC load balancer pool update to leave
+// the update_pending state. IBM VPC load balancers typically settle in 10–30 s; 15 s avoids
+// hammering the API while still reacting promptly. The IBM VPC team confirmed that polling at this
+// rate does not extend the update_pending window.
+const lbSettleRequeue = 15 * time.Second
+
```
Karthik-K-N: Lets move this to top of the file.
is there any better name?

controllers/ibmpowervsmachine_controller.go
```diff
@@ -250,14 +250,20 @@ func (r *IBMPowerVSMachineReconciler) reconcileDelete(ctx context.Context, scope
...
 func (r *IBMPowerVSMachineReconciler) handleLoadBalancerPoolMemberConfiguration(ctx context.Context, machineScope *scope.PowerVSMachineScope) (ctrl.Result, error) {
-	poolMember, err := machineScope.CreateVPCLoadBalancerPoolMember(ctx)
+	pending, err := machineScope.CreateVPCLoadBalancerPoolMember(ctx)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to create VPC load balancer pool member: %w", err)
+		return ctrl.Result{}, fmt.Errorf("failed to configure VPC load balancer pool member: %w", err)
 	}
-	if poolMember != nil && *poolMember.ProvisioningStatus != string(infrav1.VPCLoadBalancerStateActive) {
-		return ctrl.Result{RequeueAfter: 1 * time.Minute}, nil
+	if pending {
```
Karthik-K-N: should we log that its pending?

cloud/scope/powervs_machine_test.go
```diff
@@ -2133,16 +2162,298 @@ func TestCreateVPCLoadBalancerPoolMemberPowerVSMachine(t *testing.T) {
...
+				},
+			},
+			IBMPowerVSCluster: &infrav1.IBMPowerVSCluster{
+				Spec:   infrav1.IBMPowerVSClusterSpec{LoadBalancers: lbs},
+				Status: infrav1.IBMPowerVSClusterStatus{LoadBalancers: statuses},
+			},
+		}
+	}
+
+	t.Run("Busy LB is skipped not an error: returns (nil, true, nil), no POST", func(t *testing.T) {
```
Karthik-K-N: I think we should better name test cases

cloud/scope/powervs_machine.go
```diff
@@ -955,167 +955,203 @@ func (m *PowerVSMachineScope) GetMachineInternalIP() string {
...
-			loadBalancer.Name = fmt.Sprintf("%s-loadbalancer-%d", m.IBMPowerVSCluster.Name, index)
+	for index, lb := range m.IBMPowerVSCluster.Spec.LoadBalancers {
+		if lb.Name == "" {
+			lb.Name = fmt.Sprintf("%s-loadbalancer-%d", m.IBMPowerVSCluster.Name, index)
 		}
-		loadBalancers = append(loadBalancers, loadBalancer)
+		loadBalancers = append(loadBalancers, lb)
 	}
 
+	pendingWork := false
```
Karthik-K-N: can it be pendingUpdate?

cloud/scope/powervs_machine.go
```diff
@@ -955,167 +955,203 @@ func (m *PowerVSMachineScope) GetMachineInternalIP() string {
...
+		// should surface as an error so conditions/events reflect it.
+		if lbDetails.ProvisioningStatus == nil || *lbDetails.ProvisioningStatus != string(infrav1.VPCLoadBalancerStateActive) {
+			lbState := "<nil>"
+			if lbDetails.ProvisioningStatus != nil {
+				lbState = *lbDetails.ProvisioningStatus
+			}
+			switch infrav1.VPCLoadBalancerState(lbState) {
+			case infrav1.VPCLoadBalancerStateUpdatePending, infrav1.VPCLoadBalancerStateCreatePending:
+				log.V(3).Info("VPC load balancer is transiently busy, skipping this pass",
+					"loadBalancerName", lb.Name, "state", lbState)
```
Karthik-K-N:
```suggestion
				log.V(3).Info("VPC load balancer is transiently busy, skipping this pass",
					"loadBalancerName", lb.Name, "loadBalancerState", lbState)
```

cloud/scope/powervs_machine.go
```diff
@@ -955,167 +955,203 @@ func (m *PowerVSMachineScope) GetMachineInternalIP() string {
...
+				continue
+			default:
+				return false, fmt.Errorf("VPC load balancer %s is in non-recoverable state %q", lb.Name, lbState)
+			}
 		}
-		if len(loadBalancer.Pools) == 0 {
-			return nil, fmt.Errorf("no pools exist for the VPC load balancer %s", lb.Name)
+
+		if len(lbDetails.Pools) == 0 {
+			return false, fmt.Errorf("no pools exist for the VPC load balancer %s", lb.Name)
```
Karthik-K-N: Should this go just after the err check, If there is no pool, there is no need to check the lb state right?
may be we can also think of returning without error, if no pool nothing to configure.

cloud/scope/powervs_machine.go
```diff
@@ -955,167 +955,203 @@ func (m *PowerVSMachineScope) GetMachineInternalIP() string {
...
 		internalIP := m.GetMachineInternalIP()
 
-		// lbAdditionalListeners is a mapping of additionalListener's port-protocol to the additionalListener as defined in the specification
-		// It will be used later to get the default pool associated with the listener
+		// lbAdditionalListeners: port-protocol key → spec entry (used to resolve target ports).
 		lbAdditionalListeners := map[string]infrav1.AdditionalListenerSpec{}
-		for _, additionalListener := range lb.AdditionalListeners {
-			if additionalListener.Protocol == nil {
-				additionalListener.Protocol = &infrav1.VPCLoadBalancerListenerProtocolTCP
+		for _, al := range lb.AdditionalListeners {
```
Karthik-K-N: any reason for using shortname? I would recommend keeping meaning full shortform or full name.

cloud/scope/powervs_machine.go
```diff
@@ -955,167 +955,203 @@ func (m *PowerVSMachineScope) GetMachineInternalIP() string {
...
+					loadBalancerListeners[*lbListener.DefaultPool.Name] = al
 				}
 				// loadBalancerListeners map is created only with the listeners provided in the spec,
 				// and targetPort is populated only if there is an entry in the map.
 				// Inorder for the default pool 6443 to be added to all control plane machines, creating an entry in the map for the same.
-			} else if loadBalancerListener.Port != nil && *loadBalancerListener.Port == int64(6443) {
-				protocol := infrav1.VPCLoadBalancerListenerProtocol(*loadBalancerListener.Protocol)
-				listener := infrav1.AdditionalListenerSpec{
-					Port:     *loadBalancerListener.Port,
+			} else if lbListener.Port != nil && *lbListener.Port == int64(6443) {
```
Karthik-K-N: I think given its a performance improvement PR, focusing only on those changes would be good, I think only name refactoring can be kept in separate PR

cloud/scope/powervs_machine.go
```diff
@@ -955,167 +955,203 @@ func (m *PowerVSMachineScope) GetMachineInternalIP() string {
...
-			var alreadyRegistered bool
 
-			if loadBalancerListener, ok := loadBalancerListeners[*pool.Name]; ok {
-				targetPort = loadBalancerListener.Port
-				log.V(3).Info("Checking if machine label matches with the label selector in listener", "machineLabel", m.IBMPowerVSMachine.Labels, "labelSelector", loadBalancerListener.Selector)
-				selector, err := metav1.LabelSelectorAsSelector(&loadBalancerListener.Selector)
+		// wroteToThisLB tracks whether we've already issued a POST to this LB in the current pass.
+		// After one write the LB goes update_pending, so further writes would fail the active-check.
+		// Instead we continue scanning remaining pools for membership to set pendingWork correctly.
+		wroteToThisLB := false
```
Karthik-K-N: updatedThisLB? or something

Karthik-K-N (commented):
Thank you for working on this, Few small suggestions otherwise it LGTM

Karthik-K-N:
@Amulyam24 should we cherry-pick any PRs to this branch to pass this tests?

Changes LGTM, @aman4433 see if you can get the similar changes created against main before we cut the release.

/assign @Amulyam24 for final review.

cloud/scope/powervs_machine.go
```diff
@@ -973,51 +992,68 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Contex
...
-		for _, listener := range loadBalancer.Listeners {
-			listenerOptions := &vpcv1.GetLoadBalancerListenerOptions{}
-			listenerOptions.SetLoadBalancerID(*loadBalancer.ID)
-			listenerOptions.SetID(*listener.ID)
-			loadBalancerListener, _, err := m.IBMVPCClient.GetLoadBalancerListener(listenerOptions)
+		for _, listenerRef := range loadBalancerDetails.Listeners {
+			listenerOpts := &vpcv1.GetLoadBalancerListenerOptions{}
+			listenerOpts.SetLoadBalancerID(*loadBalancerDetails.ID)
+			listenerOpts.SetID(*listenerRef.ID)
+			loadBalancerListener, _, err := m.IBMVPCClient.GetLoadBalancerListener(listenerOpts)
```
prb112, line 1056: changes the variable names... right? nothing logic here

cloud/scope/powervs_machine.go
```diff
@@ -973,51 +992,68 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Contex
...
 			if err != nil {
-				return nil, fmt.Errorf("failed to list %s load balancer listener: %w", *listener.ID, err)
+				return false, fmt.Errorf("failed to get load balancer listener %s: %w", *listenerRef.ID, err)
```
prb112, line 1058: `false` being returned here is the key right?

cloud/scope/powervs_machine.go
```diff
@@ -1061,61 +1095,69 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Contex
...
-			if *loadBalancer.ProvisioningStatus != string(infrav1.VPCLoadBalancerStateActive) {
-				log.V(3).Info("Unable to update pool for VPC load balancer as it is not in active state", "loadBalancerName", *loadBalancer.Name, "loadBalancerState", *loadBalancer.ProvisioningStatus)
-				return nil, fmt.Errorf("VPC load balancer %s not in active state to update pool member", *loadBalancer.Name)
+			// Machine is not yet in this pool — registration is needed.
+			if updatedThisLB {
+				// We already wrote to this load balancer this pass; it is now update_pending.
+				// Don't attempt another write. Record that work remains and move on.
+				log.V(3).Info("Already updated load balancer this pass, deferring remaining pool registration",
+					"pool", *pool.Name, "loadBalancerName", *loadBalancerDetails.Name)
+				pendingUpdate = true
```
prb112, line 1139: 👍

cloud/scope/powervs_machine.go
```diff
@@ -1061,61 +1095,69 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Contex
...
-				Address: &internalIP,
-			})
-			log.V(3).Info("Creating VPC load balancer pool member", "options", options)
-			loadBalancerPoolMember, _, err := m.IBMVPCClient.CreateLoadBalancerPoolMember(options)
+			opts := &vpcv1.CreateLoadBalancerPoolMemberOptions{}
+			opts.SetPort(targetPort)
+			opts.SetLoadBalancerID(*loadBalancerDetails.ID)
+			opts.SetPoolID(*pool.ID)
+			opts.SetTarget(&vpcv1.LoadBalancerPoolMemberTargetPrototype{Address: &internalIP})
+			log.V(3).Info("Creating VPC load balancer pool member", "pool", *pool.Name, "IP", internalIP, "port", targetPort)
```
prb112, line 1148: options became opts... right?

prb112 (commented):
LGTM to me...

cloud/scope/powervs_machine.go
aman4433: Yep, just renames.

cloud/scope/powervs_machine.go
aman4433: Yep, just a rename, no behavior change.

cloud/scope/powervs_machine.go
aman4433: Right, Since the signature moved from `(*vpcv1.LoadBalancerPoolMember, error) `to `(bool, error)`.

controllers/ibmpowervsmachine_controller.go
```diff
@@ -55,6 +55,12 @@ import (
 	capibmrecord "sigs.k8s.io/cluster-api-provider-ibmcloud/pkg/record"
 )
 
+// loadBalancerSettleRequeueInterval is the poll interval while waiting for a VPC load balancer pool
+// update to leave the update_pending state. IBM VPC load balancers typically settle in 10–30 s;
+// 15 s avoids hammering the API while still reacting promptly. The IBM VPC team confirmed that
+// polling at this rate does not extend the update_pending window.
+const loadBalancerSettleRequeueInterval = 15 * time.Second
```
Amulyam24:
```suggestion
// 15 s avoids hammering the API while still reacting promptly.
const loadBalancerSettleRequeueInterval = 15 * time.Second
```

cloud/scope/powervs_machine.go
```diff
@@ -973,51 +992,68 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Contex
...
+			}
+			switch infrav1.VPCLoadBalancerState(loadBalancerState) {
+			case infrav1.VPCLoadBalancerStateUpdatePending, infrav1.VPCLoadBalancerStateCreatePending:
+				log.V(3).Info("VPC load balancer is transiently busy, skipping this pass",
+					"loadBalancerName", loadBalancer.Name, "loadBalancerState", loadBalancerState)
+				pendingUpdate = true
+				continue
+			default:
+				return false, fmt.Errorf("VPC load balancer %s is in non-recoverable state %q", loadBalancer.Name, loadBalancerState)
+			}
```
Amulyam24:
```suggestion
		if loadBalancerDetails.ProvisioningStatus == nil {
    return false, fmt.Errorf("VPC load balancer %s state is nil", loadBalancer.Name)
}

state := infrav1.VPCLoadBalancerState(*loadBalancerDetails.ProvisioningStatus)

switch state {
case infrav1.VPCLoadBalancerStateActive:
    // Load balancer is active and ready; proceed normally.

case infrav1.VPCLoadBalancerStateUpdatePending, infrav1.VPCLoadBalancerStateCreatePending:
    log.V(3).Info("VPC load balancer is transiently busy, skipping this pass",
        "loadBalancerName", loadBalancer.Name, "loadBalancerState", state)
    pendingUpdate = true
    continue

default:
    return false, fmt.Errorf("VPC load balancer %s is in non-recoverable state %q", loadBalancer.Name, state)
}
```

How about simplifying the check as?

Hope you have taken the known LB states into consideration.

cloud/scope/powervs_machine.go
```diff
@@ -1028,31 +1064,29 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Contex
 				// Inorder for the default pool 6443 to be added to all control plane machines, creating an entry in the map for the same.
 			} else if loadBalancerListener.Port != nil && *loadBalancerListener.Port == int64(6443) {
 				protocol := infrav1.VPCLoadBalancerListenerProtocol(*loadBalancerListener.Protocol)
-				listener := infrav1.AdditionalListenerSpec{
+				syntheticListener := infrav1.AdditionalListenerSpec{
```
Amulyam24:
```suggestion
				listener := infrav1.AdditionalListenerSpec{
```

cloud/scope/powervs_machine.go
```diff
@@ -1061,61 +1095,69 @@ func (m *PowerVSMachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Contex
...
 				}
 			}
 
-			for _, member := range listLoadBalancerPoolMembers.Members {
-				if target, ok := member.Target.(*vpcv1.LoadBalancerPoolMemberTarget); ok {
+			// Check existing membership.
+			listOpts := &vpcv1.ListLoadBalancerPoolMembersOptions{}
+			listOpts.SetLoadBalancerID(*loadBalancerDetails.ID)
+			listOpts.SetPoolID(*pool.ID)
+			existingMembers, _, err := m.IBMVPCClient.ListLoadBalancerPoolMembers(listOpts)
```
Amulyam24:
```suggestion
			existingPoolMembers, _, err := m.IBMVPCClient.ListLoadBalancerPoolMembers(listOpts)
```

Amulyam24 (commented):
Overall LGTM, minor suggestions added.

Thank you for working on this @aman4433!

Amulyam24:
@aman4433, the verify job would need to be fixed with updating the links.

cloud/scope/powervs_machine.go
aman4433: Thanks for the suggestion,  I’ll apply the restructure, but keep "non-recoverable state" for nil since an existing test checks it.

I also checked the SDK: `failed` should error, while `maintenance_pending` and `migrate_pending` should requeue like `update_pending`. `delete_pending` should remain an error.

Since LB state handling is inconsistent across three places `(powervs_machine.go, powervs_cluster.go, vpc_cluster.go)`, aligning them feels better suited to a separate PR.

Amulyam24:
@aman4433, please squash the commits

PR #2888, opened 2026-08-07
powervs: change default systemType from s922 to s1022

aman4433:
**What this PR does / why we need it**

The periodic PowerVS e2e job has failed on six consecutive runs
(Jul 31 – Aug 5). The Boskos workspaces it leases no longer support
the `s922` system type: `lon06` offers only `s1022`, and `eu-de-1`
offers `e1050`, `e1080`, `e1180`, `s1022` and `s1122`. Every machine
creation fails with `InvalidMachineConfiguration`, no control plane
comes up, and both e2e specs time out.

`s1022` is the only system type available in both workspaces, so it
is the only viable default while Boskos can allocate either one.

Changes:
- `defaultSystemType` in the mutating webhook is now `s1022`. This is
  the user-facing half of the fix — changing only the templates would
  still leave anyone omitting `systemType` defaulting to `s922`.
- All cluster templates and the e2e template default to `s1022`.
- `s922` is removed from the v1beta2 kubebuilder enum and the CRDs are
  regenerated.
- v1beta3 doc comments updated. No enum is added there — the controller
  already validates `systemType` dynamically against the PowerVS API,
  which is why v1beta3 carries only pattern validation.

**Note on the enum change**

Removing `s922` from the v1beta2 enum is a breaking change on the write
path: existing objects with `systemType: s922` will be rejected on any
write once the new CRD is applied. It is taken here because main is
pre-release and v1beta3 is the going-forward API, not because it is
non-breaking — v1beta2 is still served. Release branches are handled
separately.

**Follow-ups**

- Enum parity for `s1122` and `e1180`, which `eu-de-1` supports but
  neither API version lists.
- Whether `e980` (the other Power9, absent from both CI workspaces) is
  globally withdrawn or just not stocked in these two zones.
- `lon06` now has a single supported system type, so it has no fallback
  if `s1022` is ever withdrawn there — worth revisiting the Boskos pool.

Fixes #<issue>

/area provider/ibmcloud

### DIFF
--- a/api/powervs/v1beta2/ibmpowervsmachine_types.go
+++ b/api/powervs/v1beta2/ibmpowervsmachine_types.go
@@ -78,11 +78,11 @@ type IBMPowerVSMachineSpec struct {
 
 	// systemType is the System type used to host the instance.
 	// systemType determines the number of cores and memory that is available.
-	// Few of the supported SystemTypes are s922,e980,s1022,e1050,e1080.
+	// Few of the supported SystemTypes are e980,s1022,e1050,e1080.
 	// When omitted, this means that the user has no opinion and the platform is left to choose a
-	// reasonable default, which is subject to change over time. The current default is s922 which is generally available.
+	// reasonable default, which is subject to change over time. The current default is s1022 which is generally available.
 	// + This is not an enum because we expect other values to be added later which should be supported implicitly.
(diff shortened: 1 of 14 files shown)


Karthik-K-N (commented):
/lgtm

Lets wait for @Amulyam24 refactor PR to merge so the CI tests will pass.
Thank you.

Amulyam24 (approved):
/lgtm
/approve

PR #2909, opened 2026-08-19
fix(powervs): skip busy load balancers instead of erroring, cut requeue interval to 15s

aman4433:
## What this fixes

Main-branch equivalent of #2883 (targeting release-0.13).

During PowerVS cluster provisioning, control-plane machines register into VPC load
balancer pools. Each registration briefly locks the load balancer while it applies the
change. The old code treated a locked LB as an error and gave up on the whole reconcile
pass — including any other LB that was free the whole time. It also waited a flat 60
seconds before trying again, no matter how quickly the LB actually settled.

Result: with 2 LBs × 2 pools per machine, registering 3 control-plane machines took
15–25 minutes in production, occasionally blowing past the installer's bootstrap timeout.

## What changed

- A load balancer in a transient state (`update_pending`, `create_pending`) is now
  skipped, not errored — the other LB still makes progress in the same pass.
- A genuinely broken LB (`delete_pending`, unknown/nil status) still returns an error, so
  real failures surface through conditions and events instead of retrying forever.
- At most one write per LB per pass. If a machine needs more than one pool on the same LB,
  the rest are picked up on the next pass rather than attempted against a now-locked LB.
- Requeue interval dropped from a fixed 1 minute to 15 seconds.

`CreateVPCLoadBalancerPoolMember` now returns `(pendingUpdate bool, err error)` instead of
`(*LoadBalancerPoolMember, error)` — the caller only ever needs to know whether more work
remains, not the specific member object.

## Differences from #2883

Same design, adapted to main's structure:

- Paths: `pkg/cloud/scope/powervs/` and `internal/controllers/powervs/`
- v1beta3 types: `LoadBalancerState*` instead of `VPCLoadBalancerState*`,
  `AdditionalListener` instead of `AdditionalListenerSpec`
- `IBMPowerVSCluster.Status.LoadBalancers` is a slice on main rather than a map, so the
  load balancer ID lookup keeps main's existing loop rather than a map index

All review feedback from #2883 is incorporated here.

## Testing

Full unit test coverage for the new behavior: busy-LB skip, parallel progress across two
LBs, one-write-per-LB-per-pass deferral, multi-pass completion, nil-status handling, the
busy-vs-broken distinction (including a `delete_pending` case that must still error), and
a `create_pending`-with-zero-pools case confirming the provisioning-status check must run
before the pool-count check.

The equivalent change was validated on release-0.13 against real IBM Cloud infrastructure —
3 control-plane machines, 2 LBs, 2 pools each, comparing the stock v0.12.2 controller
against the fix:

| | Before | After |
|---|---|---|
| Time to register all 3 machines | 15m 31s | 7m 56s |
| Mean gap between writes | 86s | 45s |
| Max gap | 120s | 84s |

~1.95× faster on identical infrastructure and manifest. Both runs had matching load
balancer configuration confirmed before timing started and were checked against live IBM
Cloud state before teardown. The logic here is equivalent, so the same improvement applies.

Related: #2839

### DIFF
--- a/pkg/cloud/scope/powervs/powervs_machine.go
+++ b/pkg/cloud/scope/powervs/powervs_machine.go
@@ -481,19 +481,31 @@ func (s *MachineScope) DeleteMachineIgnition(ctx context.Context) error {
 	return nil
 }
 
-// CreateVPCLoadBalancerPoolMember creates a member in load balancer pool.
-func (s *MachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (*vpcv1.LoadBalancerPoolMember, error) { //nolint:gocyclo
+// CreateVPCLoadBalancerPoolMember registers the current machine's IP in every eligible load balancer
+// pool. It is designed for parallel progress across multiple load balancers:
+//
+//   - A transiently-busy load balancer (update_pending, create_pending) is skipped — not an error —
+//     so the other load balancer can still make progress in the same reconcile pass.
+//   - A non-transient non-active state (e.g. delete_pending, or any unknown state) is returned as
+//     an error so that a genuinely broken load balancer surfaces through conditions/events.
+//   - At most one CreateLoadBalancerPoolMember (POST) is issued per load balancer per pass. After a
+//     write the load balancer goes update_pending, so subsequent pools on the same load balancer are
+//     checked for membership only; if any are still unregistered the pending flag is set.
+//
+// Returns (pendingUpdate, err):
+//   - pendingUpdate: true when any registration is still incomplete.
+//   - err: only set for real API failures or a non-transient load balancer state.
+func (s *MachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (bool, error) { //nolint:gocyclo
 	log := ctrl.LoggerFrom(ctx)
 	loadBalancers := make([]infrav1.LoadBalancerSource, 0)
 	if len(s.IBMPowerVSCluster.Spec.LoadBalancers) == 0 {
-		loadBalancer := infrav1.LoadBalancerSource{
+		loadBalancers = append(loadBalancers, infrav1.LoadBalancerSource{
 			Type: infrav1.SourceTypeProvision,
 			Provision: infrav1.LoadBalancerProvision{
 				Name: ResourceName(s.IBMPowerVSCluster.Name, ResourceTypeLBPublic, ""),
 				Type: infrav1.LoadBalancerTypePublic,
 			},
-		}
-		loadBalancers = append(loadBalancers, loadBalancer)
+		})
 	}
 	for index, loadBalancer := range s.IBMPowerVSCluster.Spec.LoadBalancers {
 		if loadBalancer.Type == infrav1.SourceTypeProvision && loadBalancer.Provision.Name == "" {
@@ -510,6 +522,8 @@ func (s *MachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (*vp
 		loadBalancers = append(loadBalancers, loadBalancer)
 	}
 
+	pendingUpdate := false
+
 	for _, lb := range loadBalancers {
 		var lbName string
 		switch lb.Type {
@@ -519,13 +533,13 @@ func (s *MachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (*vp
 			lbName = lb.Provision.Name
 		}
 		if lbName == "" {
-			return nil, fmt.Errorf("failed to determine VPC load balancer name")
+			return false, fmt.Errorf("failed to determine VPC load balancer name")
 		}
 
-		var lbID string
 		if len(s.IBMPowerVSCluster.Status.LoadBalancers) == 0 {
-			return nil, fmt.Errorf("failed to find VPC load balancer ID")
+			return false, fmt.Errorf("failed to find VPC load balancer ID")
 		}
+		var lbID string
 		found := false
 		for _, val := range s.IBMPowerVSCluster.Status.LoadBalancers {
 			if val.Name == lbName {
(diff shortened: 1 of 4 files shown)


pkg/cloud/scope/powervs/powervs_machine.go
```diff
@@ -503,24 +517,44 @@ func (s *MachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (*vp
...
-		if loadBalancer.ProvisioningStatus == nil || *loadBalancer.ProvisioningStatus != string(infrav1.LoadBalancerStateActive) {
-			return nil, fmt.Errorf("VPC load balancer is not in active state, current state %s", ptr.Deref(loadBalancer.ProvisioningStatus, ""))
+
+		// Only transient states are safe to skip and retry. A non-transient non-active state
+		// (e.g. delete_pending, or any unknown state) means the load balancer is genuinely broken
+		// and should surface as an error so conditions/events reflect it.
+		// Note: the provisioning-status check must come before the pool-count check because a
+		// create_pending LB may legitimately have zero pools until creation completes.
+		if loadBalancerDetails.ProvisioningStatus == nil {
+			return false, fmt.Errorf("VPC load balancer %s is in non-recoverable state %q", lbName, "<nil>")
```
Karthik-K-N: what is this _`"<nil>"`_

pkg/cloud/scope/powervs/powervs_machine.go
```diff
@@ -503,24 +517,44 @@ func (s *MachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (*vp
...
+			return false, fmt.Errorf("failed to find VPC load balancer details: %w", err)
 		}
-		if loadBalancer.ProvisioningStatus == nil || *loadBalancer.ProvisioningStatus != string(infrav1.LoadBalancerStateActive) {
-			return nil, fmt.Errorf("VPC load balancer is not in active state, current state %s", ptr.Deref(loadBalancer.ProvisioningStatus, ""))
+
+		// Only transient states are safe to skip and retry. A non-transient non-active state
+		// (e.g. delete_pending, or any unknown state) means the load balancer is genuinely broken
+		// and should surface as an error so conditions/events reflect it.
+		// Note: the provisioning-status check must come before the pool-count check because a
+		// create_pending LB may legitimately have zero pools until creation completes.
```
Karthik-K-N: I think either we should remove this Note or reword it, Lets keep the current state of the code in comments.

pkg/cloud/scope/powervs/powervs_machine.go
```diff
@@ -578,61 +610,69 @@ func (s *MachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (*vp
...
-				log.V(3).Info("Unable to update pool for VPC load balancer as it is not in active state", "loadBalancerName", *loadBalancer.Name, "loadBalancerState", *loadBalancer.ProvisioningStatus)
-				return nil, fmt.Errorf("VPC load balancer %s not in active state to update pool member", *loadBalancer.Name)
+			// Machine is not yet in this pool — registration is needed.
+			if updatedThisLB {
+				// We already wrote to this load balancer this pass; it is now update_pending.
+				// Don't attempt another write. Record that work remains and move on.
+				log.V(3).Info("Already updated load balancer this pass, deferring remaining pool registration",
+					"pool", *pool.Name, "loadBalancerName", *loadBalancerDetails.Name)
+				pendingUpdate = true
+				continue
```
Karthik-K-N: should we continue or break becuase if the pool of a LB is updated we can't update any other pool of the same LB right? we should move onto the next LB, Am I right?

pkg/cloud/scope/powervs/powervs_machine.go
aman4433: It’s just a placeholder so the nil case matches the default message shape. Agreed, it reads oddly. I’ll give nil its own message instead.

Karthik-K-N (commented):
/lgtm

/hold for @Amulyam24 review

Thank you.

pkg/cloud/scope/powervs/powervs_machine.go
```diff
@@ -524,24 +538,44 @@ func (s *MachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (*vp
...
+			return false, fmt.Errorf("VPC load balancer %s has no provisioning status", lbName)
 		}
-		if len(loadBalancer.Pools) == 0 {
-			return nil, fmt.Errorf("no pools exist for the VPC load balancer %s", lbName)
+
+		loadBalancerState := infrav1.LoadBalancerState(*loadBalancerDetails.ProvisioningStatus)
+
+		switch loadBalancerState {
+		case infrav1.LoadBalancerStateActive:
+			// Load balancer is active and ready; proceed normally.
```
Karthik-K-N, line 573: If not logged anywhere we can log with V(3) so we know its active state, but if we are already logging it somwehre we can skip

pkg/cloud/scope/powervs/powervs_machine.go
```diff
@@ -551,14 +585,15 @@ func (s *MachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (*vp
...
 		}
 
+		// loadBalancerListeners: pool name → listener spec (populated from the cloud's listener list).
 		loadBalancerListeners := map[string]infrav1.AdditionalListener{}
-		for _, listener := range loadBalancer.Listeners {
-			listenerOptions := &vpcv1.GetLoadBalancerListenerOptions{}
-			listenerOptions.SetLoadBalancerID(*loadBalancer.ID)
-			listenerOptions.SetID(*listener.ID)
-			loadBalancerListener, _, err := s.IBMVPCClient.GetLoadBalancerListener(listenerOptions)
+		for _, listenerRef := range loadBalancerDetails.Listeners {
```
Karthik-K-N, line 602: good to have but in performance improve PRs lets avoid the renaming so it simply adds the changes, But its good for now but note for next time.

pkg/cloud/scope/powervs/powervs_machine.go
```diff
@@ -599,61 +631,69 @@ func (s *MachineScope) CreateVPCLoadBalancerPoolMember(ctx context.Context) (*vp
...
-			options.SetLoadBalancerID(*loadBalancer.ID)
-			options.SetPoolID(*pool.ID)
-			options.SetTarget(&vpcv1.LoadBalancerPoolMemberTargetPrototype{
-				Address: &internalIP,
-			})
-			log.V(3).Info("Creating VPC load balancer pool member", "options", options)
-			loadBalancerPoolMember, _, err := s.IBMVPCClient.CreateLoadBalancerPoolMember(options)
-			if err != nil {
-				return nil, fmt.Errorf("failed to create VPC load balancer %s pool member %w", *loadBalancer.Name, err)
+			updatedThisLB = true
```
Karthik-K-N, line 701: Just asking, what happens we just break from here instead of setting the variable and checking later and move to next LB, is this udpatedThisLB has any other use cases?

pkg/cloud/scope/powervs/powervs_machine.go
aman4433: Noted, thanks - will keep renames out of performance PRs going forward.

pkg/cloud/scope/powervs/powervs_machine.go
aman4433: The flag is needed because after a successful write, `pendingUpdate` is only set if the new member is still settling. If it’s already active, the flag stays false.

The guard at the top of the loop catches cases where another pool still needs to be processed. For example, with two pools, if pool-1 writes and returns active, pool-2 hits the guard, sets `pendingUpdate=true`, and gets requeued. If we break immediately, we’d return false and pool-2 would never be written.

`updatedThisLB` is only used in three places—init, guard, and write—so it’s purely local to the inner loop.

pkg/cloud/scope/powervs/powervs_machine.go
aman4433: Added the V(3) log for the active case

Amulyam24 (approved):
/lgtm

Karthik-K-N:
/lgtm

/unhold

Thanks

PR #2924, opened 2026-08-25
⚠️ Reorganise the code as per CAPI reference

Karthik-K-N:

This PR refactors directories to keep it simple and have used CAPI code organization as reference.

Fixes #

/area provider/ibmcloud

**Release note**:

```release-note
Reorganise the code as per CAPI reference
```


cmd/main.go
```diff
@@ -279,9 +280,9 @@ func main() {
 	// Setup the context that's going to be used in controllers and for the manager.
 	ctx := ctrl.SetupSignalHandler()
 
-	setupReconcilers(ctx, mgr, serviceEndpoint)
-	setupWebhooks(mgr)
 	setupChecks(mgr)
+	setupReconcilers(ctx, mgr, serviceEndpoint, watchFilterValue, skipCRDMigrationPhases)
+	setupWebhooks(mgr)
```
Amulyam24: Should `setupWebhooks(mgr)` be called before `setupChecks(mgr)`?

Amulyam24 (commented):
Minor comment added, LGTM!

cmd/main.go
Karthik-K-N: It might not matter as everything starts after manager start but its better to keep them in the order so no confusions, I just updated it. Thanks

Amulyam24 (approved):
/lgtm

Thank you @Karthik-K-N!

PR #2929, opened 2026-08-28
Fix golangci-lint exclusions for staticcheck v0.8.0 message format

aman4433:
staticcheck v0.8.0 (pulled in by golangci-lint 2.13.1, see #2927) reports fully-qualified type names in SA1019 messages instead of the file's import alias, so our alias-based exclusion regexes in `.golangci.yml` no longer match — surfacing 86 staticcheck issues. Replaced them with alias-agnostic patterns matching on the deprecation message text.

Also removed the stale `//nolint:unparam` on `reconcileDeleteV2`, which nolintlint flags as unused under 2.13.1. The finding still fires under the currently pinned 2.12.2 (older mvdan.cc/unparam), so it's suppressed via a config rule instead — transitional, and can be dropped once 2.13.1 lands.

`make lint` passes on both 2.12.2 and 2.13.1. Unblocks #2927.


Amulyam24 (commented):
@aman4433, can you update the version as a part of this PR?
We can close the open bump PR.

Amulyam24 (approved):
/lgtm
/approve

Thank you @aman4433!

PR #2942, opened 2026-09-03
🐛 Skip VPC security group rule creation if rule already exists

### LINKED ISSUE #2937: PowerVS: security group rules recreated on every reconcile
## Description

`reconcileVPCSecurityGroupProvision` [powervs_cluster.go#L2167](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/pkg/cloud/scope/powervs/powervs_cluster.go#L2167) finds the existing SG correctly but then calls `createVPCSecurityGroupRules` unconditionally, which always creates. So every reconcile adds the same rules again.

Hit this during an IPI install. The cluster spec had 4 rules for the cluster-wide SG; ended up with 250 (63 tcp/22, 63 tcp/5000, 62 icmp, 62 outbound-all) before hitting the per-group limit. Control plane SG had 248, kube-api-lb 62. Once a group is full the reconcile errors and provisioning stalls.

## Expected Behavior

Rules should be checked against what's already on the SG before creating.

## Area

/area provider/ibmcloud

aman4433:
Every reconcile was calling the IBM Cloud API to create SG rules unconditionally.
Rules kept adding up, hit the 250-per-group limit after ~63 passes, and broke
the cluster. Fix builds a key (direction + protocol + ports + remote), checks
existing rules before creating, and skips if already there.

**Which issue(s) this PR fixes**:
Fixes #2937

/area provider/ibmcloud

**Release note**:
```release-note
Fix duplicate VPC security group rule creation on every reconcile.

### DIFF
--- a/pkg/cloud/scope/powervs/powervs_cluster.go
+++ b/pkg/cloud/scope/powervs/powervs_cluster.go
@@ -74,6 +74,17 @@ var (
 	vpcNetworkConnectionType     = networkConnectionType("vpc")
 )
 
+// sgRuleKey is a comparable identity for a security group rule.
+type sgRuleKey struct {
+	direction string
+	protocol  string
+	portMin   int64
+	portMax   int64
+	cidrBlock string
+	address   string
+	crn       string
+}
+
 // powerEdgeRouter is identifier for PER.
 const (
 	// DEBUGLEVEL indicates the debug level of the logs.
@@ -2164,7 +2175,7 @@ func (s *ClusterScope) reconcileVPCSecurityGroupProvision(ctx context.Context, p
 	}
 
 	// 2. Reconcile Rules
-	ruleIDs, err := s.createVPCSecurityGroupRules(ctx, prov.Rules, *sg.ID)
+	ruleIDs, err := s.createVPCSecurityGroupRules(ctx, prov.Rules, *sg.ID, sg.Rules)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create VPC security group rules for '%s': %w", targetName, err)
 	}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/cloud/scope/powervs/powervs_cluster.go [hunk 1]
```diff
@@ -2204,12 +2204,122 @@ func (s *ClusterScope) createVPCSecurityGroup(ctx context.Context, name string)
...
+	protocol  string
+	portMin   int64
+	portMax   int64
+	cidrBlock string
+	address   string
+	crn       string
+}
+
+// sgRuleIDFrom returns the rule ID from any concrete SG rule type, or "" if unrecognized.
+func sgRuleIDFrom(ruleIntf vpcv1.SecurityGroupRuleIntf) string {
```
Karthik-K-N: Can we rename this?

pkg/cloud/scope/powervs/powervs_cluster.go
```diff
@@ -2204,12 +2204,122 @@ func (s *ClusterScope) createVPCSecurityGroup(ctx context.Context, name string)
...
+	case *vpcv1.SecurityGroupRuleSecurityGroupRuleProtocolIcmp:
+		if r != nil && r.ID != nil {
+			return *r.ID
+		}
+	case *vpcv1.SecurityGroupRuleSecurityGroupRuleProtocolTcpudp:
+		if r != nil && r.ID != nil {
+			return *r.ID
+		}
+	}
+	return ""
```
Karthik-K-N, line 2246: Can't we simplify this, Each switch case is doing the same thing right?

pkg/cloud/scope/powervs/powervs_cluster.go
```diff
@@ -2204,12 +2204,122 @@ func (s *ClusterScope) createVPCSecurityGroup(ctx context.Context, name string)
...
+		key = buildSGRuleKey(r.Direction, r.Protocol, nil, nil, r.Remote)
+	case *vpcv1.SecurityGroupRuleProtocolIcmptcpudp:
+		key = buildSGRuleKey(r.Direction, r.Protocol, nil, nil, r.Remote)
+	case *vpcv1.SecurityGroupRuleProtocolIndividual:
+		key = buildSGRuleKey(r.Direction, r.Protocol, nil, nil, r.Remote)
+	case *vpcv1.SecurityGroupRuleSecurityGroupRuleProtocolIcmp:
+		key = buildSGRuleKey(r.Direction, r.Protocol, nil, nil, r.Remote)
+	case *vpcv1.SecurityGroupRuleSecurityGroupRuleProtocolTcpudp:
+		key = buildSGRuleKey(r.Direction, r.Protocol, r.PortMin, r.PortMax, r.Remote)
+	default:
```
Karthik-K-N, line 2268: I feel even this can be simplified isn't it?

pkg/cloud/scope/powervs/powervs_cluster.go
```diff
@@ -2204,12 +2204,122 @@ func (s *ClusterScope) createVPCSecurityGroup(ctx context.Context, name string)
...
+func (s *ClusterScope) createVPCSecurityGroupRules(ctx context.Context, rules []infrav1.VPCSecurityGroupRule, securityGroupID string, existingRules []vpcv1.SecurityGroupRuleIntf) ([]string, error) {
 	log := ctrl.LoggerFrom(ctx)
 	var ruleIDs []string
 
-	log.V(3).Info("Creating VPC security group rules", "securityGroupID", securityGroupID, "ruleCount", len(rules))
+	log.V(3).Info("Reconciling VPC security group rules", "securityGroupID", securityGroupID, "specRuleCount", len(rules), "existingRuleCount", len(existingRules))
+
+	// Index existing rules by key so we can skip ones already present.
+	existingByKey := make(map[sgRuleKey]string, len(existingRules))
+	for _, rIntf := range existingRules {
```
Karthik-K-N:
```suggestion
	for _, rule := range existingRules {
```

pkg/cloud/scope/powervs/powervs_cluster.go
```diff
@@ -2266,33 +2376,42 @@ func (s *ClusterScope) createVPCSecurityGroupRules(ctx context.Context, rules []
...
+	remoteOption, err := s.resolveRuleRemote(remote)
+	if err != nil {
+		return "", err
+	}
+
+	// 2. Check whether this rule already exists.
+	key := sgRuleKey{
+		direction: direction,
+		protocol:  protocol,
+	}
```
Karthik-K-N, line 2394: Is this key unique? can't we have multiple rules with same direction and protocol but different port?

Karthik-K-N (commented):
Few questions. also you can refer to previous release for the reference on how to idem potently handle the reconciling the SG

aman4433:
> Few questions. also you can refer to previous release for the reference on how to idem potently handle the reconciling the SG

I checked release-0.10 to 0.13, and the SG logic is the same in all four. It uses `SecurityGroupRuleSecurityGroupRuleProtocolAll`, which is `SecurityGroupRuleProtocolAny` in vpc-go-sdk v0.90.

I’ll keep the same naming/structure with validate* and a separate remote comparison.

One thing: the old code recreates the full spec set if any rule doesn’t match. Mine checks rules individually and creates only the missing ones. Which approach do you prefer?

pkg/cloud/scope/powervs/powervs_cluster.go
aman4433: Yes, the initial direction + protocol isn't the complete key. For TCP/UDP, portMin/portMax are added, and the remote target (cidrBlock/address/crn) is also included. The full combination is used to determine uniqueness.

Karthik-K-N:
> > Few questions. also you can refer to previous release for the reference on how to idem potently handle the reconciling the SG
> 
> I checked release-0.10 to 0.13, and the SG logic is the same in all four. It uses `SecurityGroupRuleSecurityGroupRuleProtocolAll`, which is `SecurityGroupRuleProtocolAny` in vpc-go-sdk v0.90.
> 
> I’ll keep the same naming/structure with validate* and a separate remote comparison.
> 
> One thing: the old code recreates the full spec set if any rule doesn’t match. Mine checks rules individually and creates only the missing ones. Which approach do you prefer?

I think the way your doing it should be the ideal, right? what do you think?

aman4433:
> > > Few questions. also you can refer to previous release for the reference on how to idem potently handle the reconciling the SG
> > 
> > 
> > I checked release-0.10 to 0.13, and the SG logic is the same in all four. It uses `SecurityGroupRuleSecurityGroupRuleProtocolAll`, which is `SecurityGroupRuleProtocolAny` in vpc-go-sdk v0.90.
> > I’ll keep the same naming/structure with validate* and a separate remote comparison.
> > One thing: the old code recreates the full spec set if any rule doesn’t match. Mine checks rules individually and creates only the missing ones. Which approach do you prefer?
> 
> I think the way your doing it should be the ideal, right? what do you think?

Yes, I think per-rule is better.

pkg/cloud/scope/powervs/powervs_cluster.go
aman4433: Multiple types in one case give you an interface{}, so r.ID won’t compile. All six SDK types have an ID field but no shared interface exposing it, so the switch has to be expanded. The repetition is a Go limitation, not a design choice.

Karthik-K-N (commented):
thank you for identifying the bug and fixing it
/lgtm

/hold for @Amulyam24's review

Amulyam24 (commented):
/approve

/unhold

Thank you @aman4433!

PR #2957, opened 2026-09-08
🐛 Fix data race on v1beta1 conditions in IBMPowerVSCluster reconcile

Pacho20:

`reconcilePowerVSResources` and `reconcileVPCResources` run concurrently on the same `IBMPowerVSCluster`. The v1beta2 conditions were already written under `powerVSCluster.mu` via `updateCondition`, but every deprecated v1beta1 write called `deprecatedv1beta1conditions.Set/MarkTrue/MarkFalse` directly on the shared object, outside the lock.

Those setters do a read-modify-write on `Status.Deprecated.V1Beta1.Conditions` (get slice, append, sort, set back), so two concurrent writers can both read the same slice and the second write drops the first condition.

This PR adds `updateV1Beta1Condition`, the v1beta1 counterpart of `updateCondition`, and routes all 16 v1beta1 condition writes through it so both condition lists are protected by the same mutex.

Note that `main` is not affected: the refactor to `componentResult` there already has the concurrent reconcilers return their conditions for the caller to apply sequentially. This is the equivalent minimal fix for `release-0.14`, which still carries both the goroutine design and the v1beta1 conditions.

Fixes the flake in `TestIBMPowerVSClusterReconciler_reconcile`, where `NetworkReady` or another condition is intermittently missing from the expected set. Seen on unrelated PRs, e.g. #2934 (otel bump) and #2913 (doc link change).

/area provider/ibmcloud

**Release note**:

```release-note
Fix a data race in the IBMPowerVSCluster controller
```

### DIFF
--- a/internal/controllers/powervs/ibmpowervscluster_controller.go
+++ b/internal/controllers/powervs/ibmpowervscluster_controller.go
@@ -75,6 +75,14 @@ type IBMPowerVSClusterReconciler struct {
 	ClientFactory powervsscope.ClientFactory
 }
 
+// powerVSCluster wraps the IBMPowerVSCluster so that the PowerVS and VPC
+// reconcilers, which run concurrently, can record conditions on the same
+// object safely.
+//
+// Both the v1beta2 and the deprecated v1beta1 condition setters do a
+// read-modify-write on a slice hanging off the cluster status, so every write
+// has to go through mu. Use updateCondition and updateV1Beta1Condition, never
+// the conditions packages directly on cluster.
 type powerVSCluster struct {
 	cluster *infrav1.IBMPowerVSCluster
 	mu      sync.Mutex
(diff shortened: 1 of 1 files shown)


Karthik-K-N (commented):
Thanks for the PR.
but that was intentional, curious to know if you encountered any issues given its in PowerVS workflow or just for perfectness you want to get this added.

/ok-to-test

Pacho20:
> Thanks for the PR. but that was intentional, curious to know if you encountered any issues given its in PowerVS workflow or just for perfectness you want to get this added.

One of my PRs for release-0.14 had a flaky Prow job caused by this issue. My PR was completely unrelated to PowerVS, so I figured if I or someone else ever needs to merge something into release-0.14, we should fix this to prevent further flaky tests.

Do you think this change could cause any other issues? I don't want to break PowerVS support if this behavior is intentional.

Karthik-K-N (commented):
/lgtm

Karthik-K-N:
> > Thanks for the PR. but that was intentional, curious to know if you encountered any issues given its in PowerVS workflow or just for perfectness you want to get this added.
> 
> One of my PRs for release-0.14 had a flaky Prow job caused by this issue. My PR was completely unrelated to PowerVS, so I figured if I or someone else ever needs to merge something into release-0.14, we should fix this to prevent further flaky tests.
> 
> Do you think this change could cause any other issues? I don't want to break PowerVS support if this behavior is intentional.

I dont think it will cause any other issue, Happy to merge but was curious to know the reason

Thank you.

Amulyam24 (approved):
/lgtm

/cc @Prajyot-Parab

